PR #1, opened 2016-11-08
fix dead links to stripes-experiments and stripes-connect repo

julianladisch:


MikeTaylor:
Thanks for pointing this out -- the links need fixing. But I want to keep them relative (on the assumption that the repositories are checked out next to each other). Do you want to make another patch, or shall I do it?

MikeTaylor:
OK, I see it can't be sensibly done with relative links. So I will merge.

PR #6, opened 2017-04-25
Remove a line that disrupts the log in

hjiebsco:
Cannot log into UI. It seems the line resets UI during the log in process.


MikeTaylor:
Yes, I saw that in the last commit and thought it looked wrong. (Actually, I am not convinced by the whole commit -- I pinged Niels-Erik and am waiting to hear back from him.) Anyway, I will accept this.

PR #9, opened 2017-07-13
Introduce middleware for side effects. STRIPES-394

mkuklis:


MikeTaylor:
I worry that this makes stripes-core dependent on stripes-connect -- which it never was before. (It _appears_ to be dependent on stripes-connect, because for historical reasons the Trivial module is included in the same git repo, and _that_ depends on stripes-connect. But that is just an organisational error.)

Can the side-effects middleware that you put in stripes-connect easily be moved to its own package, or does it use other stripes-connect code?

mkuklis:
@MikeTaylor unfortunately there is currently no way to register additional middleware. I assumed that `stripes-core` depends on `stripes-connect` because of: https://github.com/folio-org/stripes-core/blob/master/package.json#L50

We could introduce a new package for side effects and start collecting various side effects there and expose a middleware from there too. Would that be a better approach?

MikeTaylor:
Right, you mean that all middleware has to be registered at the time the store is created and there is no `store.registerMiddleware` method that you can call after the event. True -- so stripes-connect cannot pull in the extra middleware it wants at the point where it wants it.

But, yes, what I had in mind was a separate package for the middleware. But I admit I've not read the code: if it depends on other stuff that's in stripes-connect then there would be no reason to introduce the extra level of indirection. But if (like say `AugmentCrudActions.js`) it doesn't depend on any other part of stripes-connect, then it can come out of stripes-connect into a separate git repo and NPM package. (As `AugmentCrudActions` arguable should, too.)

As for https://github.com/folio-org/stripes-core/blob/master/package.json#L50 -- that is a historical accident. Stripes-core does not depend on stripes-connect, but the Trivial module does and it's currently still in bed with stripes-core.

mkuklis:
sounds good @MikeTaylor I think it makes sense to extract it to a separate package. @skomorokh suggested https://github.com/folio-org/stripes-experiments to start with.

MikeTaylor:
Dear lord, NO!

Stay out of the hell-pit that is stripes-experiments. We'll only have to move it again. That place is a wastebasket for every bit of random stray fluff we have on use for. Please, make a dedicated repo from the start. `stripes-redux-middleware` seems right to me?

mkuklis:
Haha you must be aware of something I'm not. I like `stripes-redux-middleware` but I think the idea here is to expose couple additional `util` functions as well (not just middleware) to allow for registering `epics` (side effects). What about `stripes-redux`?

MikeTaylor:
Sure, stripes-redux sounds fine. (It would also make sense, once we have that, for me to move `CrudActionsAugmenter` into that repo.)

mkuklis:
@MikeTaylor that sounds good. I think `CrudActionsAugmenter` is a good candidate for that. I'm going to close this PR.

PR #16, opened 2017-08-02
Change the welcome message

hjiebsco:
May I suggest to change the welcome message from "Welcome to whatever this is!" to "Welcome, the Future Of Libraries Is OPEN!"? Thanks.

### DIFF
--- a/src/components/Front.js
+++ b/src/components/Front.js
@@ -3,7 +3,7 @@ import Link from 'react-router-dom/Link';
 
 export const Front = () => (
   <div>
-    <h3>Welcome to whatever this is!</h3>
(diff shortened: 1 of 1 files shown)


skomorokh:
LOL, sure, no prob, surprised that text made it this far. I do hope we can do something useful with that space or just go directly into the most recently used app but I've no objection to a more polished placeholder in the interim---I wrote the original before there was consensus on a name and it was just one in a series of prototypes.

MikeTaylor:
... but I do mourn the passing of "Welcome to whatever this is". It had a certain something, and did do a good job of conveying clearly that the code is not a product! Oh well.

PR #17, opened 2017-08-03
Handle login via SSO. Fixes STCOR-43

mkuklis:
The PR should handle login via SSO when both `ssoToken` and `userId` are provided via query params or cookies.


MikeTaylor:
Michal, this looks fine to me. Is there any specific reason why you didn't go ahead and merge your own PR?

mkuklis:
Thank you for the feedback @MikeTaylor I just wanted to make sure it looks good to you before I merged. I will do that now.

PR #18, opened 2017-08-04
Remove unused nav buttons

cowboyd:
The top navigation bar is populated via the automatically discovered list of applications. However, in addition to those, there is a list of six navigation buttons that don't appear to serve any purpose other than to demonstrate the visual style. For someone using the system this can be confusing since it immediately raises the question "should this do something?", or "should it go somewhere?"

This removes those questions by removing those buttons.

### DIFF
--- a/src/components/MainNav/MainNav.js
+++ b/src/components/MainNav/MainNav.js
@@ -174,12 +174,6 @@ class MainNav extends Component {
           <NavGroup>
             {menuLinks}
             <NavDivider md="hide" />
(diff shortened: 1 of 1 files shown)


MikeTaylor:
Folks, I am in favour of this. Is there are reason not to do it? @skomorokh @skoczko

skoczko:
I'd say go ahead and accept it. @Baroquem mentioned that they are placeholders for the widgets like Notifications and Bookmarks. But I'm in favor of hiding them before we get anything usable.

MikeTaylor:
It is done!

PR #20, opened 2017-08-07
Clean up prop-types related deprecations

cowboyd:
This removes warnings generated like the following:

>Accessing PropTypes via the main React package is deprecated, and will be removed in  React v16.0. Use the latest available v15.* prop-types package from npm instead. For info on usage, compatibility, migration and more, see https://fb.me/prop-types-docs

### DIFF
--- a/src/AddContext.js
+++ b/src/AddContext.js
@@ -1,17 +1,18 @@
 import React from 'react';
+import PropTypes from 'prop-types';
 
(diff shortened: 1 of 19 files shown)


MikeTaylor:
This fixes the stripes-core part of https://issues.folio.org/browse/STRIPES-427

PR #22, opened 2017-08-09
[STCOR-46] Make Front page component pluggable.

cowboyd:
The landing page of a stripes application is the first thing that a user sees when they navigate to it. At the moment, there is a placeholder text there, but you can't customize that text.

This change adds a new type of plugin "frontpage" which, if present will be presented to the user.

folio_ui_experimentation

Implementation Notes
====================

- if a plugged-in component is rendered in a stripes connected environment, the `Pluggable` component will attempt to connect the component that it wraps. However, if we're rendering outside the context of an app (as in the case of the frontpage), then the `connect()` method is not present on the stripes context. This adds a default `connect()` which passes through the original component without any decoration

- The `Front` component needed to be extended because it now needs to setup a context in order to support pluggability. We should consider having a default context that is set up at the very top of the render tree.

### DIFF
--- a/src/Root.js
+++ b/src/Root.js
@@ -95,6 +95,7 @@ class Root extends Component {
         user: currentUser,
         perms: currentPerms,
       },
+      connect(X) { return X; }
     });
 
     return (
@@ -106,8 +107,8 @@ class Root extends Component {
                 <MainNav stripes={stripes} />
                 <ModuleContainer id="content">
                   <Switch>
-                    <Route exact path="/" component={Front} key="root" />
-                    <Route path="/sso-landing" component={Front} key="sso-landing" />
+                    <Route exact path="/" component={()=> <Front stripes={stripes}/>} key="root" />
+                    <Route path="/sso-landing" component={()=> <Front stripes={stripes}/>} key="sso-landing" />
                     <Route path="/about" component={() => <About stripes={stripes} />} key="about" />
                     <Route path="/settings" render={() => <Settings stripes={stripes} />} />
                     {getModuleRoutes(stripes)}
--- a/src/components/Front.js
+++ b/src/components/Front.js
@@ -1,10 +1,27 @@
 import React from 'react';
+import PropTypes from 'prop-types';
 import Link from 'react-router-dom/Link';
+import AddContext from '../AddContext';
 
-export const Front = () => (
-  <div>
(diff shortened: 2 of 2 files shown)


MikeTaylor:
Thanks for this -- a really nice, clean change-set with an exemplary explanatory comment. This pull-request should be use as an exemplar in our CONTRIBUTORS document :-)

I'll go ahead and merge, but just a couple of things for next time:
* I had a Jira issue to work on this: https://issues.folio.org/browse/STCOR-46. If you want to take work that's assigned to someone else, that's fine, but it helps if you communicate it by taking the Jira issue first. (Happily, I'd not spent any real time on this).
* I'd prefer to see only one "TODO" on the three-line comment, so we can count how many TODO issues we have with a simple `grep -c TODO`. No big.
* It's been standard practice to pull stripes-components from their own source area, rather than from the top-level index, so doing it that way here doesn't really merit a comment. But if you think we should change that standard practice, I'd be interested in the reasoning. (I have no dog in the fight.)

cowboyd:
Thanks for the feedback! I'm still getting my JIRA sea-legs so to speak, and yes, totally makes sense to only have one line of `TODO`. I'm always happy to make changes if you want to hold of merging.

As for accessing components from `lib/` I have a preference for accessing them from the index. To me, `lib/` is an implementation detail of the package, and it feels like accessing private data to reach around and grab specific files from there. Among other things, package authors should have the latitude to move source files around or decorate classes and functions without the consuming code needing to change.

MikeTaylor:
Thanks, Charles. No need to hold off the merge before tiny tweaks like those I suggested, they can be done any time.

Your reasons for wanting to avoid including from `lib` in stripes-components all make sense to me, and I'm inclined to think we should start gently migrating existing uses accordingly. @JohnC-80, do you agree? Disagree? Are we missing something?

MikeTaylor:
Oh, I'd hoped @JohnC-80 would get automatically added to the "Participants" of this PR when I mentioned him in a comment, but apparently not. Do you know whether he'll get an email notification of the mention?

cowboyd:
It will depend on his notification settings, but I believe it should. He might not show up in the participants list until the thread contains a comment from him.

MikeTaylor:
Come on, @JohnC-80! Validate Charles's interpretation of GitHub pull-request comment-thread mention semantica!

PR #23, opened 2017-08-10
Init i18n setup. Fixes Stripes-447

mkuklis:


MikeTaylor:
'<Root>`'s `componentWillMount` method seems like an odd place to invoke `loadTranslations`. What was the thinking there?

MikeTaylor:
Anyway, that wrinkle aside, this all looks like very neat code. I'm proceeding with the merge. Thank you.

mkuklis:
@MikeTaylor this will hopefully go away after we introduce a way to fetch locale without authentication. For now we need to load translations initially based on the default locale. I could probably move this call to LoginCtrl.

MikeTaylor:
With the present code, do we load translations once at startup, then again after login? (I think that is going to be the only real option.)

mkuklis:
@MikeTaylor we need the translations before everything else is rendered. After we open fetching the locale from a public endpoint we will fetch it once (probably before the Root is even initialized and just pass it as a prop to Root). I just turned off fetching locate after login. Let's wait for https://issues.folio.org/browse/FOLIO-765 to be done and we can clean it up. 

BTW there is potentially another public config needed before authentication to check if the SSO is enabled in the system so as you mentioned under FOLIO-765 there are other use cases for it.

MikeTaylor:
> After we open fetching the locale from a public endpoint we will fetch it once (probably before the Root is even initialized and just pass it as a prop to Root).

I don't think we'll _ever_ escape needing to do it twice: once to know how to present the login page, and once when we've established who the user is, and what their preference is. (Consider a Spanish-speaking student at a primarily Anglophone university.)

mkuklis:
@MikeTaylor you are right! I turned fetching locale and loading translations back on after authentication. We should be able to remove `loadTranslations` from `componentWillMount` after the work in FOLIO-765 is done. For now I will keep it there because if I remove it `IntlProvider` will complain.

PR #24, opened 2017-08-10
Set initial viewport state with meta tag

cherewaty:
## Before
Currently, opening a FOLIO application on a mobile browser will usually result in the page width being calculated as much wider than the device's screen (and zoomed out):

before

## After
Instead, by default we can set the width of the page to equal the width of the device, with the zoom at 100%. Users can then zoom in more if needed. https://developer.mozilla.org/en-US/docs/Mozilla/Mobile/Viewport_meta_tag

after

### DIFF
--- a/index.html
+++ b/index.html
@@ -2,6 +2,7 @@
 <html>
   <head>
     <title>Folio UI Experimentation</title>
+    <meta name="viewport" content="width=device-width, initial-scale=1">
     <link rel="stylesheet" href="/bootstrap/css/bootstrap.min.css">
   </head>
   <body>


MikeTaylor:
Thanks for this, I merged it.

But why would this obviously correct behaviour not be the default?

cherewaty:
I think it's just a vestige from when almost no websites were a mobile screen's width; the default behavior still lingers from that era. Apple's own docs on it are looking pretty old, but explain it well: https://developer.apple.com/library/content/documentation/AppleApplications/Reference/SafariWebContent/UsingtheViewport/UsingtheViewport.html

PR #30, opened 2017-08-16
change de translation of front.about

julianladisch:


julianladisch:
Part 1 of STRIPES-458

MikeTaylor:
This is such a classic German word :-)

PR #31, opened 2017-08-17
Optimize Root render. Fixes STCOR-58.

mkuklis:


MikeTaylor:
Nice work.

One day, I will have figure out what the heck an epic is. Did you actually _need_ one for this, or did you merely find it a more convenient approach that what we've been doing for other Redux actions in stripes-core?

Anyway, I'll go ahead and merge this.

mkuklis:
@MikeTaylor thank you for merging this. I did notice a little improvement when folio loads for the first time. We could completely avoid epic here but then we would have to dispatch `OKAPI_READY` in couple different places. The actions which trigger `OKAPI_READY` may change in the future. With epic we will have one place to adjust it.

MikeTaylor:
So is the epic here just a declarative statement of which action-types should trigger the associated code-fragment?

mkuklis:
Yes. That's a nice way of thinking about it and also one of the use cases. It's a `side effect` (I actually don't like the word effect). It's a way to execute something after some other action(s) executed (and the new state made it to the redux store). The most common use case is to trigger a new action. It sits on top of rxjs so we are able to use reactive patterns to structure more complex scenarios (for example we could merge multiple streams before moving forward, filter them, etc). A nice intro to rxjs is here: https://github.com/btroncone/learn-rxjs

PR #33, opened 2017-08-18
Hide focus outline for moduleContainer

cherewaty:
### Before
before

### After
after


MikeTaylor:
Thanks for this, it's been bothering me!

PR #34, opened 2017-08-21
STCOR-61 Update user retrieval after SSO login.

szkata:
The /_self endpoint should be used for retrieving the current user instead of the /bl-users/by-id/{userId} endpoint from mod-users-bl. If the latter is used the user is required to have a permission for retrieving the user.

### DIFF
--- a/src/components/SSOLanding.js
+++ b/src/components/SSOLanding.js
@@ -21,10 +21,9 @@ class SSOLanding extends Component {
 
   componentWillMount() {
     const token = this.getToken();
-    const userId = this.getUserId();
 
-    if (token && userId) {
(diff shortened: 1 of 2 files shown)


MikeTaylor:
You seem to have made other changes in this PR, too, to do with ignoring the user-ID. What was the thinking there?

szkata:
The userId was only forwarded because this endpoint (/bl-users/by-id/{userId} ) needed it to retrieve the user. Since the /bl-users/_self endpoint extracts the user id from the authentication token it is not necessary anymore.

MikeTaylor:
Oh, of course -- makes perfect sense! Thanks.

PR #35, opened 2017-08-22
Stripes 464

jeremythuff:
This is a simple implementation of "stripes.runModule" to satisfy the requirements of [STRIPES-464](https://issues.folio.org/browse/STRIPES-464).

Currently it takes a component and optional embedded flag and returns a "connectedApp" or "runningModule." The question for reviewers is: was it a mistake to wrap stripes.connect? It allows for cleaner use of runModule, as `props.stripes.runModule(props.stripes.connect(Module))` is arduous. But, will there be a time when we want to run a module that is not connected?

[PR#58](https://github.com/folio-org/ui-users/pull/58) to ui-users is an example of this method being used to start a top level module.


MikeTaylor:
Sorry to be negative, but I am not crazy about this for three reasons.

First, the trivial one: you misspell "embedded" as "embeded" :-)

Second, the one you mention: automatically stripes-connecting the embedded module smells of Too Much Magic to me. You say that `props.stripes.runModule(props.stripes.connect(Module))` is arduous; but this is hardly something that is going to be all over the code: embedding a module is going to be a relatively rare operation, so expending a few extra keystrokes over it seems a price well worth paying.

The last is more fundamental: unless I am mis-reading, https://github.com/folio-org/stripes-core/blob/4c8e6d756f190c2ff7871f1c795d9d85e0551691/src/Stripes.js#L61 modifies the existing Stripes object, which will affect the behaviour of the top-level module. Instead, we need to give the newly created module its own Stripes object which has the appropriate `embedded` setting. It would be something like:`const newStripes = this.clone({ embedded })` and then start the new modules with that Stripes object.

Hope this is helpful.

jeremythuff:
This new take on the problem should satisfy your concerns.

Instead of "runModule" which was not really running anything, I have replaced it with configureModule, which allows for the expansion of a modules stripes object's config. I have also separated out the connecting functionality from the method. 

It can be used as such:

```
this.embeddedModule = props.stripes.configureModule(props.stripes.connect(Users), {
      embedded: true,
});
``` 

:) and I stopped trying to spell "embedded"

PR #36, opened 2017-08-23
STCOR-62 Redirect user to original url after SSO authentication

szkata:
A new component was created to redirect the user from the /sso-landing page after login to the original url. Also the path of the original login request is preserved (in loginServices).

### DIFF
--- a/src/RootWithIntl.js
+++ b/src/RootWithIntl.js
@@ -14,6 +14,7 @@ import ModuleContainer from './components/ModuleContainer';
 import Front from './components/Front';
 import About from './components/About';
 import SSOLanding from './components/SSOLanding';
+import SSORedirect from './components/SSORedirect';
 import Settings from './components/Settings/Settings';
 import LoginCtrl from './components/Login';
 import getModuleRoutes from './moduleRoutes';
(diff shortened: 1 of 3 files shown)


MikeTaylor:
What is the point in conditiotionally setting stripesUrl, only to immediately unconditionally set it?
https://github.com/qultoltd/stripes-core/blob/d11ccd105c1dee80a858a2f995dac9ebfebf1053/src/loginServices.js#L255-L259

MikeTaylor:
I admit I've lost track of how all this works now, but do we need both SSOLanding _and_ SSORedirect, or should the latter replace the former? In the present version, they seem to be fighting over the path `/sso-landing`.

szkata:
For your first question: I am only appending the path to the url, not setting it again. The condition only exists as a workaround for browsers that do not support using the window.location.origin.
I separated the SSOLanding component to two parts, the first one (SSOLanding) is responsible for processing the request - get the authentication token, set the logged in user object. The SSORedirect is rendered when the user is logged in - when the token is present - and it's purpose is only to redirect the user to the origin of his login request. For example if he wanted to reach the user page before login, he is redirected there this way. Although I am open to other solutions and if you have a suggestion please don't hesitate to share it with me.

szkata:
I updated the url assignment int our git repo to this:
```const stripesUrl = (window.location.origin || `${window.location.protocol}//${window.location.host}`) + window.location.pathname;```

MikeTaylor:
Oh, that was stupid of me to miss the `+=` in the second assignment to `stripesUrl`! Sorry!

I'm not following the whole sequence of operations in Stripes SSO, to be honest; but if you are convinced that both SSOLanding and SSORedirect are doing useful jobs, that's good enough for me! I will go ahead and merge.

Thanks for your patience.

PR #38, opened 2017-09-05
Fix redux devtools compose

wwilsman:
The previous method of composing the middleware would call the middleware initializers twice, and so a side-effect of this was that epics would receive double the actions.

This changes the compose method to use redux devtool's own compose (according to [their docs](https://github.com/zalmoxisus/redux-devtools-extension#12-advanced-store-setup)) and thus fixes the double epic calls.

This also fixes the logger logging itself (the `PERFORM_ACTION` action in the first gif below).

### Before
bug

### After
fix

### DIFF
--- a/src/configureStore.js
+++ b/src/configureStore.js
@@ -4,19 +4,15 @@ import { createLogger } from 'redux-logger';
 import initialReducers from './initialReducers';
 import enhanceReducer from './enhanceReducer';
 
+const enhanceMiddleware = window.__REDUX_DEVTOOLS_EXTENSION_COMPOSE__ || compose;
+
 export default function configureStore(initialState, stripesLogger, epics) {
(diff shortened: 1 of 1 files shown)


sadatay:
✨ fantastic work

wwilsman:
After adding those gifs, I'm thinking it has something to do with that `PERFORM_ACTION` action that get's fired after the initial action in the before gif

MikeTaylor:
Delighted to see this! I assume it fixes https://issues.folio.org/browse/STCOR-68?

wwilsman:
@MikeTaylor It might! We were seeing double actions when an action was emitted from an epic (because the epic was firing twice). It could also be that the logger was logging itself, therefore maybe causing the log to show two of each action.

PR #40, opened 2017-09-15
Wait until discovery is finished before rendering the module container

cowboyd:
## Problem

Application modules can be rendered before the services that they depend on have been discovered. The `discoverServices` thunk can potentially dispatch multiple `DISCOVERY_SUCCESS` and `DISCOVERY_FAILURE` actions while discovery is happening, but there is no action to mark when the discovery process is fully complete. Without this capability, it is possible for an application to render without knowing if its dependencies are satisfied on the backend.

## Approach

This ensures that whenever a promise is generated inside the discovery process, it is returned so that it can be used for sequencing further promises. Then, once the discovery process is complete, a final `DISCOVERY_COMPLETE` is dispatched. Because individual discovery errors are caught, this will happen regardless of whether the discovery partially failed.

To signal readiness to application modules, the `discovery.isFinished` property is added to the redux state, and app modules are _only rendered if discovery is complete_

## Topics for further discussion

- refactor the discovery process to use redux-observable epics.
- finer grained tracking of which components have been discovered. That way, if my app only needs a single interface, I don't need to wait for others to be discovered.
- have OKAPI gateway make all discovery information available in a single request (unlike the 10+ it makes now.)

### DIFF
--- a/src/Root.js
+++ b/src/Root.js
@@ -120,6 +120,7 @@ Root.propTypes = {
   discovery: PropTypes.shape({
     modules: PropTypes.object, // eslint-disable-line react/forbid-prop-types
     interfaces: PropTypes.object, // eslint-disable-line react/forbid-prop-types
+    isFinished: PropTypes.boolean,
   }),
   history: PropTypes.shape({
     length: PropTypes.number.isRequired,
--- a/src/RootWithIntl.js
+++ b/src/RootWithIntl.js
@@ -32,21 +32,23 @@ const RootWithIntl = (props, context) => {
           { token || disableAuth ?
             <MainContainer>
               <MainNav stripes={stripes} />
-              <ModuleContainer id="content">
-                <Switch>
-                  <Route exact path="/" component={() => <Front stripes={stripes} />} key="root" />
-                  <Route path="/sso-landing" component={() => <SSORedirect stripes={stripes} />} key="sso-landing" />
-                  <Route path="/about" component={() => <About stripes={stripes} />} key="about" />
-                  <Route path="/settings" render={() => <Settings stripes={stripes} />} />
-                  {getModuleRoutes(stripes)}
-                  <Route
-                    component={() => (<div>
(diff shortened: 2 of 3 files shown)


cowboyd:
Looks like this failed with the rather cryptic error message:

```
Pull request #40 opened
Connecting to https://api.github.com using id-jenkins/****** (Jenkins U/P with oath token)
Connecting to https://api.github.com to check permissions of obtain list of thefrontside for folio-org/stripes-core

GitHub has been notified of this commit’s build result

java.lang.ArrayIndexOutOfBoundsException
Finished: FAILURE
```

Looks like a build hiccough? https://jenkins-aws.indexdata.com/blue/organizations/jenkins/folio-org%2Fstripes-core/detail/PR-40/1/pipeline

MikeTaylor:
What is the status of this? Is it passing tests now?

> *Topics for further discussion*
> refactor the discovery process to use redux-observable epics.

What would be the advantage of this?

> have OKAPI gateway make all discovery information available in a single request (unlike the 10+ it makes now.)

That is https://issues.folio.org/browse/STCOR-59

zburke:
This branch runs fine for me locally. 

The first three builds appear to have failed due to a permissions issue, and then maybe @funkymalc made some config changes and ran a few manual builds to test the new config. Should we check in with him? I'm perplexed reviewing the log at https://jenkins-aws.indexdata.com/job/folio-org/job/stripes-core/view/change-requests/job/PR-40/4/console that includes "Loading trusted files from base branch master at 4a5403257c2f5598beeb6f606d57d799a498596a rather than 837c806ce56e9cb87ee67ea95c720dbd9bbe6191" The first hash is the current tip of master and the second one is the hash for this fork. That makes it sound like it ignored the fork, doesn't it?

funkymalc:
I think there is an issue here merging master into the PR.

MikeTaylor:
So we're waiting for @cowboyd to manually merge master into this PR?

PR #41, opened 2017-09-15
Upgrade ESLint and fix broken rules

cherewaty:
I noticed on our builds of `thefrontside/ui-eholdings` that we were still getting `WARN: 'Warning: Accessing PropTypes via the main React package is deprecated, and will be removed in React v16.0. Use the latest available v15.* prop-types package from npm instead. For info on usage, compatibility, migration and more, see https://fb.me/prop-types-docs'`, even though I thought it had been resolved with https://github.com/folio-org/stripes-components/pull/45.

It turns out I missed one - in searching for `React.PropTypes`, I didn't notice an instance of `import React, { PropTypes } from 'react';`.

That would've been caught by the ESLint `react/no-deprecated` rule provided as part of `eslint-config-airbnb`, but the error was not being thrown. I upgraded all the `eslint`-related packages, and then there were errors in my editor and when running `yarn lint` (`react-no-deprecated` also needed to know the target `react` version to pop up).

I fixed all the linting errors, with only a few uses of `eslint-disable-line`.

### Next Steps
It would be awesome if we ran the linter as part of a continuous integration environment, so future pull requests to `stripes-core` with ESLint errors could not be merged.


cherewaty:
Related to https://github.com/folio-org/stripes-components/pull/54

.eslintrc
```diff
@@ -12,6 +20,7 @@
     "no-unused-vars": [ 1, { argsIgnorePattern: "^_" } ],
     "curly": "off",
     "no-plusplus": "off",
-    "no-continue": "off"
+    "no-continue": "off",
+    "jsx-a11y/href-no-hash": "off"
```
cherewaty: https://github.com/airbnb/javascript/issues/1474

cherewaty:
Replaced by https://github.com/folio-org/stripes-core/pull/42

PR #42, opened 2017-09-22
Use eslint-config-stripes

cherewaty:
Switches to using shared FOLIO eslint config: https://github.com/folio-org/eslint-config-stripes


src/Root.js
```diff
@@ -48,9 +48,8 @@ class Root extends Component {
       this.epics[key] = epic;
       this.props.epics.add(epic);
       return true;
-    } else {
-      return false;
     }
+    return false;
```
cowboyd: Is this a linting rule that got triggered?

cherewaty: Yes https://eslint.org/docs/rules/no-else-return
I have mixed feelings about that rule.

cherewaty:
Re: Jenkins build failures: is it running into a cached npm registry or something? `debug`, a dependency of `finalhandler` (a dependency of `express`), had a `2.6.9` release 8 hours ago that `npm install` on Jenkins can't find.

src/Root.js
cowboyd: Yeah, these days I favor symmetry over efficiency when given a choice

MikeTaylor: As a general policy, we want ESLint to be our servant, not our master. Bondage-and-discipline rules like no-nested-ternary and no-plusplus can go, with no period of mourning required. no-else-return definitely belongs in this category!

funkymalc:
Right.  npmjs.org is cached.    The builds passes now.

src/Root.js
cherewaty: https://github.com/folio-org/eslint-config-stripes/pull/3

PR #44, opened 2017-09-25
Switch to babel-preset-env

cherewaty:
By default, `babel-preset-env` will continue to have the same behavior we had with `babel-preset-es2015`. It has additional options for delivering more modern JavaScript, depending on the browser configuration provided.

Switching will eliminate this warning:
`We're super 😸  excited that you're trying to use ES2015 syntax, but instead of continuing yearly presets 😭 , we recommend using babel-preset-env: npm install babel-preset-env. preset-env without options will compile ES2015+ down to ES5. And by targeting specific browsers, Babel can do less work and you can ship native ES2015+ to users 😎 ! Also, we are in the process of releasing v7, so give http://babeljs.io/blog/2017/09/12/planning-for-7.0 a read and test it! Thanks so much for using Babel 🙏 , please give us a follow @babeljs for updates, join slack.babeljs.io for discussion and help support at opencollective.com/babel`

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,3 +1,3 @@
 {
-  "presets": ["es2015", "stage-2", "react"]
+  "presets": ["env", "stage-2", "react"]
 }
--- a/examples/trivial/package.json
+++ b/examples/trivial/package.json
@@ -29,7 +29,7 @@
   "devDependencies": {
     "babel-core": "^6.17.0",
     "babel-eslint": "^7.0.0",
-    "babel-preset-es2015": "^6.18.0",
+    "babel-preset-env": "^1.6.0",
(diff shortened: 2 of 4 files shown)


MikeTaylor:
I am very much in favour of eliminating that warning.

MikeTaylor:
Looks like I am allowed neither to merge this PR, nor review it so that someone else can merge it.

I would ask to be given the permission to review the PR, but no doubt we don't have anyone on the team with authority to give that permission before their request to do so has been reviewed.

cowboyd:
@MikeTaylor not sure what the deal is. I was able to review it ok. Did the perms change on the repo somehow?

MikeTaylor:
Buggered if I know.

PR #45, opened 2017-09-26
STCOR-80: Allow adjust of publicPath

hjiebsco:
Hi there, the motivation of this change is to allow the build artifacts more flexible to deploy and better organized. So I can run something like: yarn build "bundle/2017-09-26_1146/" "/2017-09-26_1146/".

### DIFF
--- a/stripes.js
+++ b/stripes.js
@@ -4,6 +4,7 @@ const commander = require('commander');
 const webpack = require('webpack');
 const HardSourceWebpackPlugin = require('hard-source-webpack-plugin');
 const path = require('path');
+const fs = require('fs');
 
 const devConfig = require('./webpack.config.cli.dev');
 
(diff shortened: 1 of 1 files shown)


stripes.js
```diff
@@ -92,8 +93,19 @@ commander
       options: { stripesLoader: stripesLoaderConfig },
     }));
     config.output.path = path.resolve(outputPath);
+    if (publicPath) {
+      config.output.publicPath = publicPath;
+    }
     const compiler = webpack(config);
-    compiler.run(function (err, stats) { });
+    compiler.run(function (err, stats) {
+      // reason: index.html is not generated
```
ttamj: I'm inclined to make index.html generated rather than place this logic in here.  We can make use of HtmlWebpackPlugin or a custom plugin/template to do the same.  I feel this will be safer for future index modifications.  We can also begin to control bundle name output for cache busting in environments not using publicPath to do so.

hjiebsco: Totally agree. It would be easier if index.html is generated. Is there a plan to address that soon? Meanwhile is it OK to merge this small localized non-breaking change to meet my immediate need?

hjiebsco:
Will create separate PRs for html-webpack-plugin and exposing publicPath.

PR #46, opened 2017-09-27
Stcor 25 webpack plugin

ttamj:
This PR folds the configuration logic performed by stripes-loader into stripes-core as a webpack plugin that makes the resulting stripes configuration available as a virtual module. See STCOR-25.


package.json
```diff
@@ -39,14 +39,15 @@
...
+    "webpack-virtual-modules": "^0.1.8"
   },
   "dependencies": {
     "@folio/stripes-components": "^1.6.0",
-    "@folio/stripes-connect": "^2.7.0",
-    "@folio/stripes-loader": "^0.0.2",
+    "@folio/stripes-connect": "folio-org/stripes-connect#3c2714a189c2296f389fc6ee4e79a7ee7c962949",
     "@folio/stripes-logger": "^0.0.2",
-    "@folio/stripes-redux": "^2.0.0",
+    "@folio/stripes-redux": "folio-org/stripes-redux#138803e34c8472b657ef62df76316d941177f75f",
```
cherewaty: Could we just cut new releases of `stripes-redux` and `stripes-connect`? That seems more reliable than a commit hash.

ttamj: Yes. Until those releases are cut, things are broken for those working against the head of stripes-core. These hashes would be removed prior to the next release.

PR #50, opened 2017-09-28
Move epics to core. Fixes STCOR-82

mkuklis:
The PR should resolve our issues around rxjs operators and make side effects working again. It should be also easier to maintain it and we can get rid of stripes-redux.


mkuklis:
Thanks for the review @zburke. The stripes-redux only contained the epics work so far so we shouldn't see any other implications.

PR #54, opened 2017-10-02
STCOR-84 Use html-webpack-plugin

hjiebsco:


package.json
```diff
@@ -38,6 +38,7 @@
     "enzyme": "^2.5.0",
     "eslint": "^4.7.2",
     "eslint-config-stripes": "^0.0.1",
+    "html-webpack-plugin": "^2.30.1",
```
ttamj: FYI, The plugin will need to be included under "dependencies" rather than "devDependencies".  Dev dependencies won't get installed via npm/yarn when stripes-core is itself a dependency of clean and un-linked platform build.

hjiebsco: Moved dependency as you suggested.

ttamj:
Looks good. I like adding the hash to the prod bundles as it will provide natural cache busting independent of how Folio gets deployed.

hjiebsco:
I had the hash attribute earlier, then realized that there is already hash within bundle name itself.

PR #55, opened 2017-10-02
STCOR-80

hjiebsco:


stripes.js
```diff
@@ -90,15 +90,18 @@ commander
 
 commander
   .command('build')
-  .arguments('<config> <output>')
+  .arguments('<config> <output> [publicPath]')
```
ttamj: Can we pass publicPath with the option syntax?  See usage of .option() in the "dev" command above.  There are a few more options that will be added soon and this way we are not dependent on order.

hjiebsco: I compared both approaches and like the [publicPath] a bit better. The package.json is typically fixed with "stripes build stripes.config.js". If use the --publicPath, it seems that the package.json has to be changed for each build if changing publicPath while [publicPath] does not need to. Also, it seems that [publicPath] should be at the same level as <output> conceptually. Or as a comprise, we can keep both approaches at same time.

ttamj: The package.json doesn't have to be modified with each build. To pass options to a npm script like this, separate the options with `--`.  This is in addition to the `--` used for the option itself.  For example: `yarn build output -- --example=abc --devtool=sourcemap`

[Webpack's CLI](https://webpack.js.org/api/cli/#output-options) supports setting output options, including public path, via command line options like this, `--output-public-path`. As we support more webpack options through stripes, I think it will make sense to do so in a predictable way including webpack's names/convention where they overlap.

hjiebsco: Good to know about the double --. Changed to options approach.

PR #39, opened 2017-09-11
dropdown component update

TG-EBSCO:
Update stripes Dropdown component for login main nav

PR #58, opened 2017-10-05
Remove react-bootstrap

cherewaty:
All things `bootstrap` are being deprecated out of FOLIO: https://github.com/folio-org/stripes-components/blob/3e1a93e67c942a769838af0b2d240021f5a9f886/docs/CSSinStripes.md#bootstrap-deprecation. The `MainNav` component was the last instance of `react-bootstrap` in `stripes-core`: https://github.com/folio-org/stripes-core/pull/39

UI Modules still relying on `react-bootstrap` can add it to their own dependencies. New work should avoid using `react-bootstrap` or `bootstrap`'s CSS.


ttamj:
Should the reference to bootstrap in index.html also be removed?

cherewaty:
@ttamj yes, but since that'll be a more invasive change for everyone, I'll make a separate PR for it. This PR should be non-breaking.

PR #61, opened 2017-10-05
Remove bootstrap css

cherewaty:
All things Bootstrap are being deprecated: https://github.com/folio-org/stripes-components/blob/3e1a93e67c942a769838af0b2d240021f5a9f886/docs/CSSinStripes.md#bootstrap-deprecation 

Ripping this band-aid off is going to hurt, but the sooner we do it, the better. This will cause significant changes for the styles in all FOLIO apps. Most notable are changes to the font stack and link styles. Apps relying heavily on `stripes-components` instead of directly using `bootstrap` classes should see minimal bugs, but may want to adjust a few styles here and there.

To give some time for developers, PMs, and UX folks to become aware of this pending change (and because I'll have limited availability myself Oct. 9-13), let's not merge this before October 16.


cherewaty:
Waiting on https://issues.folio.org/browse/STCOM-75 so this causes fewer breaking changes.

skomorokh:
John's take on this is so much more relevant than mine and I wouldn't want it in until he was cool with it.

cherewaty:
Replaced by https://github.com/folio-org/stripes-core/pull/95

PR #63, opened 2017-10-05
Run lint on PRs

cherewaty:
This was the same approach taken for `stripes-components` - I assume it applies here too.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,4 +1,6 @@
 
 buildNPM {
   publishModDescriptor = 'yes'
-}  
+  runLint = 'yes'
+  runTest = 'yes'
+}
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
     "chai": "^4.0.2",
     "enzyme": "^2.5.0",
     "eslint": "^4.7.2",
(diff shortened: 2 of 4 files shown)


Jenkinsfile
```diff
@@ -1,4 +1,6 @@
 
 buildNPM {
   publishModDescriptor = 'yes'
-}  
+  runLint = 'yes'
```
cherewaty: This didn't appear to have any effect - there must be some config in Jenkins that isn't part of the repo.

funkymalc:
Looks like stripes-core is eslint error free.   I am curious what to make of the bugs and vulnerabilities reported by Sonarqube.  https://sonarcloud.io/dashboard?id=folio-org%3Astripes-core

cherewaty:
I'd rather not have sonarqube running in addition to eslint. Most of those "Remove this useless assignment" errors are disabled in eslint. It can be debated that maybe those shouldn't be disabled (some of them are easy to fix), but it's going to be difficult to maintain more than one static analysis source long-term.

MikeTaylor:
Agreed, though I don't feel strongly about it.

PR #64, opened 2017-10-05
Lint entire repo

cherewaty:
Gets `stripes-core` to 100% lint coverage.

I'm a little hesitant to merge this without any `stripes-core` test coverage, but I've pretty extensively tested these changes against https://github.com/thefrontside/ui-eholdings.

### DIFF
--- a/webpack/stripes-config-plugin.js
+++ b/webpack/stripes-config-plugin.js
@@ -1,52 +1,17 @@
-const fs = require('fs');
+// This webpack plugin generates a virtual module containing the stripes configuration
+// To access this configuration simply import 'stripes-config' within your JavaScript:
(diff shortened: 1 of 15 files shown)


webpack/stripes-config-plugin.js
```diff
@@ -8,52 +7,54 @@ const serialize = require('serialize-javascript');
 // This webpack plugin generates a virtual module containing the stripes configuration
```
ttamj: Maybe move these comments along to the exported class below, or to the top of the file.  Here its getting mashed up with another comment describing a function.

cherewaty: Comments moved to top of file.

PR #67, opened 2017-10-09
Update dependency on stripes-connect to "latest" tag.

wafschneider:


cherewaty:
Here's my understanding of the strategy we've talked about.

For packages that are so intertwined like `stripes-core` and `stripes-connect`, preferable workflows would be:
- `stripes-connect` has a new major release any time there's a breaking change, or
- they're in a mono-repo using something like Lerna, with lock-step versioning

Without making the workflow changes either of those strategies require, in the mean time we can at least make sure releases of `stripes-core` have only tagged and released dependencies.
- Dependencies stay on tagged versions as often as possible.
- When there is a breaking change to a dependency like `stripes-connect`, `stripes-core` `master` can temporarily point to the commit hash of the dependency.
- When it's time to cut a new release of `stripes-core`, all dependencies should be back on a tagged release (meaning `stripes-connect` would need to have a release first).

I'd rather not rely on `#latest`.

MikeTaylor:
I find this disturbing. Surely it's simpler, cleaner and less error-prone just to release the dependency that has the required new functionality.

wafschneider:
I am inclined to agree with @MikeTaylor on this issue, but I really have no dog in this race. The issue that we are trying to solve is that there are updates to stripes-connect since the pinned commit that have not been exercised by inclusion in the CI-generated stripes bundles for folio-testing, and so @skomorokh is understandably reluctant to release stripes-connect.

We could also try to meet this need by performing some magic in the CI to force it to always pull the latest npm for dependencies, rather than respecting the dependency expressed in package.json, but that seems likely to be brittle and difficult to troubleshoot when things don't go right.

cherewaty:
Replaced by https://github.com/folio-org/stripes-core/pull/70.

Thanks @wafschneider for keeping the conversation moving.

PR #73, opened 2017-10-13
Bundle css for production

cherewaty:
## Purpose
Right now `stripes-core` delivers CSS in the same way in dev and production mode: dynamically loaded `<style>` tags. That's what we want for hot reloading in a dev environment, but in production, we should be delivering CSS files that are separated from the app's JS bundle.

A separate CSS file is good for all kinds of reasons:
- can be loaded in parallel to JS bundle, improving time to first load
- reduces the JS bundle size
- can be cached separately from the JS bundle, meaning changes to JS that don't impact CSS won't trigger a reload of the cached CSS on a user's computer

## Approach
`styles/global.css` was not being consumed by anything, so I removed it.

I moved plugins and loaders around between the base, dev, and prod webpack configs.
#### Base:
- `HtmlWebpackPlugin`
- babel, json, and image loaders
- `bootstrap` loading

#### Dev:
- `HotModuleReplacementPlugin`
- `postcss-loader` with appropriate plugins

#### Prod:
- `ExtractTextPlugin`
- `OptimizeCssAssetsPlugin`
- `postcss-loader` with appropriate plugins, but wrapped in `ExtractTextPlugin.extract()`

This should be a non-breaking change for `stripes-core` consumers, unless there's something in their deployment workflow that would ignore the generated .css file (doubtful).

## Learning
[`ExtractTextPlugin`](https://github.com/webpack-contrib/extract-text-webpack-plugin) was the missing piece. It looks like there was previously an attempt to use it in `webpack.config.cli.dev.js`, but there was nothing being extracted to pipe into the css file.

## Screenshots
What the built index.html served up now looks like:
screen shot 2017-10-13 at 1 43 15 pm


cherewaty:
While testing this, @ttamj ran into a bunch of build errors with `folio-testing-platform`. Most of them are missing semicolons in the css. I'll kick off some work to either put some css linting in place or bulletproof the webpack css processing a little better. We should wait until those fixes are in to merge.

ttamj:
Aside from surfacing those existing css lint errors, the overall approach and organization of the webpack refactor looks great.

MikeTaylor:
Is there really much value in this B&D? Isn't it better just to accept the
semi-colon-less CSS and move on with doing Actual Work?

-- Mike.

cherewaty:
@MikeTaylor Delivering CSS in dynamic style tags is not production-grade. If consistent use of semi-colons is what we need to do to unblock this PR, so be it. The longer we let the sore of missing semi-colons fester, the more difficult it becomes to fix the problem. It will be tedious, but we can put tools like a linter in place to prevent this from ever being a major headache again in the future.

**This _is_ the actual work of building a platform.** Processes like linting, pull requests, and continuous integration decrease the amount of time for teams to build production-ready features using the platform. Unshippable software is a waste of time and money, and putting in place healthy practices to improve collaboration and stability are worth every second.

cherewaty:
@ttamj we should be good to go here, with my latest commit. The `devConfig` was colliding with the `prodConfig`. Missing semi-colons thankfully don't interfere.

PR #76, opened 2017-10-17
Don't lint node_modules. Fixes STCOR-91

zburke:


cherewaty:
I'm curious: what process have you run into that tries to lint `stripes-core`'s `node_modules` directory? I haven't been able to reproduce. `node_modules` should be ignored by default: https://eslint.org/docs/user-guide/configuring#ignoring-files-and-directories

If for whatever reason `node_modules` is not being ignored, I think I'd rather have a `.eslintignore` file instead of making the npm shortcuts more complex.

zburke:
Huh; on my machine, `yarn lint` runs like this: 
```
$ yarn lint
yarn lint v0.27.5
$ eslint ./

/Users/zburke/projects/stripes-core/examples/trivial/node_modules/hoist-non-react-statics/index.js
   5:1   error  Expected newline before "use strict" directive                   lines-around-directive
   5:1   error  'use strict' is unnecessary inside of modules                    strict
...
/Users/zburke/projects/stripes-core/webpack.config.cli.prod.js
  5:41  error  Unable to resolve path to module 'optimize-css-assets-webpack-plugin'  import/no-unresolved

✖ 53182 problems (51378 errors, 1804 warnings)
  41960 errors, 0 warnings potentially fixable with the `--fix` option.

error Command failed with exit code 1.
```
I agree that using a `.eslintignore` file is better than a complicated npm command, but in this case the real question is why we are seeing different results.

zburke:
On closer inspection, the issue is the nested `./examples/trivial/node_modules` directory. I'll create a `.eslintignore` file that addresses that; thanks for the suggestion.

MikeTaylor:
This is a manifestation of the deeper problem that the "trivial" module _really_ doesn't belong in a subdirectory of stripes-core, but in its own repo.

PR #77, opened 2017-10-17
don't lint nested node_modules directory. Fixes STCOR-91

zburke:


zburke:
@MikeTaylor when I looked at the output more carefully, it was linting a nested `node_modules` directory. The default config does indeed ignore `/node_modules/*`, but only relative to the root directory: 
https://eslint.org/docs/user-guide/configuring#ignoring-files-and-directories

PR #78, opened 2017-10-18
Turn on CSS source maps

cherewaty:
## Purpose
Source mapping will make CSS debugging much easier.

## Approach
- Set `sourceMap` to true for each of the webpack style loaders in a dev environment.
- Do not deliver source maps in a production build.

## Learning
https://github.com/postcss/postcss-loader#sourcemap

#### Before
before

#### After
after

### DIFF
--- a/webpack.config.cli.dev.js
+++ b/webpack.config.cli.dev.js
@@ -33,9 +33,17 @@ devConfig.module.rules.push({
   use: [
     {
       loader: 'style-loader',
+      options: {
+        sourceMap: true,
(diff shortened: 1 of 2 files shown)


MikeTaylor:
Regarding minify/ugly:

I would like, if we can, to have it so that folio-testing retains readable source and does source-mapping correctly -- and minification/uglification is done only for folio-demo. That's because we often find ourselves needing to track down the cause of problems that occur on folio-testing but not in development.

PR #81, opened 2017-10-25
Label a tenant's disabled modules on `/about`. Fixes STCOR-69

zburke:
In order to label a tenant's disabled modules, we need to be able to
query Okapi from within stripes-core, but `stripes.connect()`, which is
actually the result of calling `connectFor('@folio/some-module')` for
each dynamically-loaded module, wasn't available within stripes-core.

The approach taken here was to hard-code a call to
`connectFor('@folio/core')` and stash it in the context. The benefit of
using context is that we stash it in one place and every component that
wants it just specifies a static `contextTypes = { coreConnect:
PropTypes.func,}` element. The downside is that connecting a component
in stripes-core is now slightly different than connecting a component in
an app even though the end result is the same.

RootWithIntl had to be refactored into a stateful component because
of how `getChildContext` is implemented; `context` is not avaiable in
child components if the context is defined on a stateless component.

Note: To disable a module for a tenant, issue an empty DELETE request for the module, e.g. `http://localhost:9130/_/proxy/tenants/diku/modules/mod-notes-2.0.1-SNAPSHOT.29`. 

To enable a module for a tenant, issue a POST request to `http://localhost:9130/_/proxy/tenants/diku/modules` with the module-id in the body, e.g. 
```
{
	"id": "mod-notes-2.0.1-SNAPSHOT.29"
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 * Move epics to core. STCOR-82.
 * Move stripes-loader logic into stripes-core as a webpack plugin. STCOR-25.
 * Trivial app uses props.resources, not props.data. Fixes STCOR-92.
+* Label a tenant's disabled modules on `/about`. Fixes STCOR-69.
 
 ## [2.7.0](https://github.com/folio-org/stripes-core/tree/v2.7.0) (2017-09-01)
 [Full Changelog](https://github.com/folio-org/stripes-core/compare/v2.6.1...v2.7.0)
@@ -269,4 +270,3 @@
 
 * Requires v0.0.9 of `stripes-connect`.
 * First version to have a documented change-log. Each subsequent version will describe its differences from the previous one.
-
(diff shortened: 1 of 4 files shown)


MikeTaylor:
> RootWithIntl had to be refactored into a stateful component because of how `getChildContext` is implemented; `context` is not avaiable in child components if the context is defined on a stateless component.

That seems more than a little broken.

Was the switch from functional to stateful style the only change you made to `<RootWithIntl>`?

MikeTaylor:
But I don't really understand why you used context at all. Why not just put the working version of `connect` into the Stripes object, as we do for modules?

MikeTaylor:
More generally, I am inclined to like the adoption of stripes-connect for this.

I _was_ going to say that I worry about circular dependencies, with stripes-core depending on stripes-connect. But it turns out that stripes-connect does not depend on stripes-core, so I think we're good.

zburke:
cf https://github.com/folio-org/stripes-core/pull/82

zburke:
Closing in favor of #82.

PR #82, opened 2017-10-25
Label a tenant's disabled modules on `/about`. Fixes STCOR-69

zburke:
In order to label a tenant's disabled modules, we need to be able to
query Okapi from within stripes-core, but `stripes.connect()`, which is
actually the result of calling `connectFor('@folio/some-module')` for
each dynamically-loaded module, wasn't available within stripes-core.

The approach taken here was to hard-code a call to
`connectFor('@folio/core')` on `RootWithIntl` and to replace the dummy
version of `stripes.connect` with the result of that call, as is done in
`moduleRoutes.js::getModuleRoutes()` for dynamically loaded modules.
This is in contrast to the approach taken in PR #81 where `connect` was
placed in the context rather than directly on the cloned stripes object.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 * Move epics to core. STCOR-82.
 * Move stripes-loader logic into stripes-core as a webpack plugin. STCOR-25.
 * Trivial app uses props.resources, not props.data. Fixes STCOR-92.
+* Label a tenant's disabled modules on `/about`. Fixes STCOR-69.
 
 ## [2.7.0](https://github.com/folio-org/stripes-core/tree/v2.7.0) (2017-09-01)
 [Full Changelog](https://github.com/folio-org/stripes-core/compare/v2.6.1...v2.7.0)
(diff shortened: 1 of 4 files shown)


MikeTaylor:
To my mind, that is more elegant. (One good clue is that it's shorter!) It also allows me to see more clearly the structure of the change that this was in support of, which also looks good.

My vote: merge this, and dump #81.

zburke:
Mike, agreed; I think this is the better approach of the two. I'll wait a bit before merging to give Jason and Michał a chance to weigh in.

mkuklis:
@zburke thank you for spending more time on this. This PR is more aligned with how things are working in modules and how we use `connect` (via `stripes`) already. It's more intuitive and elegant. So I agree with you and Mike here. Nice work!

skomorokh:
Yup, this is good, and consistency is nice. A fine choice. I was thinking of the context as an option to avoid needing to pass this through the whole hierarchy if we mainly needed it here and there closer to the edge, but I think that needs stripes-connect changes to be feasibly convenient.

PR #85, opened 2017-10-27
STCOR-29 move build logic

ttamj:
Stripes.js now focuses on command and config file input only.  This allows build logic to more readily be reused outside of a CLI. Related to STCOR-29 and prep for FOLIO-852

### DIFF
--- a/stripes.js
+++ b/stripes.js
@@ -1,35 +1,11 @@
 #!/usr/bin/env node
 
 const commander = require('commander');
-const webpack = require('webpack');
-const HardSourceWebpackPlugin = require('hard-source-webpack-plugin');
 const path = require('path');
-const nodeObjectHash = require('node-object-hash');
-const express = require('express');
-const webpackDevMiddleware = require('webpack-dev-middleware');
-const webpackHotMiddleware = require('webpack-hot-middleware');
-const connectHistoryApiFallback = require('connect-history-api-fallback');
-
-const StripesConfigPlugin = require('./webpack/stripes-config-plugin');
-
-const cwd = path.resolve();
-const cwdModules = path.join(cwd, 'node_modules');
-const coreModules = path.join(__dirname, 'node_modules');
-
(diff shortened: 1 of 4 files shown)


skomorokh:
Nice, much tidier :+1:

cherewaty:
👏 Nice refactor! We're doing a few modifications to the webpack config in our demo app build: 
- a plugin to intercept all network requests with our `mirage-server` instance https://github.com/thefrontside/ui-eholdings/blob/master/demo/stripes.js#L114
- `eslint-loader`, to show failures at runtime (maybe this should be in `stripes-core`, but that's an issue for another time): https://github.com/thefrontside/ui-eholdings/blob/master/demo/stripes.js#L122

Could there be a way to pass in arbitrary webpack config additions/modifications to `stripes.serve()` and `stripes.build()`?

ttamj:
@cherewaty  Yes! That's my plan. I actually had a hook in there so the calling code could modify the final config, but ended up removing it as I had nothing in this PR that made use of it.  If you could use that now or soon, I'll put it back.

cherewaty:
Awesome! We would use that hook right away.

cherewaty:
@ttamj I'd prefer that `webpackOverrides()` return a new `config` object instead of mutating the one it was given. That makes operations like what we're doing with `svgloader` here easier: https://github.com/thefrontside/ui-eholdings/blob/jc/stripes-rework/demo/stripes.js#L38

I have a version that works with that approach here: https://github.com/thefrontside/ui-eholdings/blob/jc/stripes-rework/demo/stripes.js#L88

It required two changes to `build.js` and `serve.js`:
```
if (options.webpackOverrides && typeof options.webpackOverrides === 'function') {
    config = options.webpackOverrides(config);
}
```
and `config` can't be a `const` (I made it a `let`).

ttamj:
@cherewaty All set. I prefer returning a new object as well.  My work in progress using this hook was only making simple modifications, so thanks for catching that.

PR #86, opened 2017-10-30
Restore resolveLoaders config omitted from previous refactor

ttamj:
This looks to still be necessary for locating loaders in certain build scenarios


MikeTaylor:
@ttamj Since this PR is bugfix, you should just go ahead and merge it yourself -- no need to wait on someone else's opinion.

PR #87, opened 2017-11-01
Add switchMap operator to stripes-core

elrickvm:
## Purpose
When modules such as `eholdings` are loaded into a folio application the epic they receive comes from stripes-core. So if you need access to a method/rxjs-operator off of `action$` such as `action$.ofType().switchMap()` in that module it needs to be in stripes-core so when stripes-core combines its epics the operator will be present.

Without this operator being imported at the `stripes-core` level user of an module such as `eholdings` that makes use of switchMap will receive this error `Uncaught TypeError: action$.ofType(...).filter(...).switchMap is not a function` 

## Approach
Add import statement to bring in rxjs switchMap operator
`import 'rxjs/add/operator/switchMap';`

### DIFF
--- a/src/configureEpics.js
+++ b/src/configureEpics.js
@@ -2,6 +2,7 @@ import { combineEpics, createEpicMiddleware } from 'redux-observable';
 import { BehaviorSubject } from 'rxjs/BehaviorSubject';
 import 'rxjs/add/operator/mergeMap';
 import 'rxjs/add/operator/map';
+import 'rxjs/add/operator/switchMap';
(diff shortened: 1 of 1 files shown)


wwilsman:
Looks like @hjiebsco made this commit already. 😄 https://github.com/folio-org/stripes-core/commit/da62097da373aea379df5a728bc9629d082fd23f

elrickvm:
Sweet. @wwilsman ill close it. Thanks @hjiebsco

PR #92, opened 2017-11-08
Webpack build enhancements

ttamj:
- New `--sourcemap` option for builds (STCOR-64)
- Re-enables javascript minification using the ES6 compatible version of UglifyJS (STRIPES-102)
- Put React in production mode via EnvironmentPlugin (set`NODE_ENV=production` to enable)

### DIFF
--- a/package.json
+++ b/package.json
@@ -102,6 +102,7 @@
     "rxjs": "^5.4.3",
     "serialize-javascript": "^1.4.0",
     "style-loader": "^0.18.2",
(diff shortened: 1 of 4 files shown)


MikeTaylor:
> New `--sourcemap` option for builds (STCOR-64)

Hurrah!

hjiebsco:
Great work. The bundle size decreased from 8.8M to 3.7M, and everything works just fine.

PR #94, opened 2017-11-09
Return promise for build() and serve()

ttamj:
This will facilitate usage in consumers like stripes-cli by knowing when webpack has finished its initial build.


ttamj:
@cherewaty I realize now that this will require a minor update to your copy of stripes.js, from callback to a promise to render the stats.
```
stripes.build(stripesConfig, options, processStats);
```
Also note the slight change to processStats itself, breaking out the error which you can copy as well.

PR #98, opened 2017-11-13
Stripes 452

jeremythuff:
Resolves [STRIPES-452](https://issues.folio.org/browse/STRIPES-452)


src/Root.js
```diff
@@ -12,11 +14,22 @@ import { loadTranslations, checkOkapiSession } from './loginServices';
...
   constructor(...args) {
     super(...args);
     this.reducers = { ...initialReducers };
     this.epics = {};
+
+    for (const app of modules.app) {
+      if (window.location.pathname.startsWith(app.route) && app.queryResource) {
+        // This is not DRY, as it was expressed already in LocalResource is stripes-connect,
+        // And in MainNav.js in stripes-core. Both State Keys should be derived from a common mechanism.
+        this.queryResourceStateKey = `${app.dataKey ? `${app.dataKey}#` : ''}${_.snakeCase(app.module)}_${app.queryResource}`;
```
MikeTaylor: I do agree that this should be factored out to a single location -- presumably a utility function in stripes-connect, which stripes-core can use. I don't think you should hold up merging the present PR while you do that, though. Come back and do it subsequently.

src/components/MainNav/MainNav.js
```diff
@@ -55,12 +60,20 @@ class MainNav extends Component {
...
+    this.unsub = null;
+    this.queryValues = null;
 
     const moduleList = modules.app.concat({
       route: '/settings',
       module: '@folio/x_settings',
     });
 
+    for (const entry of moduleList) {
+      if (props.location.pathname.startsWith(entry.route)) {
```
MikeTaylor: Is this a precise enough test? What if `entry.route` is "foo" and the location is "/foobar"?

src/components/MainNav/MainNav.js
```diff
@@ -86,6 +99,39 @@ class MainNav extends Component {
...
+        const query = location.query ? location.query : queryString.parse(location.search);
+        const allParams = Object.assign({}, query, this.queryValues);
+        const keys = Object.keys(allParams);
+
+        let url = location.pathname;
+
+        if (keys.length) {
+          url += `?${queryString.stringify(allParams)}`;
+        }
+        this.props.history.replace(url);
```
MikeTaylor: Why was `replace` the right choice here rather than `push` as usual?

src/initialReducers.js
```diff
@@ -2,9 +2,12 @@ import { reducer as form } from 'redux-form';
...
-export default {
+const reducers = {
   // TODO: here's where you'd pull in a reducer to handle Okapi actions like auth
   okapi: okapiReducer,
   discovery: discoveryReducer,
   form,
 };
+
+export default reducers;
+
```
MikeTaylor: What is the point to the change in this file?

src/Root.js
```diff
@@ -35,6 +48,13 @@ class Root extends Component {
   }
 
   addReducer = (key, reducer) => {
+    if (this.queryResourceStateKey === key) {
+      const originalReducer = reducer;
+      const initialQueryObject = queryString.parse(window.location.search);
+      // eslint-disable-next-line no-param-reassign
+      reducer = (state = Object.values(initialQueryObject).length ? initialQueryObject : undefined, action) => originalReducer(state, action);
```
MikeTaylor: I'm not sure I understand what's going on here. A comment might not hurt.

skomorokh: Minor quibble: please avoid mutating the parameter here, couldn't you leave `reducer` as your `originalReducer` and create a new name for the changed one?

src/components/MainNav/MainNav.js
```diff
@@ -86,6 +99,39 @@ class MainNav extends Component {
...
   }
 
+  subscribeToQueryChanges(moduleInfo) {
+    if (this.unsub) this.unsub();
+    if (!moduleInfo.queryResource) return;
+    this.unsub = this.store.subscribe(() => {
+      const previousQueryValues = this.queryValues;
+      // This is not DRY, as it was expressed already in LocalResource is stripes-connect,
+      // And in Root.js in stripes-core. Both State Keys should be derived from a common mechanism.
+      this.queryValues = this.store.getState()[`${moduleInfo.dataKey ? `${moduleInfo.dataKey}#` : ''}${_.snakeCase(moduleInfo.module)}_${moduleInfo.queryResource}`];
```
skomorokh: Only apps can surface resources in the query. Apps aren't connected with a dataKey.

src/components/MainNav/MainNav.js
```diff
@@ -86,6 +99,39 @@ class MainNav extends Component {
     this.context.router.history.push('/');
   }
 
+  subscribeToQueryChanges(moduleInfo) {
+    if (this.unsub) this.unsub();
```
skomorokh: Since you're subscribing to all state changes not those of any particular resource, rather than subbing/unsubbing repeatedly couldn't you just have `this.storeKeyToReflectInTheUrlIfAny` and update that as needed based on the `props.location` changes you're currently using to trigger this subscription?

src/components/MainNav/MainNav.js
jeremythuff: In this instance we want the shape of the URL to change, but not trigger any navigation.

PR #99, opened 2017-11-15
Move global css

cherewaty:
Part two of https://github.com/folio-org/stripes-components/pull/89 (that should be merged first).

Globally-scoped CSS would now live in `stripes-components`.

The `height: 100%` set on `#root` is the one exception, since that really is for the `stripes-core` app and not an element found in `stripes-components`.


cherewaty:
@MikeTaylor was there a resulting bug?

MikeTaylor:
No. That is hardly the point.

PR #102, opened 2017-11-17
Support "_path" within anointed stripes-connect resource.

MikeTaylor:
Assignment to special parameter `_path` within the anointed
stripes-connect resource will change the URL path.

Fixes STCOR-105.


MikeTaylor:
@cherewaty I fixed the semicolon but the unused `path` I think is from your commit 15ddfd4153498095b41e851c3f480ce9c53bffcd. It looks like a trivial fix -- delete line 5 -- but I am wary of touching any of the WebPack stuff. Can you please see to this?

cherewaty:
@MikeTaylor yeah, that `path` `require()` can go away, I'm surprised Jenkins didn't spit out a warning on this PR: https://github.com/folio-org/stripes-core/pull/99

Fix: https://github.com/folio-org/stripes-core/pull/104

PR #107, opened 2017-11-17
Only use css variables from stripes-components

cherewaty:
## Purpose
Now that `stripes-components`, with `storybook` tooling, needs to know 100% of the CSS context (explained in detail in https://github.com/folio-org/stripes-components/pull/89), `stripes-core` shouldn't override the css variables set in `stripes-components`.

## Approach
`stripes-core` components using the variables previously found in `stripes-core`'s `variables.css` now import `stripes-components`' `variables.css` instead.

In a brief search of `folio-org` UI modules, I didn't find any other repos attempting to import the `stripes-core` variables. That doesn't mean there are definitely zero cases of that, but this change should be low-risk.

## Someday
It'd be awesome for FOLIO installs to be easily theme-able by tenant - in that case we would want to override the css variables (or some similar strategy, maybe that just relies on the cascade?). Until we reach that point, it's easier for `stripes-components` to completely own the css variables.

## Next steps
As much of the CSS found in `stripes-core` as possible (and its corresponding presentational components) should be moved to `stripes-components`. It'd be awesome to able to iterate on the header, notifications, login form, etc. in the Storybook environment.


skomorokh:
I'm on board with all of this right up until the "Next steps" 

> As much of the CSS found in stripes-core as possible (and its corresponding presentational components) should be moved to stripes-components. It'd be awesome to able to iterate on the header, notifications, login form, etc. in the Storybook environment.

I don't understand this one at all! Surely `stripes-components` is only for components that we forsee reusing, a library to facilitate a design system and provide consistency between apps? If it's something specific to `stripes-core`, then that component should live in `stripes-core`, just like a component only used in `ui-users` should live in `ui-users` and a component only used in `third-party-module-written-by-aliens` should only live there. 

Instead, perhaps we can do something with `stripes-cli` to make it easy to use storybook with app-specific components rather than try and put every component from the whole system in one repo?

cherewaty:
@skomorokh As I understand it, the current mission of `stripes-components` is to provide reusable components for Stripes applications to use. Adding Storybook tooling, moving the global CSS, etc. pushes `stripes-components` towards being more than just a component library; it is becoming a full-fledged design system.

The navigation, login screen, notifications, etc. feel like an essential part of a FOLIO design system. They are the most visible parts of the ecosystem. Even though they won't be used by developers outside of `stripes-core`, they're key for designers and developers to understand the world they're building in.

It makes sense to me that UI developers iterating on the look-and-feel of those critical FOLIO components would do so in the same way as any component currently in `stripes-components`, and that the whole design system is versioned together (not spread over two repositories).

JohnC-80:
The priority users for the storybook/design system are developers and designers who want to build new FOLIO **modules**. I don't see the helpfulness of including non-reusable components in the storybook resource/stripes-components repo at this point. Static documentation can cover configuration for modules and how they interact with the core systems - but actually showing those components to module builders doesn't seem helpful since they cannot actually be used within a module. System-level components should remain in stripes core.

MikeTaylor:
For what it's worth, I'm with @JohnC-80 on this one.

PR #119, opened 2017-11-30
UIEH-107 ie11

ttamj:
The virtual module, `stripes-config`, created by stripes-config-plugin does not appear to pass through Babel and therefore the arrow functions for each `getModule()` are left in the code which IE cannot deal with. 

The solution using `new Function()` appears to negate the need for `eval()` due to the similar nature in parsing.   For reference, `eval()` did not want to accept a function expression without the arrow syntax, thus the use of `new Function()` constructor.  Such parsing is needed with `getModule` for feeding app entry points to webpack via require.  This approach may change when we introduce code-splitting.

This issue could also be solved by determining how to pass the `stripes-config` virtual module through the Babel loader.  However, adjusting the one line in stripes-config-plugin was quick to verify and the remainder of `stripes-config` has nothing to transpile at the moment.  Still, transpilation is perhaps the better long term solution and should eventually be addressed particularly if we ever find a need for creating multiple virtual modules.

### DIFF
--- a/index.html
+++ b/index.html
@@ -9,5 +9,4 @@
     <div id="root">
     </div>
   </body>
-</body>
 </html>
--- a/webpack/stripes-config-plugin.js
+++ b/webpack/stripes-config-plugin.js
@@ -36,7 +36,7 @@ function parseStripesModules(enabledModules, context, alias) {
     const { stripes, description, version } = loadDefaults(context, moduleName, alias);
     const stripeConfig = Object.assign({}, stripes, moduleConfig, {
       module: moduleName,
-      getModule: eval(`() => require('${moduleName}').default`), // eslint-disable-line no-eval
+      getModule: eval(`new Function([], "return require('${moduleName}').default;")`), // eslint-disable-line no-eval
       description,
       version,
     });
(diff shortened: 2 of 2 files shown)


skomorokh:
Nice! I mean, it's still roughly equivalent in operation (and sketchiness) to the eval(), but does the trick. As could rephrasing it not to be an arrow function (though maybe the `this` binding matters here and this syntax preserves it?)

However, I'm concerned that you're spending time on IE11 compatibility. Do any beta libraries mandate that as their workstation browser in 2018? Do we anticipate the broader community of academic libraries will typically have it as their standard in 2019?

I can't unilaterally decide this for FOLIO (nor would I seek that level of responsibility!) but perhaps we should get clarification on what our target browsers are for alpha, beta, and v1 (with the understanding that including defunct, less-standard browsers like IE11 has an associated overhead).

ttamj:
Good question on `this`.  Given `getModule` only returns a require(), I don't think the binding is of a concern.

Re: browsers, This PR is not attempting to bless IE11 and I don't know of an official browser support list yet.  Folio simply didn't load at all and our current Babel configuration suggests IE11 ought to at least try.  While IE11 now loads without syntax error, I have not validated IE beyond that.

PR #127, opened 2017-12-11
Updated about page

rasmuswoelk:
Fixed styling by adding <List>-component for lists and <Headline> for all headlines.

### DIFF
--- a/src/components/About.js
+++ b/src/components/About.js
@@ -11,6 +11,8 @@ import stripesLogger from '@folio/stripes-logger/package.json';
 /* eslint-enable */
 
 import Pane from '@folio/stripes-components/lib/Pane';
+import Headline from '@folio/stripes-components/lib/Headline';
+import List from '@folio/stripes-components/lib/List';
 import Paneset from '@folio/stripes-components/lib/Paneset';
 import { isVersionCompatible } from '../discoverServices';
 import AboutEnabledModules from './AboutEnabledModules';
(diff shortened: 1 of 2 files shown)


MikeTaylor:
@rasmuswoelk Are you awaiting a code review or something before you merge this?

src/components/About.js
```diff
@@ -72,7 +80,7 @@ const About = (props) => {
   return (
     <Paneset>
       <Pane defaultWidth="30%" paneTitle="User interface">
-        <h4>Foundation</h4>
+        <Headline>Foundation</Headline>
```
ttamj: I see <Headline> defaults to an <h2> and these instances are replacing <h4>'s.  I haven't looked into the current structure of the about page, so it may not make a difference, but is switching to <h2> here intentional?

src/components/About.js
```diff
@@ -72,7 +80,7 @@ const About = (props) => {
   return (
     <Paneset>
```
cherewaty: The About page should probably be columns that stack vertically at smaller sizes instead of the responsive behavior of `Paneset`s, but we can tackle that in a future pull request.

MikeTaylor:
What is preventing us from getting this merged? I am very very ready for the About page to no longer be a maze of twisty little module IDs, all different.

MikeTaylor:
In fact, you know what? I'll just do it.

MikeTaylor:
Done, glory be!
screen shot 2017-12-13 at 00 18 45

PR #141, opened 2018-01-03
Allow more flexible symlinking of modules

skomorokh:


webpack/stripes-config-plugin.js
```diff
@@ -14,7 +14,19 @@ function loadDefaults(context, moduleName, alias) {
   if (alias[moduleName]) {
     aPath = require.resolve(path.join(alias[moduleName], 'package.json'));
   } else {
-    aPath = require.resolve(path.join(context, 'node_modules', moduleName, '/package.json'));
+    try {
+      aPath = require.resolve(path.join(context, 'node_modules', moduleName, '/package.json'));
+    } catch (e) {
+      try {
+        // The above resolution is overspecific and prevents some use cases eg. yarn workspaces
+        aPath = require.resolve(path.join(moduleName, '/package.json'));
```
ttamj: I haven't run into this symlinking scenario, but the logic looks good.  Although serving a different purpose, the tries here remind me of the generate-stripes-alias.js logic.  Do you think generate-stripes-alias also needs to be concerned with the yarn workspace use-case?

skomorokh: With workspaces, the node_modules dir for the platform is in its parent directory. The default resolution logic (./node_modules, ../node_modules, ../../node_modules all the way up) handles that well enough for react (currently the only application?) since if it's not in core it's because core is installed via npm at which point it's in the workspace's hierarchy.

generate-stripes-alias.js seems like it should be adequate for that case because it falls back to straight up attempting to resolve it. But if it's part of a more general alias system like in stripes-cli it would be good to also check the parent of the platform dir if you anticipate something outside the platform's hierarchy needing to resolve something brought in there.

PR #144, opened 2018-01-09
Updates to login page

rasmuswoelk:
- Styling changes to input fields, buttons, error message and logo
- Button is now only disabled if the username field is empty. Before it was both username and password. Due to a autofill bug in Chrome (might be in other browsers as well) the button was disabled until you clicked the page when both fields was autofilled. The solution to this is pretty dirty so I left it at just checking for a username value before removing disabled state from the button.
- Using SubmissionError from the redux docs on login failure which ensures that the submitFailed prop works as suspected (it was true no matter what you did before since the login request returns “success” in the promise). More on this: https://redux-form.com/7.2.0/docs/api/submissionerror.md/
- Removed "initialValues" since they weren't being used


cherewaty:
Just tried this out. Because the username is getting cleared out on an error, https://github.com/folio-org/stripes-core/pull/144/files#diff-487c09e67eb888ca5f4a9c7c64ea3ad1R56 evaluates to false after receiving an `SubmissionError`. The form then can't be resubmitted without refreshing the page.

I wish we had some acceptance test coverage on this, would make refactors a lot smoother :(

rasmuswoelk:
@cherewaty Ah. I thought I tested it thoroughly but I clearly didn't. I will fix this asap.

And yes, some tests would probably be a good idea..

rasmuswoelk:
@cherewaty It should work as expected now. 

After discussing with @mkuklis we decided to remove some actions (see latest commit) from the clearOkapiSession method since they seem to be unnecessary to dispatch when logging in anyway. Removing these and only resetting the password instead of the entire form on login failure makes the login (redux-)form work as expected.

I have tested it but It would be great if you could give it a test run as well 👍

rasmuswoelk:
@cherewaty Great! I guess we are ready to merge? (I don't have privileges)

PR #153, opened 2018-01-15
Allow use of Stripes without Okapi

skomorokh:
If you don't provide an Okapi section in your Stripes configuration this allows Stripes to continue building. Useful for demoing with the trivial app or building a fully FOLIO front-end that can also work standalone with the same code base. 

Currently requires hasAllPerms and disableAuth in config as we haven't made auth modular yet.


skomorokh:
Urk! Betrayed by checkmarks! I'm going to get some lunch, and then I will fix this. Apologies.

PR #155, opened 2018-01-15
Make About component responsive to narrow screens

cherewaty:
## Purpose
Two thirds of the content on the software versions (`/settings/about`) page is inaccessible on smaller screens.

## Approach
`Paneset`s are intended for layered information with relationships, but the content of the software versions page is not layered. The user interface, Okapi services, and UI/service dependencies sections are all siblings, not parent/children.

Instead we can use flexbox column CSS to lay out the page. At small sizes, the content will all be in one column. Once the screen is wide enough (wider than the `largeUp` breakpoint), we can expand out to three columns.

To consistently nest components with their stylesheets, the `About` and `AboutEnabledModules` components are now inside of an `About` directory.

#### Narrower screen
localhost_3000_settings_about iphone 7 1

#### Wider screen
localhost_3000_settings_about 1

#### Narrower screen
localhost_3000_settings_about iphone 7

#### Wider screen
localhost_3000_settings_about


cherewaty:
Follow-up TODO (separate piece of work): get rid of the `border-right` on the last `Pane`.

PR #159, opened 2018-01-18
Upgrade localforage

cherewaty:
Ever seen a warning like this when `yarn`ing a module using `stripes-core`?

screen shot 2018-01-18 at 3 29 21 pm

`localforage` has a new release that fixes it: https://github.com/localForage/localForage/pull/775


cherewaty:
I guess the build is failing because of Jenkins' caching of the NPM registry? I've been blocked by that before. Anything we can change there?

PR #160, opened 2018-01-19
Remove extraneous console.log

MikeTaylor:


cherewaty:
If CI failed the build on lint failures (assuming GitHub doesn't allow merging with a broken CI build), the extraneous `console.log()` would never have made it to `master`.

MikeTaylor:
I think the CI build _does_ fail on ESLint failures -- it certainly does for most of the Stripes modules, I've had this happen to me multiple times. The issue here is that we don't have ESLint configured to reject `console.log`: see https://github.com/folio-org/eslint-config-stripes/blob/master/index.js#L30

I'd be happy for that to change, if you want to change it. I think legitimate cases of logging to the console are rare enough that decorating them with `// eslint-disable` is reasonable.

cherewaty:
I'll check on the console settings in `eslint-config-stripes`. I agree that the use of them should be infrequent enough that disabling those lines is fine.

PR #161, opened 2018-01-19
Update babel loader test condition

ttamj:
Our fork of `react-githubish-mentions` is under the `@folio` scope and therefore stripes-core attempts to transpile it based on the current babel-loader test condition.  This results in a build error for a missing es2015 preset which Stripes does not use.  

This fix updates the current babel loader test condition to exclude specific folio-scoped modules from selection.  Tests were added before this change to ensure we continue to capture the various desired use cases.

Fixes STRIPES-499.


cherewaty:
Thanks for the good tests! 💯

PR #165, opened 2018-01-23
Passing missing id prop to NavButton

rasmuswoelk:
The id's were missing on nav buttons as mentioned here: https://issues.folio.org/browse/STCOR-132

This fixes that issue by padding the prop down to the button element.

I also added so that the element is changed to an anchor tag if a href-prop is provided. This enables right-click to open in new tab/window.


cledvina:
This works for me-- I can confirm that the ids are back!

MikeTaylor:
Woo-hoo! Stand by for a passed integration-test!

-- Mike.

PR #176, opened 2018-01-30
Removed href from the selected NavButton

rasmuswoelk:
This is the correct implementation/fix of this issue: https://issues.folio.org/browse/STRIPES-498


wwelling:
You mean as opposed to https://github.com/folio-org/stripes-form/pull/10/files?

wwelling:
I am curious, how was the previous fix not working? Can you provide steps to reproduce the incorrect behavior of the previous fix?  Would like to get a better understanding of the codebase and design.

PR #178, opened 2018-01-31
STCOR-49 translation files

ttamj:
This adds support to load module translations from language specific `*.json` files from a module's own `translations` directory.  Backwards compatibility with the `package.json` method is maintained while apps are migrated over.

Translation keys are merged and prefixed at build-time.  Merged translation files are emitted to the build's output directory where they are fetched from the browser during loginServices's `loadTranslations()`.

An optional languages filter is now available to limit the number of translations included in a build. When set, the filter will also limit the number of third-party locale resources included by `react-intl` and `moment` reducing build file quantity and size.

For example:
```
config: {
  languages: ['en', 'es'],
}
```
Implements STCOR-49


MikeTaylor:
stripes-core v2.9.0 will be forthcoming today: see STCOR-138.

MikeTaylor:
Related question: can one translation pull in the results of another? For example, if we have:
```
{
   objectSingle: "index",
   objectPlural: "indices",
}
```
Is there a way to use one of other of those in other translations, such as:
```
resultCount: "{count, number} {count, plural, one {{objectSingle} found} other {{objectPlural} found}}"
```
?

ttamj:
I haven't looked into that, so I am not sure if its possible.  I would not recommend it though. My initial concern with this use-case is anytime independently translated content is spliced together, there is a risk of generating a less than optimal, or even incorrect, result.

In one language translations A and B might equally pull in the result of translation C.  However in another language, the result for C may not be the word or form of choice appropriate for A or B.  In this case, when C gets translated, its intended context would not be known.

MikeTaylor:
Interesting. I will get @filipjakobsen to comment on this, since it's his desire to include as module metadata the object-type in both singular in plural forms that led me to think of doing this.

filipjakobsen:
I don't fully understand what the alternative is to gathering it within the app bundle? Is the alternative to just have unknowable string names that describe the same thing in different apps?

ttamj:
@filipjakobsen Is there a related Jira for the use-case in question? I'd like to continue the discussion there as the conversation is independent of this code being merged.

PR #188, opened 2018-02-05
Improve button style consistency 

cherewaty:
Follow-up to https://github.com/folio-org/stripes-components/pull/209.

1. `transparent` is no longer an available button style; replaced with `link`
2. Now that `primary` is no longer the default button style (now `default` is), the primary action buttons on login pages needed that style applied.

### DIFF
--- a/src/components/Login/Login.js
+++ b/src/components/Login/Login.js
@@ -91,7 +91,7 @@ class Login extends Component {
               />
             </div>
             <div className={authFormStyles.formGroup}>
-              <Button id="clickable-login" type="submit" buttonClass={authFormStyles.submitButton} disabled={buttonDisabled} fullWidth marginBottom0>
+              <Button buttonStyle="primary" id="clickable-login" type="submit" buttonClass={authFormStyles.submitButton} disabled={buttonDisabled} fullWidth marginBottom0>
                 {buttonLabel}
               </Button>
             </div>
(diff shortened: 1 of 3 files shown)


JohnC-80:
Hmm... does `link` seem like a better API choice than `transparent`? I think `<IconButton>` is the way to go where `transparent` was used previously, as the bulk of `transparent` buttons just housed an `<Icon>` - thoughts, @rasmuswoelk?

JohnC-80:
@cherewaty any issue with renaming the css class `.default` to `.secondary`? Seems that would effectively unbreak the `<Buttons>` that used `buttonStyle="secondary"`

cherewaty:
I haven't encountered a use of `transparent` that wasn't an icon button (but not using `<IconButton>`, so it totally makes sense to me to not make `link` available anymore either, and update the instances I came across to use `<IconButton>` instead.

cherewaty:
@JohnC-80 why I'm drawn to the `default` naming instead of `secondary`:

> The go-to button style is `default`, and if you don't specify a style for a `<Button>`, it will be `default`.

makes way more sense to me than:

> The go-to button style is `secondary`, and if you don't specify a style for a `<Button>`, it will be `secondary`.

I only found nine uses in the `folio-org` of the `secondary` button style, so I'm not worried about breaking that API. I could have pull requests open to fix all those very quickly.

cherewaty:
Update: I removed `link` from storybook, so as to discourage its use. I recommend otherwise leaving it alone for now, giving us a chance to update uses of it to `<IconButton>`s.

JohnC-80:
*Sigh... the recursive problem that didn't exist when `primary` was the default you'd never have to see and `secondary` was the purposefully-applied option when differentiating was the considered minority use-case... eh, all good for now with the changes, I suppose - bigger fish to fry than button styling...

PR #195, opened 2018-02-07
minor rearrangement of build steps. Refs STCOR-140

zburke:
There are 10 important changes here:

(0)  Use [yarn workspaces](https://yarnpkg.com/lang/en/docs/workspaces/) to manage shared third party dependencies.
(1)  `yarn install` before stripes-cli's `npm install -g`
(10) new `configure` script handles the above, and more

Using workspaces allows yarn to do a better job de-duping dependencies
by lifting a shared node_modules directory up one level. This is both
faster and more space efficient because there are many fewer copies of
things to maintain.

Running `yarn install` before installing stripes-cli seems to yield more
consistently successful results. It isn't clear that installing
stripes-cli first first is incorrect, but sometimes doing things in that
order fails and we don't know why. Running the `yarn install` first
doesn't seem to fail, so here he are.

The configure script is just a TL;DR version of the above that runs all
this stuff in one shot once you've cloned stripes-core, including
configuration of the yarn workspaces and the stripes-cli aliases for
local modules.


ttamj:
Perhaps for the general audience, we omit the `npm install -g `of stripes-cli and use `yarn global add @folio/stripes-cli` instead (no need to clone the CLI either).  The npm variation is only necessary for _developing_ the CLI as yarn doesn't appear to have an equivalent. I think the target audience (being future stripes UI developer) would primarily use the CLI as a tool, rather than something to develop.  Sticking with the simpler setup, and using one package manager, might be desired in that case.

The CLI will automatically notify users of new versions so keeping it up to date could easily be out of band from pulling stripes-* source. I've also been thinking of writing a CLI developers guide, which could contain all the necessary steps to adding a new command, etc, along with the `npm install -g` oddity.

MikeTaylor:
That is an excellent point.

@zburke since this is your PR, do you want to make that change?

zburke:
ATM, `yarn global add @folio/stripes-cli` requires using the `npm-folioci` registry. Is it stable enough to be published generally? If not, we'll need to include instructions about setting the registry as well. 

Alternatively, do we _want_ devs to use that registry anyway, e.g. so they get tip-of-master versions when running tests?

MikeTaylor:
Two things.

1. We need to get past our fear of releasing things, ui-inventory included as well as stripes-cli. Yes, we absolutely should release the latter.

2. Good call on using folioci consistently in development. I can't think of a reason not to -- can you?

ttamj:
I prefer npmfolio-ci for development.  However, that is influenced in large part by the (current) length between releases.  The one drawback is the initial patch versioning behavior (of a `x.x.0` release) which I think could yield unexpected results switching between npmfolio and npmfolio-ci.

I agree the CLI should be released (STCLI-12).  I believe the last thing I wanted to address was updating the command itself, from `stripescli` to `stripes` to avoid such an obvious breaking change out of the gate.  Also on my prerequisites, was pulling in a proper (framework only) copy of ui-testing, though that may be best to defer anyway given recent movement in testing.

MikeTaylor:
I agree that renaming to `stripes` is the way to go. Of course it's a CLI command, you invoke it from the CLI. It's as though Unix provided you with `lscli`, `grepcli`, etc :-)

zburke:
The most-recent commit addresses a few things: 

# specifically configures the `@folio` NPM registry to use npm-folioci to better align with the CI environment
# installs `stripes-cli` from NPM, rather than from source
# discuss yarn workspaces in the doc, and use them in `pull-stripes -b` instead of looping through the repos and running `yarn install` for each one
# corrected errors in the summary

Good to go?

PR #199, opened 2018-02-09
Upgrade to @folio/eslint-config-stripes

doytch:


.eslintrc
```diff
@@ -1,4 +1,7 @@
 {
-  "extends": "stripes",
-  "parser": "babel-eslint"
+  "extends": "@folio/eslint-config-stripes",
+  "parser": "babel-eslint",
+  "rules": {
+    "no-console": "off"
```
ttamj: Disabling this one could probably apply just to /util, /webpack, and maybe /test.  Everything else is in the browser and so it might still be good to catch any stray consoles there.

doytch: That's what I started out with but then there are also .logs in `stripes.js` so I just took a sledgehammer to it. We could just `/* eslint-disable */` in that file though.

ttamj: Oh, right and that's in the root...  I like your eslint-disable suggestion.  There are only a couple in there.

PR #205, opened 2018-02-12
Add emoji to pull-stripes to more easily see warnings

doytch:
When there's a bunch of output from this script it's sometimes hard to see the "not doing anything" warnings at a glance. Hopefully this helps others like it's helped me. 😄

### DIFF
--- a/util/pull-stripes
+++ b/util/pull-stripes
@@ -1,6 +1,11 @@
 clone=n
 build=n
 dir=.
+warning="⚠️ "
+
+if [ "$TERM" = dumb ]; then
+    warning="***"
+fi
 
 if [ "x$1" = x-c ]; then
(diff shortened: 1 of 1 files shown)


MikeTaylor:
For me, this is significantly _less_ visible. That's because I run everything under `M-x compile` in Emacs, and the tiny emoji is hard to spot in monochrome. Also, I can no longer search for "***", and I don't know a way to type the name of the emoji that I have to search for instead.

If you want this, I suggest you tweak it to retain the old "***" when TERM=dumb (as is the case in an Emacs compilation window) and only use the emoji when in a more intelligent terminal that can render it in nice, visible colour.

doytch:
@MikeTaylor good point!

PR #206, opened 2018-02-12
STCOR-159 Render current app as h1, setup default display

JohnC-80:
As part of a continuing effort to improve the accessibility of STRIPES UI, this PR is an initial step in giving FOLIO modules a mark-up structure that outlines nicely so that assistive technologies can have a clean way to navigate the page via its heading tags.
Ideally, pages have a structure that flows in a hierarchical fashion - 
```
<h1>
  <h2>
    <!-- h2 content -->
    <h3>
      <!-- h3 content -->
  <h2>
    <!-- h2 content -->
    <h3>
      <!-- h3 content -->
      <h4>
        <!-- h4 content -->
  <h2>
  <!-- h2 content -->
```
It's best for pages to have only a single `<h1>` tag followed by `<h2-6>` tags arranged as seen in the example.
This GREATLY eases navigation for those using assistive technologies to perceive page content (and is a low-hanging fruit on the accessibility tree that sites can sometimes miss.)
There will be future PR's in other modules to set up a hierarchical flow to the headers used in an app. Much can be handled at the level of `stripes-components` with its `<Accordion>` and `<Headline>` components.

Before a module is selected in the UI, the `<CurrentApp>` displays "FOLIO" with the title "FOLIO platform"

### DIFF
--- a/src/components/MainNav/CurrentApp/CurrentApp.js
+++ b/src/components/MainNav/CurrentApp/CurrentApp.js
@@ -4,22 +4,47 @@
 
 import React from 'react';
 import PropTypes from 'prop-types';
-import NavButton from '../NavButton';
+import classNames from 'classnames';
+import Link from 'react-router-dom/Link';
+import AppIcon from '@folio/stripes-components/lib/AppIcon';
+import Badge from '@folio/stripes-components/lib/Badge';
+import Headline from '@folio/stripes-components/lib/Headline';
+import css from './CurrentApp.css';
 
 const propTypes = {
-  currentApp: PropTypes.object,
+  currentApp: PropTypes.shape(
+    {
+      displayName: PropTypes.string, 
+      description: PropTypes.string
+    }),
+  id: PropTypes.string,
+  icon: PropTypes.oneOfType([
+    PropTypes.element,
+  ]),
+  badge: PropTypes.oneOfType([
+    PropTypes.string,
+    PropTypes.number,
+  ]),
 };
 
-const CurrentApp = ({ currentApp }) => {
-  const { displayName, description } = currentApp;
+const defaultProps = {
+  currentApp: {displayName: 'FOLIO', description: 'FOLIO platform'}
+}
+
+const CurrentApp = ({ currentApp, icon, badge, id }) => {
+
+  const displayIcon = (<div className={css.icon}>{icon || <AppIcon focusable={false} />}</div>);
+
   return (
-    <NavButton
-      label={displayName}
-      title={description}
-    />
+    <div id={id} title={currentApp.description} className={css.currentApp}>
+        { badge && (<Badge color="red" className={css.badge}>{badge}</Badge>) }
+        { displayIcon }
+        <Headline tag="h1" size="small" style={{margin: 0}} >{currentApp.displayName}</Headline>
+    </div>
   );
 };
 
 CurrentApp.propTypes = propTypes;
+CurrentApp.defaultProps = defaultProps;
 
 export default CurrentApp;
--- a/src/components/MainNav/CurrentApp/CurrentApp.js
+++ b/src/components/MainNav/CurrentApp/CurrentApp.js
@@ -37,7 +37,7 @@ const CurrentApp = ({ currentApp, icon, badge, id }) => {
     <div id={id} title={currentApp.description} className={css.currentApp}>
       {badge && (<Badge color="red" className={css.badge}>{badge}</Badge>)}
       {displayIcon}
-      <Headline tag="h1" size="small" style={{ margin: 0 }} >{currentApp.displayName}</Headline>
+      <Headline tag="h1" size="small" margin="none" >{currentApp.displayName}</Headline>
     </div>
   );
 };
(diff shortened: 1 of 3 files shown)


JohnC-80:
In the future I think we'll see a need for better ideas in the way that Settings is handled... Ideally, we don't add many (any) more non-module module-ish UI's - but I'm wondering if Notifications will get there at some point.

src/components/MainNav/CurrentApp/CurrentApp.js
```diff
@@ -4,22 +4,45 @@
...
+
   return (
-    <NavButton
-      label={displayName}
-      title={description}
-    />
+    <div id={id} title={currentApp.description} className={css.currentApp}>
+      {badge && (<Badge color="red" className={css.badge}>{badge}</Badge>)}
+      {displayIcon}
+      <Headline tag="h1" size="small" style={{ margin: 0 }} >{currentApp.displayName}</Headline>
```
rasmuswoelk: There is a margin prop for the Headline-component that can be set to "none" instead of using inline styles. I just realized that the option isn't documented in the readme of the Headline-component so I'll update that in stripes-components.

src/components/MainNav/CurrentApp/CurrentApp.js
```diff
@@ -4,22 +4,45 @@
...
+  ]),
+  badge: PropTypes.oneOfType([
+    PropTypes.string,
+    PropTypes.number,
+  ]),
 };
 
-const CurrentApp = ({ currentApp }) => {
-  const { displayName, description } = currentApp;
+const defaultProps = {
```
rasmuswoelk: Instead of having "Folio" as default it should be the last open app as default (or the first app in the list) meaning that you shouldn't be able to _not_ have any open apps.

I think we can leave it like this for now since the "last open app" feature isn't made yet. The last open app should be stored somewhere - probably on each individual user. It probably needs it's own issue I assume.

src/components/MainNav/CurrentApp/CurrentApp.js
```diff
@@ -4,22 +4,45 @@
...
+  currentApp: { displayName: 'FOLIO', description: 'FOLIO platform' },
+};
```
rasmuswoelk: Instead of having "Folio" as default it should be the last open app as default (or the first app in the list) meaning that you shouldn't be able to _not_ have any open apps.

I think we can leave it like this for now since the "last open app" feature isn't made yet. The last open app should be stored somewhere - probably on each individual user. It probably needs it's own issue I assume.

JohnC-80: What is supposed to happen when a user logs into FOLIO for the very first time? I imagine that once the 'last open app' feature is implemented, that app will actually be open on log-in, and thus will send its info to this component's currentApp prop.

src/components/MainNav/CurrentApp/CurrentApp.js
JohnC-80: Alright.. will update.

src/components/MainNav/CurrentApp/CurrentApp.js
rasmuswoelk: That's a good question! I have asked Filip for more info but if you check out the sign up prototype: http://ux.folio.org/prototype/en/sign-up

Once you are logged in for the first time it looks like it will send you to the Users-app.

JohnC-80: Given the current state of things with currentApp coming through as undefined at the start, I feel that static, self-supplied defaults are best for right now. Better than this component reaching outside of itself and leaning on something, or the consequence of there *not being an <h1> tag on the page if it doesn't render. If there are better ideas for acceptable static defaults, perhaps even a better solution to the lack-of-h1 issue, I'm all ears.

ttamj: @rasmuswoelk @JohnC-80  I'd be interested in learning about the first time login use-case as well.  While the home page is sparse in current form, surely there is a case for not having an app open.  This would be the perfect place for dashboard metrics and KPI's.

rasmuswoelk: @ttamj I'll talk to the designers and hear what they have to say. I heard about one possibility which is to have a "Welcome"-app of some sorts that is shown to the user the first time they log on to the system but I'm not sure if that's the final solution.

rasmuswoelk: @filipjakobsen You probably have more info on this matter.

filipjakobsen: @ttamj Hi Matt,

The need for KPIs etc. will be fulfilled by apps focusing on that. So instead of forcing everyone to use a certain dashboard that we guess will work (and which we would have to maintain centrally), people can make the app they want as a dashboard the app that will open when they open the system. The "To-do" app we are designing will fit this purpose for some users, while a Workflows Dashboard might for others. Still others might want a dashboard app that pulls in widgets from all the different apps, but this would require a complex UI that would function best as a separate app from my point of view.

I think we should simply turn the welcome screen into a "Welcome to FOLIO" app that can in time contain an overview of where to find information, etc. And then have that open when users first log in. It would fit the logic of the general system, and not require any exception logic for when users first log in.

Let me know if you have any questions or thoughts on this.

PR #207, opened 2018-02-14
Lock in uglify-es version because of bug on 3.3.10

cherewaty:
`@folio/ui-eholdings` is failing to create a production build in CI:

screen shot 2018-02-14 at 11 07 12 am

There's a bug in `uglify-es` causing the failure (https://github.com/mishoo/UglifyJS2/issues/2896), and a new version hasn't been cut yet. This will lock in `uglify-es` (needed by `uglify-webpack-plugin`) to 3.3.9, before the bug was introduced.

### DIFF
--- a/package.json
+++ b/package.json
@@ -114,6 +114,7 @@
     "serialize-javascript": "^1.4.0",
     "style-loader": "^0.18.2",
     "typeface-source-sans-pro": "0.0.43",
+    "uglify-es": "3.3.9",
     "uglifyjs-webpack-plugin": "^1.1.8",
     "uuid": "^3.0.0",
     "webpack": "^3.4.1",


ttamj:
FYI, See STRIPES-504.  I should have added you as a watcher on that one...

Since this is not our direct dependency, we ended up using defining it `resolutions` for the platform:

```
"resolutions": {
    "**/uglify-es": "3.3.9"
  }
```

Stripes-core may be the better place to apply it though, so it applies to all builds.

cherewaty:
Ah cool, thanks @ttamj. I'll immediately put that `resolutions` pattern in `ui-eholdings`. Do you think I should I change this PR to use that too (instead of directly defining the dependency)?

ttamj:
I think using `resolutions` will make it more obvious that we have something to back out later.  I assume defining it in stripes-core would allow it to propagate up to the CLI or platform using stripes-core.  However, I haven't tested it.

cherewaty:
Just tried adding the `resolutions` to `stripes-core`'s `package.json` - when `ui-eholdings` pointed to that `stripes-core` branch, it didn't pick up the resolution. That makes sense, I guess - only the end app should be able to define its resolutions - a dependency shouldn't be able to do that for you.

PR #218, opened 2018-02-21
Adding app icons etc.

rasmuswoelk:
This update updates the MainNav so that it displays the app icons instead of placeholders. I have been adding app icons to various apps so we should be ready to go.

Once this PR in stripes-core and this PR: https://github.com/folio-org/stripes-components/pull/244 is merged in we should have app icons in the MainNav. (The stripes-components PR needs to be merged first).

image

**In this update I also:**
- Added the new metadata in the stripes-object
- Added stripes in MainNav context so that <AppIcon /> can utilize it
- Added icon for Settings which isn't an app yet but should be in the future. This settings icon solution is temporary until Settings is a standalone app.


rasmuswoelk:
Note: I am currently not able to run `yarn run lint` properly on my local machine. I'm looking into fixing this issue.

Meanwhile I'll use the response from Jenkins to fix the linting.

PR #224, opened 2018-02-23
add curl command

rjruizes:
`curl` is much faster than installing Ruby and the Okapi CLI


ttamj:
I haven't had a chance to update the docs yet, but the CLI automates this whole process.

rjruizes:
Ah, ok. Mike Taylor pointed me at this doc to enable permissions for UI-trivial

MikeTaylor:
Oh, yes, of course -- sorry, @ttamj, I should have remembered that! You should feel free to replace this whole document with "use `stripes addperm foo`" or whatever it is.

PR #227, opened 2018-02-26
add dash-ell support to pull-stripes. STCOR-166. 

zburke:
add dash-ell support to pull-stripes instead of always iterating over a hard-coded list of modules. Fixes [STCOR-166](https://issues.folio.org/browse/STCOR-166).


ttamj:
You're too fast!  I was working on the same thing right now, bringing pull-stripes into the CLI.

ttamj:
@MikeTaylor Within the CLI, code like this can take advantage of shared logic with other CLI commands that are aware of all the modules (or aliased modules) in the current Stripes environment.  Also the command becomes accessible cross-platform for those on Windows.

MikeTaylor:
That makes sense.

OK, let's merge this one; then you can move forward with the stripes-CLI reimplementation in your own time.

PR #232, opened 2018-02-28
STCOR-170 StripesBuildError with Yarn workspaces

ttamj:
Add another entry to the tryPath list to help locate modules in a yarn workspace
Fixes STCOR-170


ttamj:
@skomorokh FYI - I found I needed this additional check for the example described in STCOR-170.

PR #248, opened 2018-03-23
set importLoaders property in webpack config

JohnC-80:
The `importLoaders` property makes css-loader use the previous loaders in the webpack config's build chain when handling css files that are brought in within css files (via `@import` or use of the `compose` functionality between files)
This will allow for greater style re-use among components, since centralized style sheets can use some of the nice features (nesting, css-variables, etc) that `postcss` gives us. Component styles can use the `composes` command for cross-file style composition as follows.
```
.newTextFieldClass {
  composes: sharedTextFieldClass from "../sharedStyles/centralizedStyles.css";
  color: #fff;
}
```
This will automatically stack classes in the DOM so that `<tag class="newTextField---h^$# sharedTextFieldClass---h^$2 ">` affectively combining styles from both classes when only `className={css.newTextFieldClass}` was supplied as a prop.
It is possible to compose of multiple class names using multiple `composes` properties.
More info here in CSS-loader's documentation: https://github.com/webpack-contrib/css-loader#composing

### DIFF
--- a/webpack.config.cli.dev.js
+++ b/webpack.config.cli.dev.js
@@ -39,6 +39,7 @@ devConfig.module.rules.push({
         localIdentName: '[local]---[hash:base64:5]',
         modules: true,
         sourceMap: true,
+        importLoaders: 1,
       },
     },
     {
--- a/webpack.config.cli.prod.js
+++ b/webpack.config.cli.prod.js
@@ -32,6 +32,7 @@ prodConfig.module.rules.push({
         options: {
           localIdentName: '[local]---[hash:base64:5]',
           modules: true,
+          importLoaders: 1,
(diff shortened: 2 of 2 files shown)


ttamj:
@JohnC-80 I haven't been authoring enough CSS to have a strong preference, however this sounds good to me.  Does this have any side effects on existing CSS?  Or is this configuration simply enabling us to use features like `composes` going forward?

JohnC-80:
@ttamj it enables features going forward. If you attempted to use `composes` to combine the styles of 2 classes in separate files at the moment, it wouldn't work because the class that's imported wouldn't be processed by postcss and the properties would simply be invalid and not show up.

PR #250, opened 2018-03-27
STCOR-179

szkata:
Add Hungarian translations to stripes-core.


translations/en.json
```diff
@@ -1,5 +1,20 @@
...
   "front.welcome": "Welcome, the Future Of Libraries Is OPEN!",
   "front.home": "Home",
-  "front.about": "Software versions"
+  "front.about": "Software versions",
+  "button.new": "+ New",
+  "button.new_tooltip": "Add {entry}",
+  "button.edit": "Edit",
+  "button.delete": "Delete",
+  "button.deleteEntry": "Delete {entry}",
+  "button.save": "Save and close",
```
doytch: Can we have this be called `button.saveAndClose` instead? Quick grep through the project shows that we're labeling some things "Save" so I think it'd be best to have both strings as an options.

zburke: If we're going to change the key, let's move it into `stripes-smart-components` as well. I just merged #251 (see [STCOR-178](https://issues.folio.org/browse/STCOR-178)) which allows us to keep translations in `stripes-components` and `stripes-smart-components`.

zburke: p.s. I’m happy to do the grunt work of moving the keys and updating the PRs against stripes-components and stripes-smart-components since you already did the hard part of pulling out the strings and doing the translation. This works looks great — you were just a step ahead of me in getting translation-keys into the stripes-libraries before we figured out how best to handle that!

szkata: I just updated this as it is not used system-wide yet.

PR #252, opened 2018-03-28
Stcor time zone

aditya-matukumalli:
Refs UIORG-55

### DIFF
--- a/src/components/Root/Root.js
+++ b/src/components/Root/Root.js
@@ -10,7 +10,7 @@ import initialReducers from '../../initialReducers';
 import enhanceReducer from '../../enhanceReducer';
 import createApolloClient from '../../createApolloClient';
 import { setSinglePlugin, setBindings, setOkapiToken, setTimezone } from '../../okapiActions';
-import { getDateTime, formatDate, formatTime, formatDateTime } from '../../../util/dateUtil';
+import { formatDate, formatTime, formatDateTime } from '../../../util/dateUtil';
 import { loadTranslations, checkOkapiSession } from '../../loginServices';
 import { getQueryResourceKey } from '../../locationService';
 import Stripes from '../../Stripes';
@@ -103,7 +103,6 @@ class Root extends Component {
       setTimezone: (timezoneValue) => { store.dispatch(setTimezone(timezoneValue)); },
       plugins: plugins || {},
       setSinglePlugin: (key, value) => { store.dispatch(setSinglePlugin(key, value)); },
-      getDateTime: (dateStr, zone) => getDateTime(dateStr, zone || timezone),
       formatDate: (dateStr, zone) => formatDate(dateStr, zone || timezone),
       formatTime: (dateStr, zone) => formatTime(dateStr, zone || timezone),
       formatDateTime: (dateStr, zone) => formatDateTime(dateStr, zone || timezone),
--- a/src/components/Root/Root.js
+++ b/src/components/Root/Root.js
@@ -6,11 +6,11 @@ import createBrowserHistory from 'history/createBrowserHistory';
 import { IntlProvider } from 'react-intl';
 import queryString from 'query-string';
 import { ApolloProvider } from 'react-apollo';
-
 import initialReducers from '../../initialReducers';
 import enhanceReducer from '../../enhanceReducer';
 import createApolloClient from '../../createApolloClient';
-import { setSinglePlugin, setBindings, setOkapiToken } from '../../okapiActions';
+import { setSinglePlugin, setBindings, setOkapiToken, setTimezone } from '../../okapiActions';
+import { getDateTime, formatDate, formatTime, formatDateTime } from '../../../util/dateUtil';
 import { loadTranslations, checkOkapiSession } from '../../loginServices';
 import { getQueryResourceKey } from '../../locationService';
 import Stripes from '../../Stripes';
@@ -40,7 +40,7 @@ class Root extends Component {
   }
 
   componentWillMount() {
-    const { okapi, store, locale } = this.props;
+    const { okapi, store, locale, timezone } = this.props;
     if (this.withOkapi) checkOkapiSession(okapi.url, store, okapi.tenant);
     // TODO: remove this after we load locale and translations at start from a public endpoint
     loadTranslations(store, locale);
@@ -76,7 +76,7 @@ class Root extends Component {
   }
 
   render() {
-    const { logger, store, epics, config, okapi, actionNames, token, disableAuth, currentUser, currentPerms, locale, plugins, bindings, discovery, translations, history, serverDown } = this.props;
+    const { logger, store, epics, config, okapi, actionNames, token, disableAuth, currentUser, currentPerms, locale, timezone, plugins, bindings, discovery, translations, history, serverDown } = this.props;
 
     if (serverDown) {
       return <div>Error: server is down.</div>;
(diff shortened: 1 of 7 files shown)


src/components/Root/Root.js
```diff
@@ -97,10 +97,16 @@ class Root extends Component {
...
       actionNames,
       locale,
+      timezone,
       metadata,
-      setLocale: (val) => { loadTranslations(store, val); },
+      setLocale: (localeValue) => { loadTranslations(store, localeValue); },
+      setTimezone: (timezoneValue) => { store.dispatch(setTimezone(timezoneValue)); },
       plugins: plugins || {},
       setSinglePlugin: (key, value) => { store.dispatch(setSinglePlugin(key, value)); },
+      getDateTime: (dateStr, zone) => getDateTime(dateStr, zone || timezone),
```
MikeTaylor: It's not clear to me that these non-Stripes-related utility functions belong on the Stripes object. Is there a reason for this?

src/components/Root/Root.js
```diff
@@ -6,11 +6,11 @@ import createBrowserHistory from 'history/createBrowserHistory';
 import { IntlProvider } from 'react-intl';
 import queryString from 'query-string';
 import { ApolloProvider } from 'react-apollo';
-
```
MikeTaylor: Best to avoid personal-preference changes, like removing blank lines, as part of a more substantial change. It makes the PR look bigger and harder to make sense of. Not a big deal, just something to bear in mind.

src/loginServices.js
```diff
@@ -45,13 +46,16 @@ export function loadTranslations(store, locale) {
 }
 
 export function getLocale(okapiUrl, store, tenant) {
-  fetch(`${okapiUrl}/configurations/entries?query=(module=ORG and configName=locale)`,
+  fetch(`${okapiUrl}/configurations/entries?query=(module=ORG and configName=localeSettings)`,
```
MikeTaylor: Why have you changed the key used in the configuration module?

MikeTaylor: Oh, I see -- it's because you've changed the structure of the data that it carries. Well, that makes a certain kind of sense, but why are you shoving locale and timezone together into a single configuration record anyway? Why don't they each get their own configuration record?

src/okapiActions.js
```diff
@@ -97,6 +104,7 @@ export { setCurrentUser,
   clearCurrentUser,
   setCurrentPerms,
   setLocale,
+  setTimezone,
```
MikeTaylor: Don't you just hate that you have to put single lines of code in about eight different places when you add a new datum like this? I don't know what we could do differently to avoid it, but it does sometimes feel like Redux wants you to fill in forms in triplicate.

MikeTaylor:
Oh, one more thing: please include the Jira issue number in commit messages and PR descriptions.

src/components/Root/Root.js
aditya-matukumalli: @MikeTaylor We have `formattedDate` and `formattedTime` utilized all over the apps, in stripes-* and ui-*. So, extending the stripes object provides a centralized way in which these functions could be accessed, instead of importing it from the `react-intl` everytime. With this, we can remove a lot of `util` files that are present in every ui-* app that primarily implement these 2 functions.

src/components/Root/Root.js
aditya-matukumalli: Sure, will take care of this.

src/loginServices.js
aditya-matukumalli: @MikeTaylor So, the timezonePicker is now added to the locale settings page along with the localePicker. And the `configManager` component that is present handles single value on submit. Hence, stored the locale and timezone under one config in a JSON structure. Something similar is being done here: https://github.com/folio-org/ui-circulation/blob/master/settings/CheckoutSettings.js

src/components/Root/Root.js
MikeTaylor: This doesn't sound right to me. The Stripes object has a specific function -- i.e. supplying Stripes-specific functionality to client code. It's not just a handy place to hang any old bit of functionality that we can't be bothered to import.

If
```
return this.props.stripes.getTImeDate(dateStr, timezone);
```
is merely a shorthand for
```
import moment from 'moment-timezone';
return moment.tz(dateStr, timezone);
```
then I think we should say what we mean.

MikeTaylor: And, needless to say, the `util` files, if they are used in many different apps, should not be replicated in those apps but moved to a central place whence they can be shared.

src/loginServices.js
MikeTaylor: It doesn't seem right to me that we're allowing our data-structures to be dictated by UI design. I would have preferred to use a simplest data structure, then write UI code that access it.

That said, I don't think it's a big deal; so if the relevant code is already written, let's run with that rather than doing the extra work to change it now.

src/components/Root/Root.js
aditya-matukumalli: Oh, my bad. I didn't realize you were talking about the `getDateTime()` function. Yeah, that can definitely be moved out of stripes. Will do it.

MikeTaylor: Well, I'm not really clear on what any of them do. Let me ask you this: what code to you have to write at present to use these other functions, and what would you be able to reduce it to if these functions came on the Stripes object?

aditya-matukumalli: So, the `getDateTime()` function is not being used anywhere else in the app. Its just being used in the `dateUtil` file which I have added with this PR, and I agree thats redundant and can be removed. I will remove it now.

As far as the other functions:

Code that is currently present:

in every app you have a `util.js` file that contains 2 functions: `formatDate` and `formatTime`
With this PR it is now just present in one file on stripes-core

Also, in each app:

for eg: https://github.com/folio-org/ui-users/blob/master/lib/Loans/ClosedLoans/ClosedLoans.js#L6. 

You have such lines in every app, that has the `util.js` file with the same function body.

Also, the `timezone` string that is now present through stripes object needn't be passed around all over the apps ( because you would have to then pass the timezone as an argument/ and pass it a prop to the `FormattedDate` and `FormattedTime` in every function that resides in all the `util` files). Now, with this PR, It just resides in stripes-core based on which the date and time can be formatted.

So, with the current PR, it comes down to: (just one line)

`props.stripes.formatDate(dateStr)`

With the already present code: 

=> import { formatDate } from '../util'
=> timezone = props.stripes.timezone;
=> `formatDate(dateStr, timezone)` 

and in the util file (this is in every app)
=> import { formattedDate } from 'react-intl'
=>  formatDate = (dateStr, timezone)) => `<formattedDate value={dateStr} timezone={timezone} />

MikeTaylor: OK, so there are two things going on here.

First, it's clearly idiotic that every app is carrying its own mostly-redundant-but-slightly-divergent `util.js` (or in some cases, I notice, `utils.js`, plural). Obviously such code should by DRYed and put in a shared location.

In general, the Stripes object is _not_ that location: such functions should usually go in `@folio/stripes-components/util' -- there's plenty of precedent. (One day, we may separate that area out into its own `@folio/stripes-util`, but it is not this day.)

_However_, in the present case, I see that `stripes.formatDate` is currying the underlying function with a value that is known to Stripes -- that is, the timezone. I do agree that this changes the equation. What this means is that what we have here really is a _Stripes-aware_ date-formatting function -- that is, aware of the timezone that Stripes has established. So, yes, that one does belong on the Stripes object.

I've not looked at the others, but you can make your own judgement based on these principles.

Sorry this has dragged on a bit. I don't like to be That Guy, but I hope it's been productive for you.

aditya-matukumalli: Sure, its been informative. Will remove the `getDateTime()` function and merge it in a bit. Thanks

MikeTaylor: Thank you!

PR #255, opened 2018-03-30
don't error when searching for translations in stripes- libraries. Fixes STCOR-183

zburke:
Fixes [STCOR-183](https://issues.folio.org/browse/STCOR-183)


zburke:
This change is more of a workaround then an actual fix, but the alternative was rewriting the whole plugin and probably some other pieces too. (The list of `stripes-` dependencies isn't dynamically available at this point in the build process, hence hard-coding them into the constructor, and the refactoring required to make them dynamically available would be significant.)

PR #257, opened 2018-04-02
UIEH-130 Create z-index management solution

cherewaty:
## Purpose
`ui-eholdings` uses `z-index` to manage layering of its search layout. The dropdown menu from the header avatar ends up underneath the third pane: https://issues.folio.org/browse/UIEH-130

Handling `z-index` can be notoriously finicky. The solution to a problem is frequently "set it to a very high number," but with many layers happening, that can get messy real quick.

## Approach
https://github.com/folio-org/stripes-components/pull/314

## TODO
- [ ] Merge `stripes-components` PR first.
- [ ] Remove commit pointing to fork.

### DIFF
--- a/.stylelintrc
+++ b/.stylelintrc
@@ -1,3 +1,6 @@
 {
-  "extends": "stylelint-config-standard"
+  "extends": "stylelint-config-standard",
+  "rules": {
+    "function-name-case": null
+  }
 }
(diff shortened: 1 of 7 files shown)


cherewaty:
https://github.com/folio-org/stripes-components/pull/317 fixed the bug reported in https://issues.folio.org/browse/UIEH-130.

Spinoff ticket: https://issues.folio.org/browse/STCOM-247

PR #258, opened 2018-04-02
Add support for audio files. Fixes STCOR-184

mkuklis:
[https://issues.folio.org/browse/STCOR-184](https://issues.folio.org/browse/STCOR-184)


mkuklis:
@ttamj would you mind taking a quick look at this change? I needed it for https://issues.folio.org/browse/UICHKOUT-68. Thanks!

PR #271, opened 2018-04-12
Extend the stripes object with a function used to calculate timezone …

aditya-matukumalli:
…agnostic dates, refs UIORG-55

- Added in the formatAbsoluteDate function that takes in a date string as an argument and calls the formatDate function woth datestring and UTC as the arguments. This helps us in displaying the
dates that are sent from the server ( in UTC ), and that needs to be the same irrespective of the timezone.

### DIFF
--- a/src/components/Root/Root.js
+++ b/src/components/Root/Root.js
@@ -106,6 +106,7 @@ class Root extends Component {
       formatDate: (dateStr, zone) => formatDate(dateStr, zone || timezone),
       formatTime: (dateStr, zone) => formatTime(dateStr, zone || timezone),
       formatDateTime: (dateStr, zone) => formatDateTime(dateStr, zone || timezone),
+      formatAbsoluteDate: (dateStr) => formatDate(dateStr, 'UTC'),
       bindings,
       setBindings: (val) => { store.dispatch(setBindings(val)); },
       discovery,


aditya-matukumalli:
With the current setup, where the server speaks UTC, for dates like the birthdays, the date returned is of the format `2018-04-11T00:00:00.000+0000` which represents `11th april 2018` `12 AM UTC` time. What this means is, it would display 11th April for any timezones east of UTC, and 10th April for timezones west of it. We do not want this behavior, as the birthdate should remain the same irrespective of the timezone. To facilitate this we added in a new function, that utilizes the already existent `formatDate` function by passing it a timezone of `UTC` explicitly instead of the tenant timezone setting. This facilitates parsing the date sent up from the backend in `UTC` timezone thereby giving us the absolute date that we desire, without any timezone interceptions or transformations.

MikeTaylor:
Thanks, that summarises admirably.

... but what I meant was a comment (maybe even the exact same one) in the code.

MikeTaylor:
I think it's quite right that the Stripes code is, on the whole, low on comments. We definitely don't want [this kind of thing](https://reprog.files.wordpress.com/2010/03/eops-comments.jpeg): 

But this is one of those rare occasions where the code, for good and legitimate reasons, does something that would be completely inexplicable to someone coming to it cold. It's good to help out such people (including our future selves).

PR #274, opened 2018-04-17
STCOR-192: Update Node.js version requirement from 6.x to >=8.11.1

julianladisch:


skomorokh:
Perhaps update the wording in the quickstart to avoid the number and reflect our general goal of compatibility with Current and LTS (once we hit production we may also have longer support windows for whichever node version our own LTS freezes at). 

And in the package.json I think >=8.0.0 is more appropriate as we don't want to trip up people who are relying on system packages or other means of installing eg. your very specific commit might break more than just the CI.

MikeTaylor:
Sorry this got dragged out for so long: thanks for pursuing it through to the bitter end!

PR #276, opened 2018-04-19
STCOR-196: Drop translation strings that are no longer in use.

julianladisch:


zburke:
Agreed, @MikeTaylor; many of the values here are used in other repositories.

julianladisch:
Yes, you are right. stripes-core serves as a string repository for other repositories.
I've checked each entry and found 5 translation keys that I cannot find in any folio-org repository.

MikeTaylor:
@rjruizes Are you confident that merging this is the right thing? Unless you know something new that I do not, the outcome of our discussion above was the this PR was mistake, and that merging it will break many applications.

I have opened a PR to revert this -- #280 -- but I will not merge it until I hear from you, in case this really was deliberate and merging was correct for reasons as-yet obscure to me.

MikeTaylor:
Update @rjruizes assures me (on Slack) that this merge is correct: that the message keys being removed are not in use. I have closed my reversion PR.

julianladisch:
The commit message contains the search link for each message key showing that it is not in use:
https://github.com/folio-org/stripes-core/pull/276/commits/aa1b22dfd8b448497429e56bc3b6386e2c9fbc4f

PR #278, opened 2018-04-19
STCON-67 && STCOM-150 Error Boundary

doytch:
STCON-67 turns off the `alert()` box when [using the `throwErrors` prop implemented here](https://github.com/folio-org/stripes-connect/pull/44). 

STCOM-150 adds [an Error Boundary at the top-level of the application.](https://reactjs.org/docs/error-boundaries.html). Error boundaries let us catch errors thrown _only_ in render() or lifecycle methods. 

React 16 unmounts the component tree when errors are thrown rather than operating in an unstable state, so error boundaries give us the opportunity to introduce a fallback UI. Contrary to what I initially thought, they are not intended to be used as a mechanism to trap errors and return to a stable state. That's why `GlobalErrorBoundary` is fairly dumb and just reports the error and doesn't do anything interesting.

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,7 +48,7 @@
   },
   "dependencies": {
     "@folio/stripes-components": "^2.0.0",
-    "@folio/stripes-connect": "^3.1.0",
+    "@folio/stripes-connect": "^3.1.2",
     "@folio/stripes-logger": "^0.0.2",
     "apollo-cache-inmemory": "^1.1.1",
     "apollo-client": "^2.0.3",
--- a/src/connectErrorEpic.js
+++ b/src/connectErrorEpic.js
@@ -3,6 +3,9 @@ const connectErrorEpic = action$ => action$
   .map((action) => {
     const { meta, payload: e } = action;
     const op = action.type === '@@stripes-connect/FETCH_ERROR' ? 'GET' : e.type;
+
+    if (!meta.throwErrors) return undefined;
+
     // eslint-disable-next-line prefer-template,no-alert
     window.alert(`ERROR: in module ${meta.module}, operation ${op}`
       + ` on resource '${meta.resource}' failed`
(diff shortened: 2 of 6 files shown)


MikeTaylor:
To be fair, it is _in general_ not possible to recover a safe state from a problem occurring in non-pure functions. How could we write a recover function that tries to figure out what's wrong in stripes-connect's part of the Redux store? It's a beautiful dream, but I wouldn't want to be the one implementing it. (Although the simple special case of non-2xx HTTP responses is simple, of course: we just don't touch the store at all.)

package.json
```diff
@@ -48,7 +48,7 @@
   },
```
MikeTaylor: What new stripes-connect facility are we depending on?

src/connectErrorEpic.js
```diff
@@ -3,6 +3,9 @@ const connectErrorEpic = action$ => action$
   .map((action) => {
     const { meta, payload: e } = action;
     const op = action.type === '@@stripes-connect/FETCH_ERROR' ? 'GET' : e.type;
+
```
MikeTaylor: Is this part of the present PR's functionality, or an unrelated change?

MikeTaylor: Ooohhh ... maybe this is the new stripes-connect functionality that we're depending on? Great if so, but the stripes-connect `CHANGELOG.md` says nothing about what's in v3.1.2. Maybe you could fix that?

MikeTaylor: _sigh_ save yourself some time and ignore everything I just wrote. I realised my copy of stripes-connect was outdated. I `git pull`ed and the `CHANGELOG.md` explains perfectly.

(Only one nit on that score: the entry does not reference a Jira  for `throwErrors` -- was there not one?)

doytch:
Yeah totally, we were just originally hoping to be able to throw in render based on `failedMutations` and have error boundaries deal with the error rendering at a top level. It ain't so simple unfortunately. :)

src/connectErrorEpic.js
doytch: Ahhh, it doesn't, good nit. :) It was STCON-67

MikeTaylor: Anyway: go ahead and merge. Thanks for handling this.

MikeTaylor:
Jason makes an excellent point. It's good that this is merged, but we should _also_ look at isolating individual apps if possible.

PR #279, opened 2018-04-20
STCOR-150: Added error boundaries around each module

doytch:
We now wrap each module in the generic [`<ErrorBoundary>` added to stripes-components](https://github.com/folio-org/stripes-components/pull/336). 

This means that if some component in ui-users throws an error during a lifecycle method, then the Users app will display the error, but other modules will remain functional.


src/RootWithIntl.js
```diff
@@ -21,6 +22,10 @@ import OverlayContainer from './components/OverlayContainer';
 import getModuleRoutes from './moduleRoutes';
 import { stripesShape } from './Stripes';
 
+const wrapInErrorBoundary = (component) => (
```
doytch: I'm defining `wrapInErrorBoundary` here purely for readability sake. There's so many brackets floating around that it made the code cleaner IMO.

skomorokh:
Whoa... that's more involved than I expected... but might come in handy as a stripes-component... I guess? What do we gain from wrapping each of the potential contents of `Switch` independently rather than just `ModuleContainer`? 

I was thinking more along the lines of just handling the error prop here: 
https://github.com/folio-org/stripes-core/blob/master/src/components/ModuleContainer/ModuleContainer.js

But ya, perhaps just put a `<ErrorBoundry>` around ModuleContainer?

Also, i18n is a consideration here. We should translate the text in the component, but I wonder if there is currently any provision in the backend to provide localised error messages?

PR #280, opened 2018-04-23
Revert "STCOR-196: Drop translation strings that are no longer in use."

MikeTaylor:
Reverts folio-org/stripes-core#276


MikeTaylor:
Not needed after all: the initial merge was correct.

PR #284, opened 2018-04-27
Create TimeZone and Locale contexts

cherewaty:
## Purpose
Work for https://issues.folio.org/browse/UIORG-55 changed the behavior of `Datepicker`. Instead of emitting an ISO-8601 datetime string in the user's local time zone, it now emits the datetime at UTC.

That needs tests. We should be able to more easily identify breaking changes.

## Approach
Adds context providers for `locale` and `timeZone`.

Much more detail at: https://github.com/folio-org/stripes-components/pull/349

## Risk
Accessing `this.context.stripes.locale` and `this.context.stripes.timezone` should continue to work as previously, but we should find a way to deprecate those in favor of using `withLocale` and `withTimeZone`.

## TODO
- [ ] Merge https://github.com/folio-org/stripes-components/pull/349 first.
- [ ] Remove fork from `package.json`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## 2.10.0 (IN PROGRESS)
 
+* Create `Locale` and `TimeZone` contexts. Fixes STCOM-259.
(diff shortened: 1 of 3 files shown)


cherewaty:
Closing, see https://issues.folio.org/browse/STCOR-208

PR #293, opened 2018-05-11
STCOR-209: Added initial TypeScript support

doytch:
I forgot I had this branch kicking around for a while and remembered it during a discussion at Wolfcon. 

Pretty small and minimal here, all we're doing is adding support for typescript file endings and bringing in awesome-typescript-loader for the loading. If I was starting from scratch I might've created this as a `RFC` JIRA but given that the code's already written I'd throw it out here. Feel free to comment on alternative loaders or approaches, I welcome it!


webpack/typescript-loader-rule.js
```diff
@@ -0,0 +1,24 @@
...
+  }
+
+  return false;
+}
+
+module.exports = {
+  test: babelLoaderTest,
+  loader: 'awesome-typescript-loader',
+  query: {
+    configFileName: locateStripesModule('.', '@folio/stripes-core', '', 'webpack', 'tsconfig.json'),
```
ttamj: @doytch Given we are already in the stripes-core source, we shouldn't have to locate stripes-core like we do other ui modules (which may exist locally outside a workspace).  I would expect `'./tsconfig.json'` or possibly `path.join(__dirname, 'tsconfig.json')` to work in this case.

PR #295, opened 2018-05-14
fix doc typo, and deprecated lint

rjruizes:


MikeTaylor:
You did of course check that the changed lifecycle components still do what we want them to?

Thanks for handling this.

rjruizes:
I couldn't test SSOLanding well since I don't have an SSO setup. The Root component was fine. 

Conceptually, moving the network requests from before to after the mount shouldn't make a difference.

MikeTaylor:
Better ping whoever `git blame` says wrote that code in the first place, and ask them to verify it.

PR #297, opened 2018-05-15
Upgrade required version of hard-source-webpack-plugin

MikeTaylor:
... and thereby of leveldown. Fixes STCOR-210.


ttamj:
Everything seems to build with the `--cache` option, but I do get the following deprecation warning:

```
[hard-source:core] The `recordsPath` option to HardSourceWebpackPlugin is deprecated in 0.6 and will be removed in 0.7. 0.6 and later do not need recordsPath. If you still need it outside HardSourceWebpackPlugin you can set recordsPath on the root of your webpack configuration.
```

MikeTaylor:
Good to know it builds OK.

And, yes, _of course_ there is a deprecation warning: what NPM package upgrade would be complete without one?

_Do_ we need `recordsPath`? This is an area of the code that I have never so much as looked at.

PR #299, opened 2018-05-16
translate module types on About page

rjruizes:
Addressing Julian's [comment](https://issues.folio.org/browse/STCOR-196?focusedCommentId=30143&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-30143) to translate the module types (_app_, _settings_, and _plugin_)

### DIFF
--- a/src/components/About/About.js
+++ b/src/components/About/About.js
@@ -63,7 +63,20 @@ const About = (props) => {
 
   function listModules(caption, list, interfaces) {
     const itemFormatter = m => (<li key={m.module}>{renderDependencies(m, interfaces)}</li>);
-    const headlineMsg = <FormattedMessage id="stripes-core.about.moduleTypeCount" values={{ count: list.length, type: caption }} />;
+    var headlineMsg;
+    switch (caption) {
+      case 'app':
+        headlineMsg = <FormattedMessage id="stripes-core.about.appModuleCount" values={{ count: list.length }} />;
+        break;
+      case 'settings':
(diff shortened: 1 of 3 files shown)


MikeTaylor:
I'm not sure this is right. `app`, `settings` and `plugin` are really opaque tokens that have meaning to programmers, just as much as they are human-readable.

julianladisch:
There may be languages where "module" has a translation that differs depending on "app", "settings", "plugin".

rjruizes:
I used Google Translate to render in German what Mike suggests (translating only "module") and what Julian suggests (translating the whole phrase):

| Mike              | Julian               |
|-------------------|----------------------|
| 3 app module      | 3 App-Module         |
| 3 settings module | 3 Einstellungsmodule |

```
"{count, number} settings {count, plural, one {module} other {modules}}"
```

I'm neither sure if the the Google translation is correct, or if Lokalize would correctly translate `"settings modules" -> "Einstellungsmodule"` the way I've done it.

MikeTaylor:
Well, what I am suggesting is that perhaps the only thing localised should be the string "There are {number} '{type}' modules", with both the number and the type being language-independent. Because even in German or Chinese, a module of type "app" is still a module of type "app".

I don't feel strongly about this, though. I guess in part it comes down to the question of who this About page is for. I originally made it for developers; but if humans are also using it now, that's cool.

julianladisch:
Even in German or Chinese, a module of type "settings" is still a module of type "settings". This is true. We do not translate the type, we only translate the displayed name of the type. The Chinese reader that does not understand English want to read the Chinese names for "app" and "settings", not the English names.

PR #300, opened 2018-05-16
include ui-vendors in pull-stripes

zburke:
Is `ui-vendors` enough of a core repository that we should include it in `pull-stripes`? Seems like "yes" to me.


skomorokh:
Why another day? What does stripes-cli's `stripes platform pull` not do that warrants maintaining this script?

MikeTaylor:
Another day because I don't want to prevent merging this PR.

PR #301, opened 2018-05-20
Adding system skeleton loading

rasmuswoelk:
We will be adding skeleton loading to various parts of the system to improve perceived performance when fetching data. 

This PR adds a general system skeleton that will be rendered until the app is ready to be shown.

This replaces the "Loading translations..."-text.

## Before
may-20-2018 18-30-30

## After
may-20-2018 18-29-02

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,7 @@
 * Retrieve tenant module details in one swell foop. Fixes STCOR-200. Available from v2.9.6.
 * Add build option to disable JS minification, STCOR-197
 * Upgrade required version of hard-source-webpack-plugin, and thereby of leveldown. Fixes STCOR-210.
+* Added system skeleton loading
 
 ## [2.9.0](https://github.com/folio-org/stripes-core/tree/v2.9.0) (2018-02-01)
 [Full Changelog](https://github.com/folio-org/stripes-core/compare/v2.8.0...v2.9.0)
--- a/src/components/Root/Root.js
+++ b/src/components/Root/Root.js
@@ -16,6 +16,7 @@ import { loadTranslations, checkOkapiSession } from '../../loginServices';
 import { getQueryResourceKey } from '../../locationService';
 import Stripes from '../../Stripes';
 import RootWithIntl from '../../RootWithIntl';
+import SystemSkeleton from '../SystemSkeleton';
 
 import './Root.css';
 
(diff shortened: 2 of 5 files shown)


MikeTaylor:
This looks much worse to me than how things are now. It's a complete fiction, and it bears no relation to what finally renders.

rasmuswoelk:
I'm sorry you feel that way, @MikeTaylor. I know that this system-wide skeleton is a bit generic since we obviously can't really know what will render. I could have picked a better example for the gifs (e.g. the Users-app) but that doesn't change the fact it's impossible to know for sure what will render.

However, I don't think that it's important that the skeleton is a 1:1 version of the content that will finally render. It feels faster by giving an indication of what the content _could_ or is going to look like before it has loaded. 

Once the app is loaded we can add skeleton loading for smaller parts of the UI - such as the results list, preview pane etc. which will mimic the actual content in a more precise way but still it will be a skeleton and not the actual content.

may-21-2018 06-43-16

(Note: The font jumps once it's loaded. This shouldn't happen in production. Not sure what it happens on my local machine.)

cherewaty:
@rasmuswoelk what if we split this into two parts?
- The `SystemSkeleton` could be just for the header navigation served up by `stripes-core`.
- A skeleton loading state for `<SearchAndSort>` in `stripes-smart-components`.

I think we ultimately have to leave it up to each UI module what its loading state looks like. As currently built, the transition into an app not using `<SearchAndSort>` (like Check-in, Check-out, Settings, etc.) would be very jarring.

A skeleton loading state for the `stripes-core` header navigation would still be a huge improvement over the "Loading translations..." text, and I think restricting this PR to just the `stripes-core` header would make it shippable. The module container could just have a white or light gray background like https://developer.apple.com/ios/human-interface-guidelines/icons-and-images/launch-screen/ until the module kicks in with its own loading state.

MikeTaylor:
I should really butt out of this, since we're really talking about a UX decision rather then something technical that I have some actual expertise in. But, hey, why change the habit of a lifetime?

I'm very much on board with Jeremy's suggestion. The problem with the mockup you show here, @rasmuswoelk, is that it actively lies about what's going to be displayed. That doesn't (to me) make anything feel like it's loading faster -- it makes it seem that the system is confused, or even broken.

rasmuswoelk:
You have a point. I'll remove the content specific skeleton and keep the header.

MikeTaylor:
Sweet -- thanks for hearing us.

PR #303, opened 2018-05-21
STCOR-207 Set document lang attribute for assistive technology

JohnC-80:
# Purpose
So that assistive technology can pick up on which language the document should be read in, the language code should be indicated in the html as a `lang` attribute on the `<html>` tag.

https://www.w3.org/International/questions/qa-html-language-declarations

# Approach
This is set to "en" as default in the base index.html of stripes-core, which is updated later once the locale information loads from okapi. 

Corresponds with another PR on `ui-org` for when the locale is changed. https://github.com/folio-org/ui-organization/pull/72

### DIFF
--- a/index.html
+++ b/index.html
@@ -1,5 +1,5 @@
 <!doctype html>
-<html>
+<html lang="en">
   <head>
     <meta charset="utf-8">
     <title>FOLIO</title>
--- a/src/loginServices.js
+++ b/src/loginServices.js
@@ -54,7 +54,11 @@ export function getLocale(okapiUrl, store, tenant) {
           if (json.configs.length) {
             const localeValues = JSON.parse(json.configs[0].value);
             const { locale, timezone } = localeValues;
-            if (locale) loadTranslations(store, locale);
+            if (locale) {
+              loadTranslations(store, locale);
+              // set title of document using first 2 characters of locale.
+              document.documentElement.setAttribute('lang', locale.substr(0, 2));
+            }
             if (timezone) store.dispatch(setTimezone(timezone));
           }
         });


Robdel12:
This _may not_ work if we're changing the language attribute dynamically (after page render). 

This is a test done in Chrome (which doesn't pick up the change): https://streamable.com/ixckg
This is a test done in Safari (which _does_ pick up the change): https://streamable.com/5ez61

I don't have access to a windows machine right now but I'd love to know what NVDA (Chrome / FireFox) & JAWS (IE11) do with this.

zburke:
Rats. @Robdel12, if this technique doesn't work reliably, is there one that does? Do we have force a reload, not just a re-render?

rasmuswoelk:
I guess users won't change the language that often.. so a page refresh could do the trick?

JohnC-80:
@ttamj any way to get in front of the initial page render and set the `lang` attribute in the index.html?

JohnC-80:
Windows, setting the document lang after the initial load (used a timeout to set initial, with buttons changing the lang via js afterward)
JAWS 2018 in IE11 (and Chrome) works.
NVDA does not in Chrome or Firefox.
Here's an ugly capture I did of IE11... https://streamable.com/ji48v

Robdel12:
@JohnC-80 great work! So to summarize:
NVDA does _not work_ (Tested: Firefox / Chrome)
JAWS works (Tested: IE11)
VoiceOver works with Safari (not chrome)

If it's possible to get the language set before the initial render that would be the most ideal solution. One thing that sprung to mind would be server side rendering, which is not possible right now.

But if we can't do that then it _seems_ NVDA users would be the only ones that are impacted by this. Not ideal but at least it's not most of the screen readers. (I'd be curious to see how Talkback handles this too)

ttamj:
@JohnC-80 We can set the language before initial render.  However, the current way we handle index.html (statically rather than dynamically) limits us to supplying only what we know about the tenant at build-time.  Should the tenant change their language preference later, a new client build would be necessary.  This is also true today for changing something like the branding logo, so perhaps it is not that much of an issue for now.

JohnC-80:
@ttamj Sorry for the eyes going elsewhere for a moment. When a tenant changes the localization dropdown in `UI-ORG` the `lang` attribute on the `<html>` tag would change once that form is saved... so session there forward on the smarter screen-readers would be covered. I'm not sure what the UX would be when a re-build of the tenant's bundle would be from the UI (Probably silent? But don't know that we've encountered it before.) or for asking users to restart if the language isn't picked up. (Probably text in the UI in UI-ORG)
The static `en` setting that I've put into place in `index.html` covers us temporarily for not having a setting there initially - but it's certainly not what we want to keep for the long run. Getting that tenant setting in there at build time is key.
I feel like this can be merged in now and the tenant bundle-build setup can happen separately (unless it's somehow incredibly convenient and simple for @ttamj to operate on this branch.)

ttamj:
I agree. This approach is reasonable for now.

My build-time suggestion is only a means to inject a more accurate tenant default.  I have created STCOR-214 to capture this.  It would be similar to how the logos are handled today.  We don't yet have the facility to auto-generate builds from the UI and that would be an entirely separate effort.

PR #304, opened 2018-05-21
translate login page

zburke:
Thread the `stripes` object through to the login page so we can
translate its embedded strings.

Fixes [STCOR-213](https://issues.folio.org/browse/STCOR-213).


zburke:
Tick: 
draw_the_tick

PR #310, opened 2018-06-02
STCOR-195: More German translations.

julianladisch:


julianladisch:
Localise.co does not accept any pull request any longer.

PR #313, opened 2018-06-05
Fix prefix when retrieving query resource key. Fixes STCOR-220

mkuklis:
Fixes https://issues.folio.org/browse/STCOR-220

When `dataKey` is used for given `module` all resources in redux store are prefixed with `dataKey#`. For example:

`patrons#folio_users_records`

The previous implementation was missing `#` so it would try to retrieve a wrong key:

`patronsfolio_users_records`


mkuklis:
@MikeTaylor thank you for approving this! I was just writing a bit more to explain what is going on and then I was planning to add you for review.

PR #314, opened 2018-06-05
STCOR-221 Add babel plugins

cherewaty:
## Purpose
`ui-eholdings` has been adding two babel plugins for a while in a [`.stripesclirc.js`](https://github.com/folio-org/ui-eholdings/blob/master/.stripesclirc.js):
### [babel-plugin-transform-decorators-legacy](https://github.com/loganfsmyth/babel-plugin-transform-decorators-legacy)
`ui-eholdings` makes use of decorators in the "interactors" for UI tests. Example: https://github.com/folio-org/ui-eholdings/blob/master/tests/pages/datepicker.js. `ui-eholdings` only adds this plugin in the `test` environment, but I don't see any harm in adding it for all builds.

### [babel-plugin-remove-jsx-attributes](https://github.com/wireapp/babel-plugin-remove-jsx-attributes)
Used to remove `data-test` and `data-test-*` attributes from production builds. `ui-eholdings` relies on `data-test-*` attributes for UI testing instead of IDs (with some exceptions). There were discussions recently about taking this approach in more of the `stripes` ecosystem. This won't break any existing tests using IDs, but will make `data-test` attributes a more attractive option for everyone.

## Approach
I have doubts about the elegance of the way I'm injecting `babel-plugin-remove-jsx-attributes` into production builds (and only production builds). Suggestions welcome.

### DIFF
--- a/webpack.config.cli.prod.js
+++ b/webpack.config.cli.prod.js
@@ -55,4 +55,13 @@ prodConfig.module.rules.push({
   }),
 });
 
+// Remove all data-test or data-test-* attributes
+const babelLoaderConfig = prodConfig.module.rules.find(rule => rule.loader === 'babel-loader');
+
+babelLoaderConfig.options.plugins = (babelLoaderConfig.options.plugins || []).concat([
+  [require.resolve('babel-plugin-remove-jsx-attributes'), {
+    patterns: ['^data-test.*$']
+  }]
+]);
+
 module.exports = prodConfig;
--- a/package.json
+++ b/package.json
@@ -60,6 +60,7 @@
     "babel-eslint": "^8.2.2",
     "babel-loader": "^7.0.0",
     "babel-plugin-react-transform": "^2.0.2",
+    "babel-plugin-transform-decorators-legacy": "^1.3.5",
     "babel-polyfill": "^6.16.0",
     "babel-preset-env": "^1.6.0",
     "babel-preset-react": "^6.16.0",
--- a/package.json
+++ b/package.json
@@ -60,6 +60,7 @@
     "babel-eslint": "^8.2.2",
     "babel-loader": "^7.0.0",
     "babel-plugin-react-transform": "^2.0.2",
+    "babel-plugin-remove-jsx-attributes": "^0.0.2",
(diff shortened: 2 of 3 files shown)


webpack.config.cli.prod.js
```diff
@@ -55,4 +55,14 @@ prodConfig.module.rules.push({
   }),
 });
 
+// Remove all data-test or data-test-* attributes
+const babelLoaderConfigIndex = prodConfig.module.rules.findIndex((rule) => {
+  return rule.loader === 'babel-loader';
+});
+prodConfig.module.rules[babelLoaderConfigIndex].options.plugins = [
```
wwilsman: This seems like it would override previously defined babel plugins. 

Maybe something like this?

```js
const babelLoaderConfig = prodConfig.module.rules.find(rule => rule.loader === 'babel-loader');

babelLoaderConfig.options.plugins = (babelLoaderConfig.options.plugins || []).concat([
  [require.resolve('babel-plugin-remove-jsx-attributes'), {
    patterns: ['^data-test.*$']
  }]
]);
```

cherewaty: Oops, you're right, that would overwrite. Changed 👌

ttamj:
I wonder if moving all of this into `babel-loader-rule` would help, such that we return a fully configured rule (prod or dev) up front.  Then apply the appropriately generated `babel-loader-rule` in the prod/dev webpack config rather then the base. This would avoid the need to reach back and tweak it.

ttamj:
Circling back, I think your approach injecting `babel-plugin-remove-jsx-attributes` is okay the way it is.  After reviewing the dev and prod configs while working on the Webpack 4 branch, it probably makes sense to defer something like my earlier suggestion until after the upgrade anyway.

cherewaty:
Great, I'll gladly be a webpack 4 tester to validate what you've got going there.

Re: decorators: I didn't turn on the `legacy: true` option for `babel-plugin-transform-decorators-legacy`, so we're at least using a newer draft. When we start looking at Babel 7, we'll replace `babel-plugin-transform-decorators-legacy` with [`babel-plugin-proposal-decorators`](https://new.babeljs.io/docs/en/next/babel-plugin-proposal-decorators.html).

PR #324, opened 2018-06-18
STCOR-226: Dynamically set document.title

doytch:
This adds the infrastructure to dynamically set the `document.title` based on the page and (optionally) record being displayed.

- `<TitleManager>` takes `page` and `record` props and builds out a title with the help of the `react-titled` package.
  - Because our desired titling strategy isn't hierarchical (most specific to least specific or vice versa, such as "Record - Page - FOLIO"), but custom ("Page - Record - FOLIO"), we need to be kind of cute with our title-building. I'd have written it to be cleaner if we could just concat titles but 🤷‍♂️ 
- `<AppRoute>` cleans things up a bit for non-module pages like Home, Settings and Login.
- Module and non-module routes currently all dynamically set their titles based on `module.displayName` and a `displayName` prop.
  - I'm gonna start working on a more i18n-friendly approach after this gets merged but wanted to get this through before tackling that.

In `ui-xxx` code, to set the `record` component, you'd just drop in a component like so: 
`<TitleManager record={getFullName(user)} />`. I've already got this going in `ui-users`, so after this gets merged I'll finish that up along with `ui-inventory` and put up PRs for those.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## 2.11.0 (IN PROGRESS)
 * Update release procedure to guard against publishing with unreleased dependencies, STCOR-225
+* Added `<TitleManager>` to help compose `document.title`. Fixes STCOR-226
+* Added `<AppRoute>` component to clean up generic, non-module routing.
 
 
 ## [2.10.0](https://github.com/folio-org/stripes-core/tree/v2.10.0) (2018-06-06)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 * Update release procedure to guard against publishing with unreleased dependencies, STCOR-225
 * Added `<TitleManager>` to help compose `document.title`. Fixes STCOR-226
-* Added `<AppRoute>` component to clean up generic, non-module routing.
+* Added `<TitledRoute>` component to clean up generic, non-module routing.
 * Update release procedure to guard against publishing with unreleased dependencies, STCOR-225
 * Fix typo in [The Stripes Module Developer's Guide](doc/dev-guide.md). Fixes STCOR-227.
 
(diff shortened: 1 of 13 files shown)


skomorokh:
I'm on board with this overall but think we should leave out the `AppRoute` component for now. Reason being that it seems like a good place to delineate major sections of apps--the kinds of things that show up in breadcrumbs and the type-to-search app switcher. And I think it's mostly those sections that we'd want to surface in the title, not necessarily every route within an app? Although, there may be something to be said for using a wrapper for all routes in terms of options that it leaves open later. At any rate, let's discuss that piece more and see how it meshes with Filip's plans but merge the rest so we don't hold up getting at least the app name in the title.

Not super thrilled with things importing from `stripes-core`, but I don't even know why at this point. Previously I'd hoped to have a stripes-test harness that ran an app at a time and mocked via redux but that's water quite far under the bridge. And also was unsettled by circular dependencies but I've pretty much just accepted that we have those and they haven't caused actual problems? Is anyone else ill at ease with importing from core or should I just disregard that as a non-issue from now on?

doytch:
@skomorokh 
**AppRoute**
I'm not sure I totally understand? I put it in because when I started writing stuff like `<Route path="/" key="root" exact component={<TitleManager page="Home"><ErrorBoundary><Front stripes={stripes} /></ErrorBoundary></TitleManager>} />` over and over again I threw up in my mouth a little at the amount of ugly and repeated code. 😝 

**Importing from -core**
It gave me pause at the start as well, but I think the current alternative in our environment is that we instead import from components, _in_ core. That's what causes circular dependencies. I don't think importing from core in components actually causes any circular deps.

doytch:
> the kinds of things that show up in breadcrumbs and the type-to-search app switcher. And I think it's mostly those sections that we'd want to surface in the title, not necessarily every route within an app?

I'd be more likely to agree if things like "Log in" and other non-apps [weren't expressly called out for inclusion on the JIRA](https://issues.folio.org/browse/STCOR-226).

skomorokh:
Okay, I'm sold, definitely see how we'll need something like this even if we have a more elaborate notion of `AppSection` or w/e further down the line. Super minor but I actually got confused by the name, especially since it's intended for non-app things. Maybe more literal, like `TitledRoute`?

doytch:
@skomorokh Agree on the weird naming. Changed it.

PR #327, opened 2018-06-20
STCOR-226: Add ModuleContext which stores i18n'd module metadata

doytch:
**Updated**

- ModulesContext.js creates a React Context of modules. The modules are originally pulled from the `stripes-config` virtual module and then translated by `RootWithIntl`. In the future, the actual translation work can get pulled out into another component but I'm tempted to leave it as-is atm because the translation of icons and permission sets may change where/how we want to do it.
- `stripes-core` now provides `withModule` and `withModules` HOCs. The latter provides all the translated modules, while the former takes either a module name or a `mapStateToProps`-y callback (with props as the arg) that returns the module name, and returns just that module.
   - `SearchAndSort` is an example of an app that makes us of that mSTP format.
- Various pages updated to remove their translations now that it's being done centrally.
- `TitledRoute` updated to translate generic Folio routes.

### DIFF
--- a/doc/app-metadata.md
+++ b/doc/app-metadata.md
@@ -50,7 +50,7 @@ The `stripes` structure contains the following extension fields:
 Field | Location | Description and notes
 :---- | :------- | :--------------------
 Type | `type` | The type of the module: may be `app` for a regular full-page application, `popover` for an application that can run in popover mode as well as in regular full-page mode, or `settings` for modules that only provide an entry in the Settings page.
-Short title | `displayName` | This is needed in the header, where, based on user preference or screen size, the names of the apps may be displayed next to their icons.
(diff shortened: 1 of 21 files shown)


doytch:
I've gone ahead and changed the interface here. New:

- ModulesContext.js which creates a React Context of modules. The modules are originally pulled from the `stripes-config` virtual module and then translated by `RootWithIntl`. In the future, the actual translation work can get pulled out into another component but I'm tempted to leave it as-is atm because the translation of icons and permission sets may change where/how we want to do it.
- `stripes-core` now provides `withModule` and `withModules` HOCs. The latter provides all the translated modules, while the former takes either a module name or a `mapStateToProps`-y callback (with props as the arg) that returns the module name, and returns just that module.
   - `SearchAndSort` is an example of an app that makes us of that mSTP format.

PR #330, opened 2018-06-20
STCOR-208: new context api

rjruizes:
https://issues.folio.org/browse/STCOR-208

Add StripesProvider

### DIFF
--- a/src/StripesContext.js
+++ b/src/StripesContext.js
@@ -3,7 +3,7 @@ import React from 'react';
 export const StripesContext = React.createContext();
 
 export function withStripes(WrappedComponent) {
-  return class extends React.Component {
+  class WithStripes extends React.Component {
     render(props) {
       return (
         <StripesContext.Consumer>
@@ -12,4 +12,10 @@ export function withStripes(WrappedComponent) {
       );
     }
   };
+  WithStripes.displayName = `WithStripes(${getDisplayName(WrappedComponent)})`;
+  return WithStripes;
+}
+
+function getDisplayName(WrappedComponent) {
+  return WrappedComponent.displayName || WrappedComponent.name || 'Component';
 }
--- a/src/StripesContext.js
+++ b/src/StripesContext.js
@@ -1,4 +1,4 @@
-import React from "react";
+import React from 'react';
 
 const Context = React.createContext();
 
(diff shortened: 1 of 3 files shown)


cherewaty:
`withStripes` could probably use a `displayName` to have nicer output in the React console: https://reactjs.org/docs/higher-order-components.html#convention-wrap-the-display-name-for-easy-debugging

src/StripesContext.js
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+
+const Context = React.createContext();
+
+export const StripesProvider = Context.Provider;
+export const StripesConsumer = Context.Consumer;
```
cowboyd: All things being equal, I think we can just do this in a single line:

```js
export default const Context = React.createContext();
```

Then in the wrapper:

```js
<Context.Consumer>
  { stripes => .....}
</Context.Consumer>
```

And in the provider:

```js
import StripesContext from 'stripes-context';

render() {
  return <StripesContext.Provider> ..... </StripesContext.Provider>
}
```

This is also the recommend in the React docs.

src/StripesContext.js
```diff
@@ -0,0 +1,18 @@
...
+
+export function withStripes(WrappedComponent) {
+  return class extends React.Component {
+    render(props) {
+      return (
+        <StripesConsumer>
+          {stripes => <WrappedComponent {...props} stripes={stripes} /> }
```
cowboyd: Should we account for the case where there is no stripes context, or if required, then perhaps even put a PropType on our wrapper component to validate for the stripes object?

rjruizes: `stripes` is not a prop in this `WithStripes` component; `stripes` is a prop to the WrappedComponent, and will have a PropType there.

cowboyd: Ah, yes. I think what I was really getting at was: what do we do in the event that there isn't a stripes context.... will it be an exception, or merely undefined. In the event that it is undefined, should we just render the component unwrapped.

rjruizes: If there is no provider above the consumer, no exception occurs. There will be PropTypes warning from the WrappedComponent. I don't see an advantage for changing rendering when stripes is undefined.

cowboyd: The advantage is that you just pass the component without wrapping it. It saves slightly in render time, but also the tree looks like:

MikeTaylor: > What do we do in the event that there isn't a stripes context.... will it be an exception, or merely undefined. In the event that it is undefined, should we just render the component unwrapped.

That's got to be an exception. The whole purpose of `withStripes` is to get a Stripes object. If there isn't one to be had, there's no value in ploughing on regardless, and delaying the inevitable error until something less diagnosable happens.

PR #331, opened 2018-06-21
Use eslint-config-stripes 2.0

cherewaty:
Code changes reflect new styles from the `react/sort-comp` rule. 

With https://github.com/folio-org/eslint-config-stripes/pull/30, future rule changes should be less surprising.


cherewaty:
@rjruizes I saw `eslint-plugin-react` rules changing out of nowhere on you in https://github.com/folio-org/stripes-core/pull/330; this should prevent those surprise rule changes from happening anymore.

PR #332, opened 2018-06-21
Remove enzyme and react-addons-test-utils

cherewaty:
## Purpose
`yarn install` gave warnings about unmet peer dependencies for `enzyme` and `react-addons-test-utils`. The `enzyme` version specified didn't work with React 16, and `react-addons-test-utils` is deprecated as of React 16.

The only test using `enzyme` in this repo was `routes.js.disabled`, which appears to have been disabled since March 2017: https://github.com/folio-org/stripes-core/commit/f803bc3500864d01c5ec603e3fd7edf5b8bfd8b9#diff-9c45fa6561a34e0c866002ae185407f0

## Approach
Remove `enzyme`, `react-addons-test-utils`, and `routes.js.disabled` test.

`stripes-connect` is the only other `folio-org` repo using `enzyme`, and it defines its own dependency on it.


cherewaty:
bump @skomorokh

cherewaty:
Nothing else in `folio-org` was using `mock-require`, so I removed that too.

PR #337, opened 2018-06-27
throttling the updateQueryResource to see if the state is being commi…

aditya-matukumalli:
…tted, refs UICHKOUT-425


mkuklis:
@aditya-matukumalli looks good! Let's try it out.

PR #345, opened 2018-06-28
Add removeEmpty. Part of STCOR-229

mkuklis:


mkuklis:
@zburke yes all these other commits were trying to solve the same issue until we landed with `removeEmpty` in this last commit. 

We discovered that babel was adding `Symbol` props to prototype of some objects.
 
In our case this was happening with `stateQuery`. The object would look like this:

statequery

This is currently happening in all `folio-*` environments but not in our local build. The reason why this is happening is still not clear but I suspect we use different babel setup when building things locally compared to what is being used when we build `folio-*` envs.

Unfortunately we learned that `omitBy` from `lodash` is also taking prototype chain into consideration when omitting/filtering values. After `omitBy` was executed all prop symbols from the example above were carried up to the top object (pretty funny actually) and `cleanStateQuery` contained all these symbols.

We were able to resolve it with coming up with our own `removeEmpty` helper which only looks up the top props with `forOwn`.

aditya-matukumalli:
@zburke Adding to @mkuklis comment above, when **if (isEqual(cleanStateQuery, cleanLocationQuery)) condition would just fail**, the then this would eventually hit the **history.push(url)** logic in the code-block which would result in pushing the same url the second time onto the history stack and hence requiring two back clicks to get back to the previous page.

ttamj:
> The reason why this is happening is still not clear but I suspect we use different babel setup

@mkuklis Regarding potential Babel differences, perhaps we should clean up [leftover Babel dependencies](https://github.com/folio-org/ui-users/blob/603b7abc813ae2d74c2403dc3c1f4bc3eb809a05/package.json#L208-L213) in the UI modules.  Everything should be using Babel from stripes-core now.  In fact, some of these dependencies, like babel-preset-es2015, are no longer in use.  Whether or not this contributes to the differences you observed, it does muddy the waters.

mkuklis:
@ttamj that sounds like a great idea. I've been ignorant about our babel / webpack / build process. Could we just remove these from `package.json`?

mkuklis:
Just created this one: https://issues.folio.org/browse/STCOR-230 to cover it.

PR #346, opened 2018-06-28
Cleanup console logs. Part of STCOR-229

mkuklis:


mkuklis:
@aditya-matukumalli thank you for working on this with me!

PR #347, opened 2018-06-29
STCOR-175 webpack 4

ttamj:
## Purpose
This PR upgrades stripes-core to use Webpack 4.  Bringing us up to date with Webpack is a helpful step toward reintroducing code-splitting.

## Approach
Most of the effort was a refactor of our own Stripes plugins to support changes in Webpack 4's plugin system.  Several dependencies were updated or replaced for compatibility with Webpack 4.  Items of note include:

* Refactored Stripes config/translations/branding plugins (STCOR-217) to support the plugin system
* Added `favicons-webpack-plugin` to replace logic that manually injected the favicon in a non-plugin friendly way (gaining device-specific favicons in the process)
* Replace `extract-text-webpack-plugin` with `mini-css-extract-plugin` since the `extract-text-webpack-plugin` is no longer recommended for CSS in Webpack 4

## Testers wanted
I'd like to gather feedback from a few folks willing to try out Webpack 4 prior to merging.  Fear not! Trialing these changes can be accomplished while leaving your existing development platform/workspace entirely untouched.

### Here is how:

1) Clone and install this STCOR-175-weback-4 branch to a directory outside your workspace. Do this instead switching branches on an existing clone of stripes-core. Dependency changes would otherwise require a more disruptive clean yarn install to switch back and forth.

```
$ git clone -b STCOR-175-webpack-4 https://github.com/folio-org/stripes-core.git ./temp-stripes-core
$ cd temp-stripes-core
$ yarn install
```

2) Create an alias to use the new stripes-core
```
$ stripes alias add @folio/stripes-core ../path/to/temp-stripes-core
```

3) Now cd to your platform and build/serve with the CLI as usual.
```
$ cd ../path/to/your/platform
$ stripes serve stripes.config.js
```

4) To revert back, simply remove the alias.
```
$ stripes alias remove @folio/stripes-core
```

### Notes:
* When including an alias like this, additional duplication warnings are expected.  This is because the module was brought in outside of your standard node_modules hierarchy. 

* Running Karma tests via Stripes-CLI will also require the `STCLI-61-webpack-4` branch of Stripes CLI.  This is not necessary to build and serve.

* Webpack 4 includes some additional checking by default.  You will see some new warnings for things like the following. These can be addressed independently in their respective modules. For example:
```
WARNING in ./node_modules/@folio/eholdings/node_modules/@folio/stripes-components/lib/Datepicker/index.js 1:0-38
"export 'Calendar' was not found in './Calendar'
```


ttamj:
@cherewaty  FYI, Building with eholding's [demo configuration](https://github.com/folio-org/ui-eholdings/blob/cc2a783ad1309e293a1e63813b47c15e1b3e2064/demo/stripes.config.js#L26) will require a new favicon.  While the plugin will output an `.ico`, it doesn't seem to like them for input.  Switching to a `.png` works.

cherewaty:
I'm not afraid of a clean `yarn install` 😂 

Here's what I did with `ui-eholdings`:
1. Created a fresh branch
2. `yarn cache clean`
3. Deleted `yarn.lock` and `node_modules`
4. Switched `"@folio/stripes-core": "^2.10.0"` in `devDependencies` to `"@folio/stripes-core": "git+https://git@github.com/folio-org/stripes-core.git#STCOR-175-webpack-4"`
5. `yarn`
6. Deleted the favicon from `demo/stripes.config.js`, so I wasn't dealing with that
7. `yarn start`

That worked. The server booted, app seemed to be working. 👌 

When running `yarn start --mirage` instead:
screen shot 2018-07-02 at 4 12 33 pm
Maybe we need some changes to how we hook into `webpackOverrides`?

ttamj:
>I'm not afraid of a clean yarn install 😂

Excellent!  Thank you for testing this out.

I found the problem. Eholdings pulls in its own [copy of webpack](https://github.com/folio-org/ui-eholdings/blob/6bc7a7e0502d78572c86369d7f8f787279a3e282/package.json#L53) to [use the EnvironmentPlugin](https://github.com/folio-org/ui-eholdings/blob/6bc7a7e0502d78572c86369d7f8f787279a3e282/.stripesclirc.js#L36) for Mirage. Upgrading webpack inside of ui-eholdings resolves the issue.  `yarn upgrade webpack@^4`

cherewaty:
Upgrading `ui-eholdings`' requested `webpack` to `4.0.0` fixed the `mirage` problem. Tests passed too, so I can give this PR a big ol' 👍 .

PR #354, opened 2018-07-09
Upgrade to eslint-config-stripes 2.0.3

cherewaty:
Follow-up to https://github.com/folio-org/stripes-core/pull/353: we shouldn't have any more surprise rule changes with `eslint-config-stripes`, thanks to https://github.com/folio-org/eslint-config-stripes/pull/33

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,7 +33,7 @@
     ]
   },
   "devDependencies": {
-    "@folio/eslint-config-stripes": "^2.0.0",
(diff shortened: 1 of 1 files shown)


cherewaty:
@ttamj good call, https://www.npmjs.com/package/eslint-config-stripes is out in the public NPM registry, but without tags newer than 1.0 pushed.

PR #358, opened 2018-07-18
STCOR-208: pass redux args to stripes-connect

rjruizes:
The [`connect()`](https://github.com/folio-org/stripes-connect/blob/master/connect.js#L187) method currently uses legacy context API to get two redux arguments, `addReducer` and `store`.

My first attempt was to use the new context API and using a context consumer to wrap `Wrapper`, but it did not work. `Requests` would get context, but `NotificationsMenu` could not get context. Only if I made the context provider the immediate parent to `connectedNotificationsMenu` would it get context. 

Instead, I add the redux args to the `connectFor` curry function.


MikeTaylor:
Obviously we need to sort out the issues with the related stripes-connect PR before we can think clearly about this one. But the big deal is that that we've got to be much more careful about compatibility. We can't tell which versions of stripes-connect might have to work alongside which versions of stripes-core, so we need to take measures to ensure that advancing either one of them past their present state will not prevent them working with the present version of the other.

rjruizes:
Closed in favor of https://github.com/folio-org/stripes-core/pull/360

PR #361, opened 2018-07-24
UIU-563 - Access Change my password from Folio Top Toolbar

natali-zaitseva:
* modified user profile pop-up to include "Change Password" item
* added function to check user credentials exists in local storage

### Need to be merged along with https://github.com/folio-org/ui-myprofile/pull/3

### DIFF
--- a/src/components/MainNav/MyProfile/MyProfile.js
+++ b/src/components/MainNav/MyProfile/MyProfile.js
@@ -14,8 +14,10 @@ import List from '@folio/stripes-components/lib/List';
 import NavDropdownMenu from '../NavDropdownMenu';
 import NavButton from '../NavButton';
 import css from './MyProfile.css';
+import { withModules } from '../../Modules';
+import userDropdownChecks from '../../../userDropdownLinksService';
 
-export default class MyProfile extends Component {
+class MyProfile extends Component {
   static propTypes = {
     stripes: PropTypes.shape({
       user: PropTypes.shape({
@@ -26,7 +28,12 @@ export default class MyProfile extends Component {
         showPerms: PropTypes.bool,
         showHomeLink: PropTypes.bool,
       }),
+      okapi: PropTypes.object,
     }).isRequired,
+    modules: PropTypes.shape({
+      app: PropTypes.array,
+      settings: PropTypes.array,
+    }),
     onLogout: PropTypes.func.isRequired,
   };
 
@@ -39,12 +46,41 @@ export default class MyProfile extends Component {
 
     this.state = {
       dropdownOpen: false,
+      userLinks: [],
     };
 
     this.toggleDropdown = this.toggleDropdown.bind(this);
     this.getDropdownContent = this.getDropdownContent.bind(this);
     this.getUserData = this.getUserData.bind(this);
     this.getProfileImage = this.getProfileImage.bind(this);
+  }
+
+  componentDidMount() {
+    const { modules, stripes } = this.props;
+    const userDropdownLinks = ([].concat(modules.app, modules.settings))
+      .filter(({ links }) => links && Array.isArray(links.userDropdown))
+      .reduce((result, { links }) => result.concat(links.userDropdown), []);
+
+    userDropdownLinks.forEach((link, index) => {
+      const linkFunction = link.check;
+      if (!linkFunction) {
+        this.createLink(link, index);
+      } else if (typeof userDropdownChecks[linkFunction] === 'function') {
+        if (userDropdownChecks[linkFunction](stripes)) {
+          this.createLink(link, index);
+        }
+      }
+    });
+  }
+
+  createLink(link, index) {
+    const buttonId = `clickable-menuItem${index}`;
+    const newItem = (
+      <NavListItem id={buttonId} key={buttonId} type="button" onClick={() => this.navigateByUrl(link.route)}>
+        <FormattedMessage id={link.caption} />
+      </NavListItem>
+    );
+    this.setState({ userLinks: this.state.userLinks.concat(newItem) });
   }
 
   toggleDropdown() {
@@ -69,9 +105,13 @@ export default class MyProfile extends Component {
     return (<Avatar alt={user.name} title={user.name} />);
   }
 
-  onHome = () => {
+  navigateByUrl = (url) => {
     this.toggleDropdown();
-    this.context.router.history.push('/');
+    this.context.router.history.push(url);
+  }
+
+  onHome = () => {
+    this.navigateByUrl('/');
   }
 
   getDropdownContent() {
@@ -117,6 +157,7 @@ export default class MyProfile extends Component {
                   <FormattedMessage id="stripes-core.front.home" />
                 </NavListItem>
             }
+            {this.state.userLinks}
             <NavListItem id="clickable-logout" type="button" onClick={onLogout}>
               <FormattedMessage id="stripes-core.logout" />
             </NavListItem>
@@ -139,3 +180,5 @@ export default class MyProfile extends Component {
     );
   }
 }
+
+export default withModules(MyProfile);
--- a/doc/dev-guide.md
+++ b/doc/dev-guide.md
@@ -122,6 +122,12 @@ A module is presented as an [NPM package](https://en.wikipedia.org/wiki/Npm_(sof
 
 * `queryResource` -- if defined, this is the name of an anointed stripes-connect resource whose contents always reflect the query parameters of the URL, and which can be mutated to change the URL. See [URL navigation](#url-navigation) below.
 
+* `links` -- object which is collecting module specific links joint by area of addition. For instance `userDropdown` defines the menu under user profile icon.
+  * `userDropdown` -- defines the list of the links under user profile icon.
+    * `caption` -- the label of the item.
+    * `route` -- partial URL where navigate to.
+    * `check` -- if defined, this is the name of a function which checks whether to show item or not.
+
 When a user enters an application module, its top-level component -- usually `index.js` is executed, and whatever it exports is invoked as a React component. When a user enters a settings module or the settings of an application module, that same component is invoked, but now with the special `showSettings` property set true.
 
 The class exported by a module may also have a static data member, `actionNames`. If provided, this must be an array of strings, each of them the name of an action that can be invoked by hot-keys (see [below](#enabling-hot-keys)). Stripes will aggregate the action-names exposed by the available modules and provide a combined list as [the `actionNames` member](#actionNames) of the Stripes object.
--- a/src/userDropdownLinksService.js
+++ b/src/userDropdownLinksService.js
@@ -1,0 +1,8 @@
+const userDropdownChecks = {
+
+  isLocalLogin(props) {
+    return !props.okapi.ssoEnabled;
+  }
+};
+
+export default userDropdownChecks;


src/components/MainNav/MyProfile/MyProfile.js
```diff
@@ -47,6 +55,32 @@ export default class MyProfile extends Component {
...
+    if (links && Array.isArray(links)) {
+      const linkItems = links[0].userDropdown;
+
+      linkItems.forEach((item, index) => {
+        this.isShownToUser().then((response) => {
+          const showToUser = response.status === 200;
+
+          if (showToUser) {
+            const itemCaption = item.caption;
+            const buttonId = 'clickable-chagepw' + index;
```
MikeTaylor: Hardwiring `changepw` into the ID can't be right -- this is a general-purpose facility for adding links into the user dropdown menu, and it merely _happens_ that we are using it at the moment to implement a change-password link. Something like `'clickable-usermenu' + index` would be better.

src/components/MainNav/MyProfile/MyProfile.js
```diff
@@ -47,6 +55,32 @@ export default class MyProfile extends Component {
...
   }
 
+  componentDidMount() {
+    const links = this.props.modules.settings.map(setting => setting.links);
+
+    if (links && Array.isArray(links)) {
+      const linkItems = links[0].userDropdown;
+
+      linkItems.forEach((item, index) => {
+        this.isShownToUser().then((response) => {
```
MikeTaylor: Woah -- we're doing an HTTP request for each item in the menu, every time we render the menu? That _can't_ be right. Dump all this stuff, and simply only add the contents of the `links` array if the user identity is already known.

MikeTaylor: ... i.e. if `this.props.stripes.user` is defined.

src/components/MainNav/MyProfile/MyProfile.js
```diff
@@ -74,6 +108,20 @@ export default class MyProfile extends Component {
     this.context.router.history.push('/');
   }
 
+  isShownToUser = () => {
```
MikeTaylor: I don't think we need this function at all.

src/components/MainNav/MyProfile/MyProfile.js
```diff
@@ -47,6 +55,32 @@ export default class MyProfile extends Component {
...
+
+      linkItems.forEach((item, index) => {
+        this.isShownToUser().then((response) => {
+          const showToUser = response.status === 200;
+
+          if (showToUser) {
+            const itemCaption = item.caption;
+            const buttonId = 'clickable-chagepw' + index;
+            const newItem = (
+              <NavListItem id={buttonId} type="button" onClick={() => this.onChangePasswordHandler(item)}>
```
MikeTaylor: Similarly, we can't wire an invocation of a change-password-handler into the user-menu! All we can do here -- and all we need to do -- is redirect to the URL that' been configured in the `items` array.

MikeTaylor:
I think the problems here go much deeper than your specific suggestions indicate, @VictorSoroka1. As noted in my earlier comments, the whole approach seems mistaken. We can and should do this in a much simpler way.

src/components/MainNav/MyProfile/MyProfile.js
```diff
@@ -74,6 +108,20 @@ export default class MyProfile extends Component {
     this.context.router.history.push('/');
   }
 
+  isShownToUser = () => {
+    const { url } = this.props.stripes.okapi;
```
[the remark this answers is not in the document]
natali-zaitseva: @VictorSoroka1 changed.

src/components/MainNav/MyProfile/MyProfile.js
```diff
@@ -74,6 +108,20 @@ export default class MyProfile extends Component {
...
+    const { tenant } = this.props.stripes.okapi;
+    const { token } = this.props.stripes.okapi;
+    const { username } = this.props.stripes.user.user;
+
+    return requestUserCredentialsExistance(url, tenant, token, username);
+  };
+
+  onChangePasswordHandler = (item) => {
```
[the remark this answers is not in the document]
natali-zaitseva: renamed

src/components/MainNav/MyProfile/MyProfile.js
```diff
@@ -47,6 +55,32 @@ export default class MyProfile extends Component {
...
 
+  componentDidMount() {
+    const links = this.props.modules.settings.map(setting => setting.links);
+
+    if (links && Array.isArray(links)) {
+      const linkItems = links[0].userDropdown;
+
+      linkItems.forEach((item, index) => {
+        this.isShownToUser().then((response) => {
+          const showToUser = response.status === 200;
```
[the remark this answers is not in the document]
natali-zaitseva: moved

src/components/MainNav/MyProfile/MyProfile.js
```diff
@@ -47,6 +55,32 @@ export default class MyProfile extends Component {
     this.getProfileImage = this.getProfileImage.bind(this);
   }
 
+  componentDidMount() {
+    const links = this.props.modules.settings.map(setting => setting.links);
```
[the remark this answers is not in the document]
natali-zaitseva: done

src/components/MainNav/MyProfile/MyProfile.js
natali-zaitseva: changed

src/components/MainNav/MyProfile/MyProfile.js
natali-zaitseva: changed

src/components/MainNav/MyProfile/MyProfile.js
```diff
@@ -26,7 +28,12 @@ export default class MyProfile extends Component {
         showPerms: PropTypes.bool,
         showHomeLink: PropTypes.bool,
       }),
+      okapi: PropTypes.object,
     }).isRequired,
+    modules: PropTypes.shape({
+      app: PropTypes.array,
```
VictorSoroka1: Do we really need `app` propType specified in PropTypes?

src/components/MainNav/MyProfile/MyProfile.js
```diff
@@ -47,6 +55,38 @@ export default class MyProfile extends Component {
...
+              }
+            });
+          }
+        }
+      });
+    }
+  }
+
+  createLink(link, index) {
+    const buttonId = 'clickable-menuItem' + index;
```
VictorSoroka1: Let`s use es6 template strings. Change to

```
const buttonId = `clickable-menuItem${index}`;
```

src/components/MainNav/MyProfile/MyProfile.js
```diff
@@ -39,6 +46,7 @@ export default class MyProfile extends Component {
 
     this.state = {
       dropdownOpen: false,
+      userLinks: []
```
VictorSoroka1: Please add `,` at the end of the line per consistency.

src/components/MainNav/MyProfile/MyProfile.js
MikeTaylor: Sure we do, if we're accessing `props.modules.app`.

src/components/MainNav/MyProfile/MyProfile.js
MikeTaylor: This is the kind of thing that ESLint usually picks up. @natali-zaitseva, did you run `yarn lint` on your modified code?

src/components/MainNav/MyProfile/MyProfile.js
MikeTaylor: (After reading on ...)

OK, I see that we do _not_ access `props.modules.app`. So you were right to question this, @VictorSoroka1.

MikeTaylor: But ...

A better question would be _why_ we don't use `props.modules.app`. I think we should be extracting links from modules of type `app` as well as those of type `settings`. That entails a change to line 60, below.

src/components/MainNav/MyProfile/MyProfile.js
```diff
@@ -47,6 +55,38 @@ export default class MyProfile extends Component {
     this.getProfileImage = this.getProfileImage.bind(this);
   }
 
+  componentDidMount() {
+    const { modules } = this.props;
+    const navLinks = modules.settings.map(({ links }) => links);
```
MikeTaylor: To include entries from both app and settings-style modules, use something like
```
    const navLinks = ([].concat(modules.app, modules.settings)).map(({ links }) => links);
```

src/components/MainNav/MyProfile/MyProfile.js
```diff
@@ -47,6 +55,38 @@ export default class MyProfile extends Component {
...
+
+    if (navLinks && Array.isArray(navLinks)) {
+      const links = navLinks[0].userDropdown;
```
MikeTaylor: Why are we looking only at the first element of `navLinks`? (I'm not saying this is wrong, just not following the logic yet.)

src/components/MainNav/MyProfile/MyProfile.js
```diff
@@ -47,6 +55,38 @@ export default class MyProfile extends Component {
...
+      links.forEach((link, index) => {
+        if (this.props.stripes.user) {
+          if (!link.check) {
+            this.createLink(link, index);
+          } else if (typeof userDropodownChecks[link.check] === 'function') {
+            userDropodownChecks[link.check]().then((canShow) => {
```
MikeTaylor: Do we really need an async API with promises here? Why not just a simple function returning a boolean?

src/components/MainNav/MyProfile/MyProfile.js
VictorSoroka1: Makes sense.

src/components/MainNav/MyProfile/MyProfile.js
VictorSoroka1: This call is reasonable in order to check that the user has a local password. Without an async call, there is no way to check this.

src/components/MainNav/MyProfile/MyProfile.js
VictorSoroka1: Yep, that should be changed to not only check the first module with the links object.

VictorSoroka1: Yep that should be changed to not only handle the first module with `links`, but every of them

natali-zaitseva: you are correct. changed

src/components/MainNav/MyProfile/MyProfile.js
natali-zaitseva: thanks, updated the code

src/components/MainNav/MyProfile/MyProfile.js
natali-zaitseva: added

MikeTaylor:
Please hold off on merging this till I've had a chance to read.

src/components/MainNav/MyProfile/MyProfile.js
```diff
@@ -47,6 +55,38 @@ export default class MyProfile extends Component {
...
+      const userDropdownLinks = ([].concat(modules.app, modules.settings))
+        .filter(({ links }) => links && Array.isArray(links.userDropdown))
+        .reduce((result, { links }) => result.concat(links.userDropdown), []);
+
+      userDropdownLinks.forEach((link, index) => {
+        const linkFunction = link.check;
+        if (!linkFunction) {
+          this.createLink(link, index);
+        } else if (typeof userDropodownChecks[linkFunction] === 'function') {
+          userDropodownChecks[linkFunction]().then((canShow) => {
```
MikeTaylor: I still don't understand why we've introduced the complexity of asynchronous functions here, with the promises API and the consequent re-rendering. Surely this is a simple boolean? Why can't we just say
```
  if (userDropodownChecks[linkFunction]()) {
    this.createLink(link, index);
  }
```
?

VictorSoroka1: One of the requirements of the folio-org/ui-myprofile#3 is to check that user has its local password. We check it by making an async call to the server. Otherwise it is not possible.

MikeTaylor: It is not acceptable to make a call to the server every time we generate the user menu.

And I am 100% certain it's not necessary. Do an SSO login in one session and a direct login in another. Poke around inside the two sessions' Stripes objects. I bet you'll find a difference.

And if you don't we can _make_ one, just by having the SSO login procedure set `stripes.thidsIsAnSSOSession` to true.

VictorSoroka1:
Natalia, please also update `/docs/dev-guide.md` with a new `links` option for modules.

natali-zaitseva:
@ttamj or somebody who has rights please merge this pull request

PR #365, opened 2018-08-01
Add My Profile to pull script. Part of STRIPES-541

VictorSoroka1:
https://issues.folio.org/browse/STRIPES-541

### DIFF
--- a/util/pull-stripes
+++ b/util/pull-stripes
@@ -30,6 +30,7 @@ dirs=`echo '
     ui-trivial
     ui-vendors
     ui-tags
+    ui-myprofile
     eslint-config-stripes
 '`
 


skomorokh:
Alternatively, I think we should actually remove this script since AFAIK it's deprecated in favour of stripes-cli with `stripes platform pull`

VictorSoroka1:
I noticed that there is similar comment merged recently - https://github.com/folio-org/stripes-core/pull/357/files.

Also the pull script is specified as part of the guide - https://github.com/folio-org/stripes-core/blob/master/doc/dev-guide.md

mkuklis:
@skomorokh some people still use the `pull-stripes` directly. I think it's fine to merge this one but perhaps we should fix the docs too and add a little note about the deprecation?

PR #367, opened 2018-08-03
Select service point preference when logging in. Fixes STCOR-235

mkuklis:
https://issues.folio.org/browse/STCOR-235

Guys this is kind of a continuation of work @doytch did in `ui-users` in https://github.com/folio-org/ui-users/pull/451

I'm currently adding the preferred service point to `currentUser` but perhaps this should be available directly from `stripes`. What do you think?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 * New context API, RootContext. Legacy context API should be removed in next major version. Available from v2.10.4; STCOR-208
 * Update react-intl-safe-html version. STRIPES-545
 * Add _all_ properties to `stripesShape`. Fixes STCOR-236. Available from v2.10.5.
+* Automatically select service point preference when logging in. Fixes STCOR-235.
 
 ## [2.10.0](https://github.com/folio-org/stripes-core/tree/v2.10.0) (2018-06-06)
 [Full Changelog](https://github.com/folio-org/stripes-core/compare/v2.9.0...v2.10.0)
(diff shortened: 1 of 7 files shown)


doytch:
So I think there's a bit of a nomenclature issue here. The problem is that there's a lot of references to "preferred service point" where a more appropriate term would be "current service point" or something like that. Reason is that "preferred service point" is an entry into the `service-points-users` table which, like you implemented, is automatically selected when logging in. However, UIU-547 will let users manually select another service point to be working from when they're logged in. So I think showing _that_ service point in an area with variables named some variation on "service point preference" is going to get confusing pretty quickly.

doytch:
Oops, forgot about your question on placement. The way my brain works, I think it makes sense to keep it to `currentUser`. The preferred service point (and in the near-future, array of service points the user can select for a session) is obviously tightly related to the user themselves, not to the overall stripes setup. 

Second, putting it on `stripes` could lead to confusion about whether an array of service points is the total number of service points in the system, or the current user's service points. I had to get around this in parts of my code through overlySpecificVariableNames that I'd prefer to avoid using in something like the `stripes` object.

Overall, I don't think it saves us much effort to put it right onto `stripes` and the lack of logical scoping  hurts more IMO.

mkuklis:
Thank you both!

@aditya-matukumalli good catch about the spread operator. I will clean that up.

@doytch thanks for the suggestion about renaming it to `current service point`. That makes more sense. Also yes you are right this will need to be refactored when I start on UIU-547 but for not I just want to get this in :stuck_out_tongue_winking_eye:

Also I will keep `curServicePoint` under `currentUser` for now.

PR #372, opened 2018-08-09
Fix re-mounting in settings

mkuklis:
@doytch this PR moves the `connect` in `<Settings/>` from render into constructor. Also while chasing multiple re-mounting I noticed that switching from `component` to `render` in `TitledRoute` limits it to a single remount.

### DIFF
--- a/src/components/Settings/Settings.js
+++ b/src/components/Settings/Settings.js
@@ -24,77 +24,87 @@ const getSettingsModules = (modules) => (
   )
 );
 
-const Settings = (props) => {
-  const stripes = props.stripes;
-  const settingsModules = getSettingsModules(props.modules);
+class Settings extends React.Component {
+  static propTypes = {
+    stripes: stripesShape.isRequired,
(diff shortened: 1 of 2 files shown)


doytch:
👌

zburke:
Wow, nice sleuthing here. In hindsight it is obvious that using `connect` in a stateless component effectively puts it `render` where it does not belong. Do you think we should include such a warning in the [stripes-connect docs](/folio-org/stripes-connect/blob/master/doc/api.md)?

MikeTaylor:
I guess we could -- but as you say, it _should_ have been obvious :-)

I mean, do we want to list _all_ the things that shouldn't go in functional components?

(This kind of thing is one of many reasons why I am not quite such a big fan of functional components as the wider React community seems to think I should be.)

mkuklis:
@zburke I think it would be good to add that warning for new comers. I think this is perhaps obvious for us now but I remember when I started nothing was :)

PR #373, opened 2018-08-09
Update translation strings as of 20180809T174646

folio-translations:


folio-translations:
Closing older pull request.

PR #381, opened 2018-08-22
Introduce handler modules. Part of STCOR-240

mkuklis:
@MikeTaylor this is a first piece which executes handler modules after login. Please let me know what you think. This should be close to what we were discussing (I hope) and what you have outlined in the email. 

I'm also in the process of creating a first handler module called `ui-servicepoints` It would be nice to get this merged so we can move forward with the service points work.

### DIFF
--- a/src/RootWithIntl.js
+++ b/src/RootWithIntl.js
@@ -17,6 +17,7 @@ import Front from './components/Front';
 import SSOLanding from './components/SSOLanding';
 import SSORedirect from './components/SSORedirect';
 import Settings from './components/Settings/Settings';
+import Handlers from './components/Handlers';
 import TitleManager from './components/TitleManager';
 import LoginCtrl from './components/Login';
 import OverlayContainer from './components/OverlayContainer';
@@ -53,6 +54,7 @@ class RootWithIntl extends React.Component {
                     <MainContainer>
                       <OverlayContainer />
                       <MainNav stripes={stripes} />
+                      <Handlers stripes={stripes} />
                       { (stripes.okapi !== 'object' || stripes.discovery.isFinished) && (
                         <ModuleContainer id="content">
                           <Switch>
--- a/src/RootWithIntl.js
+++ b/src/RootWithIntl.js
@@ -24,6 +24,7 @@ import OverlayContainer from './components/OverlayContainer';
 import getModuleRoutes from './moduleRoutes';
 import { stripesShape } from './Stripes';
 import { StripesContext } from './StripesContext';
+import events from './events';
 
 class RootWithIntl extends React.Component {
   static propTypes = {
@@ -54,7 +55,7 @@ class RootWithIntl extends React.Component {
                     <MainContainer>
                       <OverlayContainer />
                       <MainNav stripes={stripes} />
-                      <Handlers stripes={stripes} />
+                      <Handlers event={events.LOGIN} stripes={stripes} />
                       { (stripes.okapi !== 'object' || stripes.discovery.isFinished) && (
                         <ModuleContainer id="content">
                           <Switch>
--- a/src/components/Handlers/Handlers.js
+++ b/src/components/Handlers/Handlers.js
@@ -6,6 +6,8 @@ import { stripesShape } from '../../Stripes';
 class Handlers extends React.Component {
   static propTypes = {
     stripes: stripesShape.isRequired,
+    event: PropTypes.number,
+    data: PropTypes.object,
     modules: PropTypes.shape({
       handler: PropTypes.array,
     }),
@@ -14,13 +16,13 @@ class Handlers extends React.Component {
   constructor(props) {
     super(props);
 
-    const { stripes, modules } = props;
+    const { event, stripes, modules, data } = props;
 
-    this.handlers = modules.handler.reduce((acc, m) => {
+    this.components = modules.handler.reduce((acc, m) => {
       const module = m.getModule();
-      const handler = module[m.handlerName];
-      if (!handler) return acc;
-      const component = handler(stripes);
+      const eventHander = module[m.handlerName];
+      if (!eventHander) return acc;
+      const component = eventHander(event, stripes, data);
       if (component) {
         acc.push(stripes.connect(component));
       }
@@ -29,8 +31,9 @@ class Handlers extends React.Component {
   }
 
   render() {
-    const { stripes } = this.props;
-    return (this.handlers.map(Component => (<Component stripes={stripes} />)));
+    const { stripes, data } = this.props;
+    return (this.components.map(Component =>
+      (<Component key={Component.name} stripes={stripes} data={data} />)));
   }
 }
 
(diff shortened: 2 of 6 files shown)


src/components/Handlers/Handlers.js
```diff
@@ -0,0 +1,40 @@
...
+    this.components = modules.handler.reduce((acc, m) => {
+      const module = m.getModule();
+      const eventHander = module[m.handlerName];
+      if (!eventHander) return acc;
+      const component = eventHander(event, stripes, data);
+      if (component) {
+        acc.push(stripes.connect(component));
+      }
+      return acc;
+    }, []);
```
MikeTaylor: (Side-note: it always seems stupid to me that the arguments to `reduce` are this way around. What it means is that every single use of it has a long-ass closure, and then dangling off the end of it, a vestigial `[]` or `{}`. It's most enough to make one define `Array.prototype.ecuder = (acc, fn) => reduce(fn, acc)`.)

src/RootWithIntl.js
```diff
@@ -53,6 +55,7 @@ class RootWithIntl extends React.Component {
                     <MainContainer>
                       <OverlayContainer />
                       <MainNav stripes={stripes} />
+                      <Handlers event={events.LOGIN} stripes={stripes} />
```
MikeTaylor: Won't this fire on _every_ render, rather than only when the user has just logged in?

src/components/Handlers/Handlers.js
```diff
@@ -0,0 +1,40 @@
...
+      handler: PropTypes.array,
+    }),
+  };
+
+  constructor(props) {
+    super(props);
+
+    const { event, stripes, modules, data } = props;
+
+    this.components = modules.handler.reduce((acc, m) => {
```
MikeTaylor: This is delightfully simple and elegant. Seeing the code makes me more confident that we got the abstraction right.

src/components/ModuleTranslator/ModuleTranslator.js
```diff
@@ -23,6 +23,7 @@ class ModuleTranslator extends React.Component {
       app: (originalModules.app || []).map(this.translateModule),
       plugin: (originalModules.plugin || []).map(this.translateModule),
       settings: (originalModules.settings || []).map(this.translateModule),
+      handler: (originalModules.handler || []).map(this.translateModule),
```
MikeTaylor: What does this do? What kind of thing is `originalModules.handler`? A string, a function? From its use in `<Handlers>`, it _looks_ like it's the module structure -- in which case does it make any sense to translate it? (I realise this is a comment not so much on your code as it is on my own ignorance.)

src/components/Modules/withModule.js
```diff
@@ -13,7 +13,8 @@ const findModule = (modules, moduleName) => {
   return (
     findModuleInArray(modules.app, moduleName) ||
     findModuleInArray(modules.plugin, moduleName) ||
-    findModuleInArray(modules.settings, moduleName)
+    findModuleInArray(modules.settings, moduleName) ||
+    findModuleInArray(modules.handler, moduleName)
```
MikeTaylor: This function is starting to looks silly. Do we really need an explicit enumeration, in code, of all the supported module types? Surely we can just just walk through the keys of `modules`?

(Not that this is really the concern of the present PR. Just a matter of hygiene.)

mkuklis:
Thanks for your comments @MikeTaylor 

I maybe missing something here but I don't see a clean way right now to hook this up right after login. We would have to do it somewhere here:

https://github.com/folio-org/stripes-core/blob/master/src/loginServices.js

or perhaps here (after successful login):

https://github.com/folio-org/stripes-core/blob/master/src/components/Login/LoginCtrl.js#L42

but how would that look like? We want to return a list of (ideally connected) components which would be attached to the react tree (after login).

...

Now when I thought about it a bit more perhaps one way could be be to check if the `currentUser` or `token` have changed in

https://github.com/folio-org/stripes-core/blob/master/src/RootWithIntl.js

via `getDerivedStateFromProps` and if it did call some sort of function to get all available handler components. Does that sound good to you?

Do you have any other ideas?

src/components/Modules/withModule.js
mkuklis: Yes this is definitely something we can clean up.

src/components/ModuleTranslator/ModuleTranslator.js
mkuklis: @MikeTaylor I'm not sure either. I had to add this in order to actually get the `handler` in a lower level components. I think it happensbecause the ModuleContext it's being overwritten `ModuleTranslator` for some reason.

https://github.com/folio-org/stripes-core/blob/9d33e12950b322ae2a4ca23721168d7658df3808/src/components/ModuleTranslator/ModuleTranslator.js#L41

But this is really not related to our `handler` work here. We can clean that up later.

src/components/Handlers/Handlers.js
mkuklis: Yes I agree. `reduce` always looks ugly even when you try to make it look pretty. Ruby has a couple nice versions which I always liked: 
https://ruby-doc.org/core-2.1.0/Enumerable.html#method-i-reduce

MikeTaylor: Ruby is, without a doubt, the cleanest language I've used extensively. It's not just that you can say so many things in it, it's that they all feel natural. The bitter but honest truth is that I never really felt comfortable around closures until I realised I'd been using them all the time in Ruby without even realising it: see https://reprog.wordpress.com/2010/02/27/closures-finally-explained/ if interested.

src/components/ModuleTranslator/ModuleTranslator.js
MikeTaylor: OK, having looked at the stripes-core code, I realise what's going on here. This is another instance, like in `findModule`, where we're just manually iterating over the different known kinds of module. So, yes, what you've done makes perfect sense; and it's open to a nicer refactoring like what you did to `findModule` but that's not _necessary_.

MikeTaylor:
> The handler execution will only happen once in the Handlers constructor and only after user is already logged in.

Well, that is great news.

But ... why?!

I mean, I am delighted. But also confused. But you need not wait until you have unconfused me before going ahead and merging. Again, you know, this is really nice code.

mkuklis:
I think this is a nice property of react's reconciliation algorithm. The component will be recreated completely if certain conditions are met (for example component's key has changed) but most of the time it will just re-render (without reinitialization).

MikeTaylor:
"A nice property of react's reconciliation algorithm" sounds worryingly like "it happens to work out right by accident, I hope that stays true" :-)

zburke:
I'm starting to wonder if they named it "React" not because the DOM will automatically react to user events but because developers must constantly react to whatever latest-favorite-shiny-thing-but-watch-out-we-might-change-the-implementation-in-the-next-release thing is happening at Facebook.

MikeTaylor:
Ha ha, only serious.

PR #383, opened 2018-08-22
Handle sso check exception correctly

mkuklis:


MikeTaylor:
Beautifully concise. It's always gratifying when a fix is this simple.

PR #384, opened 2018-08-23
Add ability to call handlers from MyProfile. Part of STCOR-244

mkuklis:
@MikeTaylor this is a continuation of work which should allow us for calling handlers from `MyProfile` component (or any other place).

In the `MyProfile` case the handler will execute for each link with the `event` present in the `userDropdown` config:

```
   .....
    "links": {
      "userDropdown": [
        {
          "caption": "ui-servicepoints.userDropdown.switchServicePoint",
          "event": "EVENT_NAME"
        }
      ]
    }
``` 

Also I had to refactor the previous implementation of `Handlers` into a HOC in order to extract the code from the constructor into a separate function.


src/components/Handlers/withHandlers.js
```diff
@@ -0,0 +1,39 @@
...
+    const component = getComponentFromHandler(event, stripes, m, data);
+    if (component) {
+      acc.push(stripes.connect(component));
+    }
+    return acc;
+  }, []);
+}
+
+export default function withHandlers(WrappedComponent) {
+  class WithHandlers extends React.Component {
```
MikeTaylor: I find it a bit confusing that you're using the same name for the class and the function that creates it.

src/components/Handlers/withHandlers.js
```diff
@@ -0,0 +1,39 @@
...
+      return (
+        <WrappedComponent
+          getComponentsFromHandlers={getComponentsFromHandlers}
+          getComponentFromHandler={getComponentFromHandler}
+          {...this.props}
+        />
+      );
+    }
+  }
+  WithHandlers.displayName = `WithHandlers(${getDisplayName(WrappedComponent)})`;
```
MikeTaylor: This is a nice, polite touch. Not sure a separate `getDisplayName` function is warranted, but that's no big.

src/components/Handlers/Handlers.js
```diff
@@ -2,32 +2,23 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { withModules } from '../Modules';
 import { stripesShape } from '../../Stripes';
+import withHandlers from './withHandlers';
```
MikeTaylor: I guess I'm a little uneasy about this being called `withHandlers` when it doesn't put handlers into the props. It ought really be have a clumsy name like `withGetComponentsFromHandlers` since that's what it provides. I can see why you didn't do that, but in the long run we might be better served by explicitness than by terseness, despite our justified fear of ending up with Java-style names like `PutFixedDueDateScheduleStorageFixedDueDateSchedulesByFixedDueDateScheduleIdResponse`.

src/components/Handlers/withHandlers.js
```diff
@@ -0,0 +1,39 @@
+import React from 'react';
+
+function getDisplayName(WrappedComponent) {
+  return WrappedComponent.displayName || WrappedComponent.name || 'Component';
+}
+
+function getComponentFromHandler(event, stripes, module, data) {
+  const m = module.getModule();
+  const eventHander = m[module.handlerName];
```
MikeTaylor: (I should have asked this when reviewing the previous PR, but don't forget to document `handlerName` in the Stripes developer guide.)

src/components/MainNav/MyProfile/MyProfile.js
```diff
@@ -34,53 +33,60 @@ class MyProfile extends Component {
...
+    const { stripes, getComponentFromHandler } = this.props;
+    const { check, event } = link;
+    const buttonId = `${kebabCase(module.displayName)}-clickable-menuItem${index}`;
+
+    if (!check || (isFunction(validations[check]) && validations[check](stripes))) {
+      if (event) {
+        const HandlerComponent = getComponentFromHandler(event, stripes, module);
+        return (<HandlerComponent key={buttonId} stripes={stripes} />);
+      } else {
+        return this.renderNavLink(link, buttonId);
```
MikeTaylor: I like how, in passing, you've simplified how this gets done, too.

src/components/Handlers/Handlers.js
```diff
@@ -2,32 +2,23 @@ import React from 'react';
...
-      const eventHander = module[m.handlerName];
-      if (!eventHander) return acc;
-      const component = eventHander(event, stripes, data);
-      if (component) {
-        acc.push(stripes.connect(component));
-      }
-      return acc;
-    }, []);
+    const { event, stripes, modules, getComponentsFromHandlers, data } = props;
+    this.components = getComponentsFromHandlers(event, stripes, modules.handler, data);
```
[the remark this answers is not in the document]
mkuklis: Good point @MikeTaylor.

MikeTaylor:
(Oh, and I think the name "MyProfile" is pretty weird for the user-dropdown -- especially as we have a completely different thing, a UI module, that is called MyProfile. But that's not something for the present PR to address.)

src/components/Handlers/withHandlers.js
mkuklis: @MikeTaylor yeah I stole this pattern from a different HOC in `stripes-core`: https://github.com/folio-org/stripes-core/blob/master/src/components/Modules/withModules.js
I wonder perhaps we don't even need a HOC here since all we do is just use these 2 functions.

src/components/Handlers/Handlers.js
mkuklis: @MikeTaylor exactly perhaps HOC is not needed now when I look at it again this morning. Also 100% agree about these long names. Do you have any suggestions for a better name?

mkuklis:
@MikeTaylor thanks for reviewing this. There is still one big issue here. Since we are calling:

```
  const HandlerComponent = getComponentFromHandler(event, stripes, module);
  return (<HandlerComponent key={buttonId} stripes={stripes} />)
```

and rendering it as a part of the dropdown the returned component (from `ui-servicepoints` looks like this:

```
<NavListItem type="button" onClick={() => this.showModal()}>
  <FormattedMessage id="ui-servicepoints.userDropdown.switchServicePoint" />
  <ServicePointsModal stripes={stripes} open={this.state.open} />
</NavListItem>
```

This works well and the modal will be rendered and it opens (it's all good here).
Now if I try to close the dropdown menu (when the modal is open) the modal will be also closed since it's part of the dropdown tree.

I've tried different things here but I'm out of ideas here. Because of this issue I'm still not completely sold on the event handler idea.

MikeTaylor:
I don't know what the solution is, and I can't give it my attention right now because (surprise!) I have to be in a meeting. But this _can't_ be hard to fix. I bet @JohnC-80 or someone will have a solution.

MikeTaylor:
Nice.

PR #389, opened 2018-08-24
Refactor handler component creation. Part of STCOR-237

mkuklis:
@MikeTaylor this approach solves the issue with the nav bar being open (when we open a SP modal). It also aligns with what you had in mind yesterday and actually cleans up the handler component implementation. 

Here are the changes for the handler:

https://github.com/folio-org/ui-servicepoints/pull/2/files

### DIFF
--- a/src/components/MainNav/ProfileDropdown/ProfileDropdown.js
+++ b/src/components/MainNav/ProfileDropdown/ProfileDropdown.js
@@ -93,16 +93,16 @@ class ProfileDropdown extends Component {
     return null;
   }
 
-  onNavItemClick(link, module) {
-    const onClickHandler = (link.event) ? this.createHandlerComponent : this.navigateByUrl;
+  onNavItemClicked(link, module) {
+    const handler = (link.event) ? this.createHandlerComponent : this.navigateByUrl;
     this.toggleDropdown();
-    onClickHandler(link, module);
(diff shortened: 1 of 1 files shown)


src/components/MainNav/ProfileDropdown/ProfileDropdown.js
```diff
@@ -57,33 +59,49 @@ class ProfileDropdown extends Component {
     }, []);
   }
 
+  setInitialState(callback) {
```
MikeTaylor: Remind me, what is the advantage of `setInitialState` over just setting it in the constructor? (I'm not disagreeing, just seeking insight.)

mkuklis: @MikeTaylor normally there is no difference but this case is a bit different. We need to reset the state (specifically `HandlerComponent`) in order to unmount the previously mounted modal (if present). Otherwise react will keep reusing the same modal (same instance) over and over again (which means the constructor of the HandlerComponent will never be called again). The consequence of this is that the SP modal would open for the first time but after you close it and try to open it again nothing would happen.

The alternative approach here could be to use a random `key`. For example:

<HandlerComponent stripes={this.props.stripes} key={`key-${Date.now()}`} />

PR #393, opened 2018-08-29
FOLIO-1422 i18n: Index into locale-data module for locales array

doytch:
I'll explain this backwards:

- Translation strings weren't being shown for non-English languages. We'd either show the label ID (`ui-users.meta.title`) or the original English string. These are the fallback strings.
- The fallback strings were showing because `react-intl` was giving us an empty `messages` array [at this line](https://github.com/yahoo/react-intl/blob/763f3fcbf87abc1a2a1021f684217f56bf8d4581/src/components/provider.js#L134).
- Scroll up from that line and you'll see it was giving us an empty `messages` array because it didn't have locale data for the currently selected locale.
- So why don't we have all the locales loaded into react-intl when we clearly call `addLocaleData` at loginServices.js:36? `import` returns a `Module` object, not the default export. It's behaving like `require` here rather than the `import` we know and love.
- Why is that? [I'm pretty sure it's this breaking change in Webpack 4](https://medium.com/webpack/webpack-4-import-and-commonjs-d619d626b655) that changes the behaviour of `import()`, based on peeking at `node_modules/react-intl/locale-data/de.js`

Now, my read on this suggests that if react-intl changes their locale-data exports to have `__esModule: true` listed on them, we'd need to revert this change. But, this appears to work for now and makes sense. Just something to keep in mind.

### DIFF
--- a/src/loginServices.js
+++ b/src/loginServices.js
@@ -33,7 +33,7 @@ function getHeaders(tenant, token) {
 export function loadTranslations(store, locale) {
   const parentLocale = locale.split('-')[0];
   return import(`react-intl/locale-data/${parentLocale}`)
-    .then(intlData => addLocaleData(intlData))
+    .then(intlData => addLocaleData(intlData.default))
     .then(() => fetch(translations[parentLocale]))
     .then((response) => {
       if (response.ok) {
--- a/src/loginServices.js
+++ b/src/loginServices.js
@@ -33,7 +33,7 @@ function getHeaders(tenant, token) {
 export function loadTranslations(store, locale) {
   const parentLocale = locale.split('-')[0];
   return import(`react-intl/locale-data/${parentLocale}`)
-    .then(intlData => addLocaleData(intlData.default))
+    .then(intlData => addLocaleData(intlData.default || intlData))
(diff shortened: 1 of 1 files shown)


dltj:
Nice bit of work.  Looking forward to what the reviewers say.

doytch:
@zburke Yeah the only quirk is I'm not sure if they could test if they're being imported as an ES module and only set the property then. They definitely would want to maintain compatibility between being imported as a CommonJS and ES6 module. It's been a while since I've looked at the differences, but I remember you needed to jump through a couple of hoops with stuff like NPM package files when publishing both versions of a lib for consumption so I don't know if you can snoop the way you're being imported (specifically for CJS vs ESM, you can do it for AMD etc).
@ttamj Good call. Done.

MikeTaylor:
Amazing work, there, by @doytch -- many thanks for getting to the bottom of this.

I'm also amazed as your ability to dispassionately describe WebPack's backwards-incompatible change to `import` without editorialising. I find this kind of behaviour absolutely astonishing. As fine as ES6 is as an actual programming language, the culture surrounding it is as toxic, in its very different way, from Java culture.

PR #403, opened 2018-09-04
STCOM-331 Move Pluggable

cherewaty:
## Purpose
`Pluggable` doesn't meet the definition of a purely presentational component (https://issues.folio.org/browse/STCOM-298), since it's coupled to `stripes-config`. Its most logical new home is in `stripes-core`.

screen shot 2018-09-04 at 5 07 46 am

https://issues.folio.org/browse/STCOM-331

## Approach
Used `git-subtree` to preserve as much git history for these files as possible from `stripes-components`.

## Next Steps
- [ ] Introduce deprecation warning in `stripes-components`
- [ ] Adjust imports in `folio-org` `ui-*` modules
- [ ] Remove `Pluggable` from `stripes-components` `v4.0.0` branch

### DIFF
--- a/src/Pluggable.js
+++ b/src/Pluggable.js
@@ -26,6 +26,7 @@ const Pluggable = (props) => {
 
   if (!props.children) return null;
   if (props.children.length) {
+    // eslint-disable-next-line no-console
     console.error(`<Pluggable type="${props.type}"> has ${props.children.length} children, can only return one`);
   }
   return props.children;


zburke:
We should bump the patch version when adding features like this to libraries. That'll allow dependent libraries, i.e. any ui-* app that will be looking for `<Pluggable>` in its new location, to set the minimum version of stripes-core they depend on.

MikeTaylor:
Definitely! I'll see to it.

PR #406, opened 2018-09-10
UIU-551: Added actions/reducers to set the current user's service points

doytch:
This is needed for UIU-551. 

We want to show the `ServicePointsModal` from `ui-servicepoints` whenever a user edits themselves in the Users app. This required a few changes:

- New actions/reducers in Redux.
- Exposing those via `loginServices`
- Passing props to a handler component returned from `HandlerManager`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Change history for stripes-core
 
+## [CURRENT]
+
+* Added ability to set the current user's service points in the session object. Enables UIU-551.
+
 ## [2.11.0](https://github.com/folio-org/stripes-core/tree/v2.11.0) (2018-09-04)
 [Full Changelog](https://github.com/folio-org/stripes-core/compare/v2.10.0...v2.11.0)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,8 @@
 # Change history for stripes-core
 
-## [CURRENT]
+## 2.12.0 (IN PROGRESS)
 
-* Added ability to set the current user's service points in the session object. Enables UIU-551.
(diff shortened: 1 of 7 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Change history for stripes-core
 
+## [CURRENT]
```
MikeTaylor: It's more usual to designate this as

> ## 2.12.0 (IN PROGRESS)

Not a big deal.

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
...
+
+* Added ability to set the current user's service points in the session object. Enables UIU-551.
```
MikeTaylor: Add "Available from v2.11.1", so client code that needs this functionality knows which version to depend on.

src/components/HandlerManager/HandlerManager.js
```diff
@@ -13,6 +13,7 @@ class HandlerManager extends React.Component {
     modules: PropTypes.shape({
       handler: PropTypes.array,
     }),
+    props: PropTypes.object,
```
MikeTaylor: Wait -- there's a prop called `props`? That's just _asking_ for confusion. Surely better to call it something like `subcomponentProps`?

doytch: I thought about it, but decided to use the `redux-form` name. It was a choice between what's obvious and memorable and what could be confusing. So since the `props` nomenclature is already in use in stuff we currently use, it broke the tie for me.

MikeTaylor: OK, I guess.

PR #407, opened 2018-09-10
Refactor handler events to strings. Part of UIU-551.

mkuklis:
The reason for using strings instead of numbers is because we use strings here:

https://github.com/folio-org/ui-servicepoints/blob/master/package.json#L20


MikeTaylor:
The you'd better fix the docs, too: https://github.com/folio-org/stripes-core/blob/master/doc/dev-guide.md#handlers-and-events

PR #415, opened 2018-09-12
Use eslint-config-stripes 3.2

cherewaty:
Includes upgrades for `eslint` and `babel-eslint`.

### DIFF
--- a/examples/trivial-okapi/TenantList.js
+++ b/examples/trivial-okapi/TenantList.js
@@ -30,12 +30,9 @@ class TenantList extends Component {
       <li key={tenant.id}>
         {tenant.name}
         {' '}
-        [
         <button type="button" tabIndex={0} onClick={() => mutator.tenants.DELETE(tenant)}>delete</button>
-        ]
-        [
+        {' '}
         <Link to={`${pathname}/edit/${tenant.id}`}>Edit</Link>
-        ]
       </li>
     ));
     return (
--- a/examples/trivial-okapi/TenantList.js
+++ b/examples/trivial-okapi/TenantList.js
@@ -28,8 +28,14 @@ class TenantList extends Component {
     if (!data.tenants) return null;
     const tenantNodes = data.tenants.map(tenant => (
       <li key={tenant.id}>
-        {tenant.name} [<button tabIndex={0} onClick={() => mutator.tenants.DELETE(tenant)}>delete</button>]
-          [<Link to={`${pathname}/edit/${tenant.id}`}>Edit</Link>]
+        {tenant.name}
+        {' '}
+        [
+        <button type="button" tabIndex={0} onClick={() => mutator.tenants.DELETE(tenant)}>delete</button>
+        ]
+        [
+        <Link to={`${pathname}/edit/${tenant.id}`}>Edit</Link>
+        ]
       </li>
     ));
     return (
--- a/src/components/About/AboutEnabledModules.js
+++ b/src/components/About/AboutEnabledModules.js
@@ -33,7 +33,8 @@ class AboutEnabledModules extends React.Component {
 
       return (
         <li key={key} style={style}>
-          {this.props.availableModules[key]} (<tt>{key}</tt>)
+          {this.props.availableModules[key]}
+          {`(<tt>${key}</tt>)`}
         </li>
       );
     };
--- a/src/components/About/AboutEnabledModules.js
+++ b/src/components/About/AboutEnabledModules.js
@@ -34,7 +34,7 @@ class AboutEnabledModules extends React.Component {
       return (
         <li key={key} style={style}>
           {this.props.availableModules[key]}
-          {`(<tt>${key}</tt>)`}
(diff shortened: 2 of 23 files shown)


examples/trivial-okapi/TenantList.js
```diff
@@ -28,8 +28,14 @@ class TenantList extends Component {
...
-        {tenant.name} [<button tabIndex={0} onClick={() => mutator.tenants.DELETE(tenant)}>delete</button>]
-          [<Link to={`${pathname}/edit/${tenant.id}`}>Edit</Link>]
+        {tenant.name}
+        {' '}
+        [
+        <button type="button" tabIndex={0} onClick={() => mutator.tenants.DELETE(tenant)}>delete</button>
+        ]
+        [
+        <Link to={`${pathname}/edit/${tenant.id}`}>Edit</Link>
+        ]
```
wwilsman: Is there a reason for this array notation here?

src/components/About/AboutEnabledModules.js
```diff
@@ -33,7 +33,8 @@ class AboutEnabledModules extends React.Component {
 
       return (
         <li key={key} style={style}>
-          {this.props.availableModules[key]} (<tt>{key}</tt>)
+          {this.props.availableModules[key]}
+          {`(<tt>${key}</tt>)`}
```
wwilsman: Because of the string interpolation, I think there might be issues with this possibly being escapes internally in React (`dangerouslySetInnerHTML={{ _html }}` exists for this reason).

src/components/MainNav/Notifications/NotificationsMenu.js
```diff
@@ -102,8 +102,8 @@ class NotificationsMenu extends React.Component {
     return (
       <Row key={notification.id} onClick={() => { this.onClickNotification(notification); }} className={this.getNotificationClass(notification.date)}>
         <Col xs>
-          <div><strong>{notification.source} {formattedDate}</strong></div>
-          <p>{text} ({link})</p>
+          {`<div><strong>${notification.source} ${formattedDate}</strong></div>`}
+          {`<p>${text} (${link})</p>`}
```
wwilsman: Same comment here about string interpolation. I don't think JSX would properly be evaluated inside of template literals. And even the native DOM elements are actually special JSX components under the hood.

src/components/SSOLanding.js
```diff
@@ -42,12 +42,18 @@ class SSOLanding extends Component {
     const token = this.getToken();
 
     if (!token) {
-      return <div>No <tt>ssoToken</tt> cookie or query parameter</div>;
+      return (
+        <div>
+          {'No <tt>ssoToken</tt> cookie or query parameter'}
```
wwilsman: [Here's a link to React docs on dangerouslySetInnerHTML](https://reactjs.org/docs/dom-elements.html#dangerouslysetinnerhtml). The concern with using HTML inside of strings (especially with template literals) is about XSS attacks

src/moduleRoutes.js
```diff
@@ -50,7 +50,7 @@ function getModuleRoutes(stripes) {
                       </div>
                     </AddContext>
                   </StripesContext.Provider>);
-                }
+              }
```
wwilsman: Was this intentional? It doesn't look like it's indented properly with the `}` below

src/components/SSOLanding.js
zburke: What's the value of this change (and the others that are similar) to begin with? I don't see how the new code, with string interpolation that causes problems deep in React, is in any way superior to the code it replaced.

src/moduleRoutes.js
cherewaty: Fixed

examples/trivial-okapi/TenantList.js
cherewaty: No. Removed brackets.

src/components/About/AboutEnabledModules.js
cherewaty: Fixed all of these.

src/components/SSOLanding.js
cherewaty: https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-one-expression-per-line.md

The rule is designed to make JSX construction very deliberate.

src/components/About/AboutEnabledModules.js
```diff
@@ -33,7 +33,8 @@ class AboutEnabledModules extends React.Component {
 
       return (
         <li key={key} style={style}>
-          {this.props.availableModules[key]} (<tt>{key}</tt>)
+          {this.props.availableModules[key]}
+          <tt>{`(${key})`}</tt>
```
wwilsman: Only concern here is does this cause a visual regression if `tt` is has alternate font styles? If there are no differences in styling, then ignore this

cherewaty: It will cause a visual change to the parentheses, but not one that matters.

PR #419, opened 2018-09-14
Extract lastVisited into a separate higher order component. Part of STCOR-254

mkuklis:
@zburke and @MikeTaylor this is another attempt to solve [STCOR-254](https://issues.folio.org/browse/STCOR-254) by extracting `lastVisited` into a separate HOC module and make it available via react context. This would allow us to update the `lastVisited` from stripes-form which is illustrated here:

https://github.com/folio-org/stripes-form/pull/32/files

I'm still not sure about some of the naming conventions. For example the context is exposing `goBack` but we are not really going back when we call it. We are just updating `lastVisited` for the current module to the previous path.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
 * Support mod-login-saml's requirement for the new authtoken interface in addition to the old interface. Refs STCOR-76. Available from v2.12.1.
 * Expose timeZone through react-intl provider
 * Export classes and functions intended for external use
-* Extract LastVisited to a separate module. Fixes STCOR-254. Available from v2.12.2.
+* Extract `lastVisited` to a separate higher order component. Fixes STCOR-254. Available from v2.12.2.
 
 ## [2.12.0](https://github.com/folio-org/stripes-core/tree/v2.12.0) (2018-09-13)
 [Full Changelog](https://github.com/folio-org/stripes-core/compare/v2.11.0...v2.12.0)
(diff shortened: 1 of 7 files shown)


zburke:
This is nice work! I wasn't sure how to even _think_ about this problem, let alone solve it, but I can follow the work you've done easily. Nice! 

To your question about the naming convention, I agree that `context.goBack()` doesn't feel right. If this were a java program, we would could just use `context.rememberTheUrlYouWereVisitingBeforeYouNavigatedAwayFromADirtyFormGratuitousUseOfTheFactoryPatternAndAnInterface()` and call it a day, but we can do better! `cachePreviousUrl()` maybe? `logPreviousUrl()`?

Also, I'm not convinced we need to use `history.replace()` instead of `history.push()` in stripes-form. That'll change the behavior of the back button (right?) but I think the only issue, at least as far as I understand STCOM-306, is with returning to an app by clicking its icon in the navigation-menu.

mkuklis:
Thank you both for the comments.

@zburke thank you for suggesting another name. I really like `cachePreviousUrl()`.

Regarding the `push` versus `replace` that sounds good. The only thing about `push` is that when somebody hits 'Close without saving' and then hits back button the form will appear again.

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 * Support mod-login-saml's requirement for the new authtoken interface in addition to the old interface. Refs STCOR-76. Available from v2.12.1.
 * Expose timeZone through react-intl provider
 * Export classes and functions intended for external use
+* Extract LastVisited to a separate module. Fixes STCOR-254. Available from v2.12.2.
```
MikeTaylor: This is picky, but I would not characterise this as a separate _module_ -- a term that has a specific meaning in the world of JavaScript (and indeed another in the world of Stripes).

mkuklis: Good point @MikeTaylor! I will change it now.

PR #420, opened 2018-09-14
Start exporting things

cherewaty:
I poked around the `folio-org` to see what other repos are importing from `stripes-core`. Those have been added here.

Putting these in an index forces us to think about `stripes-core`'s external API.

Encourages:
```
import { TitleManager } from '@folio/stripes-core';
```

instead of 
```
import TitleManager from '@folio/stripes-core/src/components/TitleManager';
```

Consumers of `stripes-core` APIs shouldn't need to know the directory structure.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 * Support mod-login-saml's requirement for the new authtoken interface in addition to the old interface. Refs STCOR-76. Available from v2.12.1.
 * Expose timeZone through react-intl provider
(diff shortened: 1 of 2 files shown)


MikeTaylor:
Thanks for handling this. Much needed.

cherewaty:
Changelog entry added. I think this is a really good candidate worth cutting a minor release for (will make https://issues.folio.org/browse/STCOM-331 go smoothly). I'll open another PR for that.

PR #423, opened 2018-09-17
Release v2.13.0

cherewaty:
* Support mod-login-saml's requirement for the new authtoken interface in addition to the old interface. Refs STCOR-76. Available from v2.12.1.
* Expose timeZone through react-intl provider
 * Export classes and functions intended for external use


cherewaty:
Easy conflict to fix :) I'll get this right out.

PR #426, opened 2018-09-20
[STCOR-251] Add BigTest Infrastructure

wwilsman:
## Purpose

[[STCOR-251] Add BigTest Infrastructure](https://issues.folio.org/browse/STCOR-251)

## Approach

The typical `bigtest` folder structure was dropped in which includes mirage setup and relevant directories, interactors, the tests themselves, and helpers to mount the app under test.

### Testing "Harness"

In UI module repos one of the app mount helpers was the `Harness`. This was first created in eHoldings to work around the nature of `stripes-core/src/index.js` which does some global setup at the root level. For testing, this setup needs to be repeatable, so the `Harness` was created to encapsulate that setup in a reusable way.

For `stripes-core`, we can eliminate the need for this harness by moving all of that setup into a reusable module. This will also eliminate the need for the `Harness` in all other repos once they update to include the new `stripes-core/src/App` module which exports a reusable `StripesCore` component that mounts the entire stripes platform. This also has the benefit of making the `index.js` file **only** responsible for mounting the app during production & development.

Another necessary requirement to make `StripesCore` reusable was to remove the root-level debounced functions within the login services. The two functions were debounced to 5s, which would require the tests to wait every 5s before running another test. I don't know what the intent of the debounced functions were, but if it was to fix an unfortunate bug somewhere, we can now write tests for this and fix it in a more reliable manner.

### Stripes Config

One of the most important things that needs tests is how `stripes-core` handles stripes config objects for various modules. Since these config objects are provided through the virtual `stripes-config` package, we need a way to mock this config for our tests.

This is done via _mutating_ the objects exported by `stripes-config`. Since this virtual package exports plain objects, we can mutate them in our testing hooks (setup & teardown) and provide custom data for `stripes-config`. This is done for `modules` and `config` exported by `stripes-config`, but other mocks might be needed in the future.

### Translations

In UI module tests, the translations endpoint is allowed to pass through the mirage handlers so translations can be provided via `stripes-core`.

Since we're testing `stripes-core`, we need to capture this translations endpoint and send our own translation data needed for our tests. The default translations include `stripes-core`, `stripes-components`, `stripes-smart-components`, and `stripes-form`. Since only `stripes-components` is a dependency of `stripes-core`, those two package's `en` translations are included by default in this mocked endpoint. In the future, we'll probably want to add other languages here as well.

### Existing Tests

There was previously only one suite of tests for core itself, and the other tests were specific to the webpack plugins. The core tests were moved into a `unit` folder within `bigtest/tests` so that they are run in the browser environment where core is consumed.

The webpack tests, of course, cannot be moved into the browser. The preexisting package scripts for running the tests were adjusted to run only the webpack tests. The main `test` script now will run `test:webpack` and `test:core` which runs core tests in a browser via `stripes-cli`.

The webpack tests can be ran alone with `yarn test:webpack` and can be watched with `--watch`. The core tests can be ran with `yarn test:core` and automatically watch for changes unless `--karma.singleRun` is provided.

### Development

Thanks to `stripes-cli`, you can now run `yarn start` to boot up an empty `stripes-core`. You can also provide `--mirage [scenario]` to boot up `stripes-core` with mirage and an optional scenario from `bigtest/network/scenarios`.

## Todo

- [x] Release the next version of `stripes-cli` which allows `stripes test karma` to work with all stripes packages including this one.


wwilsman:
Hmmmm. So the issue with that JenkinsFile for this repo is the `test` command now runs two different types of tests. So we can't pass in options to the `test:webpack` portion but only to the `test:core` portion (the command run is `test:webpack && test:core`).

wwilsman:
Updated the JenkinsFile to run only core tests (and not webpack tests). I moved the previous `test` script to `test-all` to preserve it and just made `test` alias `test:core`.

ttamj:
> One of the most important things that needs tests is how stripes-core handles stripes config objects for various modules. Since these config objects are provided through the virtual stripes-config package, we need a way to mock this config for our tests.

Have you considered mocking the virtual stripes-config with a static file?  The mock could be injected by defining an alias in the webpack config via `beforeBuild`.
```
return (config) => {
  config.resolve.alias['stripes-config'] = 'path/to/my/mock/stripes-config';
  // ...
}
```
This may simplify some setup, bypassing the need for mutations.  The stripes-config is effectively an interface between the server-side build logic and the running client-side platform. A static mock should allow us to bypass the virtual module generation for unit tests entirely. We may even be able to run the same tests against various configs, by pointing to different mocks.  Going a step further, we can and test this "contract" by validating that the build-time logic generates the stripes-config we expect (via new webpack tests).

wwilsman:
So the reason for not using static mocks is so different tests can test different aspects of supplying different configs. You mention pointing the same tests at different configs, but that seems more like a unit-style test where you want to validate the same results with different scenarios. We want to test how those scenarios effect the way in which users would experience the app. Users wouldn't be supplying different configs, they'd be interacting with the app handling those configs.

For example, the nav tests I've started specify an additional `module` on demand and check that clicking it goes to the specified route. Another test might deal with how the navbar handles displaying dozens of modules. This would be difficult to achieve with static mocks unless we create multiple bundles.

We can still write unit-style tests for a single piece handling multiple configs, but the main point is we want to test the app how a user would interact with it in different scenarios. Whether that's with different translations, different modules, different configs, etc.

wwilsman:
I'd also like to add, by avoiding multiple bundles and just using mocks in-browser, the user tests are much faster. I do agree that mutating the config is kinda gross, but I'd rather have a legitimate way to mock this data on-demand rather than relying on the bundler when dealing with user-facing concerns.

wwilsman:
Still issues with starting chrome 😕

ttamj:
> but the main point is we want to test the app how a user would interact with it in different scenarios.

@wwilsman Good explanation thanks.

> That way everything would be scoped to the current instance of StripesCore instead of being available "globally" through imports.

Agreed.  I would eventually like to see `stripes-config` imported in just one place.

test/.eslintrc
```diff
@@ -2,13 +2,9 @@
 // lexically bound "this" prevents access to the Mocha test context.
 // See https://mochajs.org/#arrow-functions
 {
-  "env": {
-    "mocha": true
```
ttamj: I can see how this would be misleading with BigTest's own `@bigtest/mocha`.  Is there a preference for using eslint `/* global */` comments over this setting where a standard Mocha environment does apply?

wwilsman: I prefer to be explicit with globals similar to treating them as imports. But since those webpack tests are started directly with mocha, I think it's fine to set the mocha env for those tests so I added a `.eslintrc` file to the `test/webpack` directory to re-enable it

wwilsman:
> Agreed. I would eventually like to see stripes-config imported in just one place.

@ttamj I was thinking about this last night, and the fact that I'm mutating this raised a concern for me that any module in the entire FOLIO ecosystem could do this. Even external dependencies if they first check if it's available within the bundle.

skomorokh:
> any module in the entire FOLIO ecosystem could do this (...) Even external dependencies

Yup, the entire NPM ecosystem we pull in. And they can all do anything they want in the runtime, the javascript we bundle isn't sandboxed per app. Stripes config? they can mutate the prototype for `Object`. There isn't too much we can do about that without massive development overhead is there?

That said, we should probably wrap at least the login [in an iframe](https://hackernoon.com/im-harvesting-credit-card-numbers-and-passwords-from-your-site-here-s-how-9a8cb347c5b5) with a smaller attack surface though. Maybe also other sensitive places like where we enter credentials for third party APIs.

wwilsman:
> I realize the existing tests may be disabled for debugging purposes, but I don't think we should merge until that's addressed in CI.

I agree. And yea, the scripts were adjusted so we could test if Jenkins would work with just running the core tests.

When there are multiple types of tests like this, we can usually tell Circle to just run two different jobs in parallel. One for the webpack tests and one for the core tests (and future parallel jobs for the various browsers). You can see the Circle config in this PR, and the link to the latest run is here: https://circleci.com/workflow-run/30f466db-be0b-4db6-81fd-9160075f6e07

I'm not suggesting a CI switch-up with this PR; I was just wondering if there is a way to run parallel jobs like this easily with Jenkins via some config? Or does it need custom CI scripts that @funkymalc would need to add to the Jenkins pipeline repo?

cherewaty:
Replaced by https://github.com/folio-org/stripes-core/pull/430

PR #429, opened 2018-09-26
STCOR-256: Main navigation "Show all applications"-dropdown and more

rasmuswoelk:
Currently, the main navigation can get a bit crowded. This PR changes that by hiding some of the applications on smaller screens and allowing users to show all installed applications in a dropdown instead.

image

## Issue
https://issues.folio.org/projects/STCOR/issues/STCOR-256

## Changes and additions
- Replaced the current nav buttons list with a new component `<AppList>`
- Created component `<AppList>` that handles the rendering of visible nav items as well as the "Show all applications"-dropdown
- Adjusted main navigation for mobile screens
- Removed title-attribute from nav buttons (replaced with aria-label)
- Improvement of accessibility
- Minor changes and some clean up in `<MainNav>`

... and more.

## Accessibility
Accessibility is key so I have put in a lot of work in order to meet the requirements for keyboard accessibility for the applications-dropdown. 

- Dropdown toggle button has `aria-expanded` (true/false) and  `aria-haspopup="true"`
- On dropdown open: focus currently selected/active app
- On dropdown close: switches focus back to dropdown toggle button
- When you reach the end of the list: Switch focus back to the first item. This traps the focus inside the list. (A user can escape/close the dropdown by hitting "ESC" on the keyboard. I am not sure if this approach is the correct so feedback would be appreciated.)
- Each item in the list has `role="menuitem"`. On the Mac screen reader this reads out that the user is inside a menu, on a menu item and that the menu (dropdown) can be closed by pressing "ESC" on the keyboard

I have tested these changes using the build-in screen reader for Mac. I would appreciate if some of you guys could test in your favorite screen reader.

## Preview
I have uploaded the changes to our test domain so that you can try it out.

http://folio-styling.testing.samhaeng.com

You can also check out the changes in this video demonstration:
https://www.dropbox.com/s/88v4yzrtfoa27a2/STCOR-256.mp4

## Going forward
These changes are only the beginning of the planned changes for the main navigation. 
In the future we will add a filter field for the applications dropdown as well as re-ordering of applications and more.

The newly created components are split into smaller child components corresponding to the planned functionality additions to hopefully ease the progress.


cherewaty:
Should the "Apps" label only show past a certain breakpoint? So small mobile view would look like this?

folio-styling testing samhaeng com_ iphone 5_se 1

JohnC-80:
In short, this will need some a11y iteration, but as far as cleaning up the space, it's looking good... and I think that tightening up the a11y experience can be handled iteratively as we've discussed...
 
JAWS, a slightly overbearing screen-reader, will announce the button as a menu, and say "Press space to activate the menu and navigate with arrow keys" - which is a common convention among dropdowns, select lists, etc.. I think we should reach out to the a11y SIG about this interaction model. 
Now that I can use the tab-navigation version, I'm leaning towards focusing the next top-level item rather than going back to the beginning of the app list in the dropdown - this is because as long as the dropdown is open, the user is stuck there for tabbing, despite other ways of navigating around the DOM...  
The dropdown itself will tell you that it's a menu, but for some reason it doesn't announce the expanded/collapsed status.

I tested on JAWS 2018 in Chrome

rasmuswoelk:
The test site is up again. The previous backend endpoint was offline.

@cherewaty I initially did make it like that (code is just commented out - no caret icon though) but Filip requested that we preserve the "Apps"-label as long as we don't have anything else in the Main Nav (breadcrumbs etc.) on mobile. I have no problem with removing the label on mobile straight away - breadcrumbs or not.

@JohnC-80 Thanks for testing on JAWS!

I did some testing with the Mac screen reader — and while I'm not an expert in how the ideal screen reader interaction should be — I could actually follow the instructions and use the proposed keyboard shortcuts. Check out the GIF below.

I assume that as long as the commands actually work then we are good? But of course if other screen readers will read out and/or assume some other keyboard shortcuts to work, then we are in trouble.

sep-28-2018 13-40-46

E.g. when a menu item is focused it will tell you that you can close the menu by using "ESC". This gives the user a way out of the focus trap. And by pressing the other commands I am able to browse through the list and select items.

I also noticed that it doesn't read out the collapsed status. Not sure how to achieve that.

As for focus trapping: I can change it to continue to the next menu item in the Main Nav - I guess that it should then close the menu?

Another option is to add a fleuron-icon in the bottom of the list that has an aria-label of e.g. "You have reached the end of the list. Tab again to go to the top or press ESC to close the menu.". I know that Filip wants to add fleuron's to the end of the various lists in the Folio-system to indicate that the user has reached the end of a list.

Let me know what you think and I can make the appropriate changes 👍 

And for accessibility changes in general: Can we release the changes to the testing sites and then get some more feedback from the accessibility SIG's?

rasmuswoelk:
@cherewaty Cool, thanks!

Actually, I have just removed the label from mobile screen sizes. I think it's best since it could quickly become an issue on very small devices - even without breadcrumbs.

rasmuswoelk:
There are some test errors. I'll look into this.

rasmuswoelk:
Merging. We'll adjust the accessibility aspect after getting proper feedback 👍

PR #430, opened 2018-09-26
[STCOR-251] Add BigTest Infrastructure

cherewaty:
Replaces https://github.com/folio-org/stripes-core/pull/426; Jenkins has a security measure that prevents `Jenkinsfile` changes from being picked up when a PR is opened from a fork.


cherewaty:
@funkymalc is working on an idea for running additional scripts like `yarn test:webpack` in the Jenkins build.

ttamj:
I do see similar errors getting logged on folio-testing right now, so perhaps they are not related.

PR #431, opened 2018-09-26
Turn on react-hot-loader

cherewaty:
With the setup changes from https://github.com/folio-org/stripes-core/pull/430, we can now use [`react-hot-loader`](https://github.com/gaearon/react-hot-loader) when running `yarn start` with `stripes-core`:

A three line-change in each UI module will activate their ability to take advantage of hot reloading too (I'll open some PRs to demonstrate after this merge).

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1,7 +1,4 @@
-import 'babel-polyfill';
-import React from 'react';
-import { render } from 'react-dom';
+import { hot } from 'react-hot-loader';
+import AppContainer from './AppContainer';
 
-import StripesCore from './App';
-
-render(<StripesCore />, document.getElementById('root'));
+export default hot(module)(AppContainer);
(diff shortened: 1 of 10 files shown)


src/index.js
```diff
@@ -1,7 +1,4 @@
-import 'babel-polyfill';
-import React from 'react';
-import { render } from 'react-dom';
+import { hot } from 'react-hot-loader';
+import AppContainer from './AppContainer';
 
-import StripesCore from './App';
-
-render(<StripesCore />, document.getElementById('root'));
+export default hot(module)(AppContainer);
```
wwilsman: I'm surprised that this works. `AppContainer` doesn't export anything for this to hook into.

I think it would be a little more explicit if we had a `HotApp` module we import into the index and render instead of the `AppContainer` file. Especially because the naming `AppContainer` suggests it is a component, but that file actually renders the app.

cherewaty: Yeah, that change makes sense. Commit added to move the render back to `index.js`, create a `HotApp`, and remove `AppContainer`.

PR #434, opened 2018-09-27
STCOR-262

funkymalc:
Add a 'runScripts' option to the shared NPM build pipeline which can be configured in the Jenkinsfile as a map consisting of NPM script name and script arguments.


cherewaty:
Thanks @funkymalc 🙌  looks good

PR #437, opened 2018-09-28
STCOR-257 Remove notifications dropdown

cherewaty:
https://issues.folio.org/browse/STCOR-257

localhost_3000_ iphone 5_se


cherewaty:
We've got code coverage reporting on for `stripes-core` now, and the line coverage is right around 50%. Keeping around code that's unused will impede our ability to get that number up to a much healthier percentage.

The git history will remain, for when we return to this future.

ttamj:
@zburke Are there open JIRA's related to the Notification work that we could point to here?

cherewaty:
Rebased to fix conflicts with https://github.com/folio-org/stripes-core/pull/429

MikeTaylor:
> We've got code coverage reporting on for stripes-core now, and the line coverage is right around 50%. Keeping around code that's unused will impede our ability to get that number up to a much healthier percentage.

This line of reasoning is ${think of a more polite synonym for "ridiculous"}.

Code coverage metrics are our servant, not our master. Deleting perfectly good code in order to make an arbitrary number go up is not helping us to get FOLIO written.

PR #438, opened 2018-09-28
Update test helpers

wwilsman:
## Purpose

While adjusting `ui-checkin` to use the new `setupApplication` helper from core, I came across an issue with the `clearModules` helper removing the pre-exististing module for `ui-checkin` added by the webpack plugin.

I also discovered that the `withTranslations` helper was faulty, in that you would have to include each modules translations as they are needed. This is not ideal, especially when a module requires a package such as `stripes-form`, in this case the module would need to also provide those translations itself.

Mirage only supports a default scenario out-of-the-box. For this reason, we manually load scenarios after mirage has initialized. However, we were including the default scenario in tests, when there should be no scenarios loaded automatically for tests.

## Approach

Save a reference to the original `stripes-config` options that we modify, and once clearing out those config objects, restore them to the original options provided by the webpack plugin.

Instead of mocking the translations endpoint, stub the `setTranslations` action creator to inject additional translations. When we mocked the endpoint, we needed to manually supply _all_ translations. This led to issues noted above. There was no way to intercept the request using mirage or pretender, so I dug one level down and stubbed the action creator. Since action creators do not contain any logic, I think this is a suitable mechanism for injecting additional translations as it's only one layer away from the original network mocks.

Finally, I had a problem in `ui-checkin` with duplicate mirage records only to discover the duplication was coming from the default scenario. Mirage internally only applies the default scenario if no other scenarios are provided and if _not_ in a testing environment. Since we do not pass along scenarios to mirage's initialization options, we need to manually apply this logic ourselves. 

Mirage also conditionally loads factories based on the presence of the default scenario when not in test mode. I can't foresee a situation in which we would *not* want factories, so now we always load factories unconditionally.


wwilsman:
@cherewaty Yup! That `--mirage` option is handled by `network/boot.js`. In the other repos, we just need to adjust that file to include local mirage options. [I did this here for ui-checkin](https://github.com/folio-org/ui-checkin/pull/71/files#diff-40d485ce78b8c91a4b1af6e5ae45ad53) (note also: since we use `startMirage` from core, the import in `boot.js` needs to change to reflect that too).

PR #441, opened 2018-10-01
STCOR-255 Make current app clickable

cherewaty:
Resolves https://issues.folio.org/browse/STCOR-255

### DIFF
--- a/src/components/MainNav/CurrentApp/CurrentApp.js
+++ b/src/components/MainNav/CurrentApp/CurrentApp.js
@@ -4,6 +4,7 @@
 
 import React from 'react';
 import PropTypes from 'prop-types';
+import Link from 'react-router-dom/Link';
 import AppIcon from '@folio/stripes-components/lib/AppIcon';
 import Badge from '@folio/stripes-components/lib/Badge';
 import Headline from '@folio/stripes-components/lib/Headline';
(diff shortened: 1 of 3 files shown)


src/components/MainNav/CurrentApp/CurrentApp.js
```diff
@@ -31,13 +32,25 @@ const defaultProps = {
 const CurrentApp = ({ currentApp, iconData, id, badge }) => {
   const { displayName, description, module } = currentApp;
   const iconKey = module && module.replace(/^@[a-z0-9_]+\//, '');
+  let Element = 'div';
+  const elementProps = {};
```
wwilsman: I hate that this is a `const` that we mutate. Probably just eslint being overzealous about `const` use?

cherewaty: I updated to this to reduce the nails-on-the-`const`board effect:
```
let elementProps = {};

if (currentApp && (currentApp.home || currentApp.route)) {
  Element = Link;
  elementProps = { to: (currentApp.home || currentApp.route) };
}
```

cherewaty: And yes, it was ESLint being silly.

PR #443, opened 2018-10-01
Remove unnecessary typeof check

cherewaty:
SonarCloud scan picked this up as a "bug".

The ternary was returning the exact same thing in both conditions.

### DIFF
--- a/src/components/Notification/ToastNotification.js
+++ b/src/components/Notification/ToastNotification.js
@@ -106,18 +106,7 @@ class ToastNotification extends React.Component {
   }
 
   render() {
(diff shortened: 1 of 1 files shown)


zburke:
Clearly, SonarCube is right about this. But looking at the very first edition of the code, before we threw the lint formatter at it, I'm will guess the intent of the ternary was to decide whether to display `this.props.message` before the button. Dunno for sure, of course. @JohnC-80, do you recall?

cherewaty:
@JohnC-80 I'll leave this for you to merge / decide what to do with.

PR #444, opened 2018-10-02
Replaced content of <CurrentApp> with a <NavButton>

rasmuswoelk:
The design and functionality of the `<CurrentApp>` are currently basically the same as a `<NavButton>`. To avoid the same logic (and styles) being written twice, I have replaced the contents of CurrentApp with a NavButton.

## Changes
- Added NavButton to CurrentApp (removes visited link color etc.)
- Added ariaLabel prop to NavButton
- Replaced title-attribute with aria-label
- Made sure the app icon is working when on "Settings"

## Note
I have added an aria-label for active applications that reads out:

`Current open application: {appName} (Click to go home)`.

I figured that this would be useful information for the user. I would like some feedback on this approach.

## Before
image

## After
oct-02-2018 12-58-26


translations/stripes-core/en.json
```diff
@@ -66,5 +66,6 @@
   "loginViaSSO": "Login via SSO",
 
   "mainnav.showAllApplicationsButtonLabel": "Apps",
-  "mainnav.showAllApplicationsButtonAriaLabel": "Show all applications"
+  "mainnav.showAllApplicationsButtonAriaLabel": "Show all applications",
+  "mainnav.currentAppAriaLabel": "Current open application: {appName} (Click to go home)"
```
cherewaty: I don't know that we need this extra text, but we can proceed with it.

PR #446, opened 2018-10-03
Fix currentUser error in tests

wwilsman:
## Purpose

https://github.com/folio-org/ui-myprofile/pull/31 encountered an error where a component was attempting to access `stripes.user.user.username` when `stripes.user.user` was undefined.

This stripes property is populated via  the `Login` component either automatically or when manually logging in. When using `disableAuth` the `\login` route and `Login` component is never mounted, so the store's `currentUser` always ends up empty.

## Approach

Instead of using `disableAuth`, supply the initial state of the store with fake user data, including a fake token, and `perms.all`.

#### Questions

I assume `perms.all` should be the only permission necessary, but looking at dev and production admin accounts shows a lot more perms. Do I need to include any other permissions?

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -15,7 +15,7 @@ import Root from './components/Root';
 export default class StripesCore extends Component {
   static propTypes = {
     history: PropTypes.object,
-    initiailState: PropTypes.object
+    initialState: PropTypes.object
   };
 
   constructor(props) {
--- a/src/App.js
+++ b/src/App.js
@@ -1,6 +1,7 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import { okapi as okapiConfig, config } from 'stripes-config';
+import merge from 'lodash/merge';
 
 import connectErrorEpic from './connectErrorEpic';
 import configureEpics from './configureEpics';
@@ -13,7 +14,8 @@ import Root from './components/Root';
 
 export default class StripesCore extends Component {
   static propTypes = {
-    history: PropTypes.object
+    history: PropTypes.object,
+    initiailState: PropTypes.object
   };
 
   constructor(props) {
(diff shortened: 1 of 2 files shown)


src/App.js
```diff
@@ -13,7 +14,8 @@ import Root from './components/Root';
 
 export default class StripesCore extends Component {
   static propTypes = {
-    history: PropTypes.object
+    history: PropTypes.object,
+    initiailState: PropTypes.object
```
cherewaty: `initialState`

wwilsman: Fixed

wwilsman:
@ttamj I added an option for the `setupApplication` helper to allow passing in more perms as the module needs them. If we want to include more perms here by default we can drop them in now or as they come up. But there is now at least a way for other modules to specify more permissions.

This also allows modules to test when a users **does not** have specific permissions.

ttamj:
Okay, so these two are _not_ related.

My concern with `hasAllPerms` is that it would effectively bypass UI permission checks altogether.  Doing so would prevent the scenario you describe where you might test that a user does not have a permission.
https://github.com/folio-org/stripes-core/blob/751681309071b7cabd6270ebbb67717630bea8bf/src/Stripes.js#L93-L96

Since `hasAllPerms` is not related, this is not going to be an issue.

However, looking into it further, `perms.all` is the permission for the mod-permissions module, which gives a user, not "all permissions", but the ability to perform CRUD operations on permissions and assign permissions to users.
https://github.com/folio-org/mod-permissions/blob/56423022a78598e305e081c2391bc412be8c9882/descriptors/ModuleDescriptor-template.json#L150-L156

This single permission seems a bit arbitrary to provide for the unit tests.  Are back-end permissions even necessary for running front-end unit tests, given the back-end is mocked?

wwilsman:
Are permissions gathered at any other point besides when logging in? The changes here still bypass the login since a fake `token` is provided. If permissions are gathered at another point, then the back-end mocks should suffice. If permissions are only gathered upon login, then I'll remove the `perms.all` and just leave in the ability to specify custom permissions.

zburke:
I am relatively confident that permissions are only gathered upon login.

wwilsman:
Thanks @zburke. We decided to leave in the `setupApplication({ permissions })` option so that modules that make use of this helper can declare specific permissions on a per-test basis.

PR #447, opened 2018-10-03
Remove HMR router hack

cherewaty:
`key={Math.random()}` was preventing this from appearing in the console: `Warning: You cannot change <Router history>`; but that warning does not appear to actually affect any functionality.


zburke:
per @mkuklis: The random key was reinitializing children components including `MainNav` which also causes issues with the navigation. Refs [STRIPES-562](https://issues.folio.org/browse/STRIPES-562).

PR #449, opened 2018-10-03
Export BigTest helpers

cherewaty:
### Purpose
With `@folio/stripes`, modules won't be able to import un-exported classes/functions from `@folio/stripes/core`.

### Questions
- Are these the names we want to export?
- Anything else we might need to export re: BigTesting?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Change history for stripes-core
 
+## 2.16.0 (IN PROGRESS)
+
+* Export BigTest helpers
+
 ## [2.15.1](https://github.com/folio-org/stripes-core/tree/v2.15.1) (2018-10-03)
 [Full Changelog](https://github.com/folio-org/stripes-core/compare/v2.15.0...v2.15.1)
 
--- a/index.js
+++ b/index.js
@@ -7,3 +7,6 @@ export { setServicePoints, setCurServicePoint } from './src/loginServices';
 export { default as TitleManager } from './src/components/TitleManager';
 export { default as HandlerManager } from './src/components/HandlerManager';
 export { default as coreEvents } from './src/events';
+
+export { default as setupStripesCore } from './test/bigtest/helpers/setup-application';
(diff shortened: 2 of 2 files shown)


wwilsman:
In `stripes-smart-components`, I'm also utilizing `withModules` and `clearModules` from the `stripes-config` BigTest helpers

cherewaty:
hmmmm I think `stripes-smart-components` will still be able to:
```
import { withModules, clearModules } from '@folio/stripes-core/test/bigtest/helpers/stripes-config';
```

... but modules wouldn't be able to. Let's double check that with @ttamj

wwilsman:
That works for me. I wasn't quite sure so I thought I'd mention it. I don't think modules will ever need those helpers

ttamj:
I wonder if, as with the interactor discussion earlier, we consider exporting a separate interface for the test infrastructure. It feels a little cleaner and doing so would help guard against any failed tree-shaking attempts that may leave testing bits in the bundle.

For example, we might establish these exports under `/test/index.js` and similarly re-export via `stripes`.  Then a consumer of `stripes` (v1.1) might pull it in like so:

```
import { startMirage } from `@folio/stripes/core/test`;  // ... ?
```

Sorry I hadn't factored these sorts of exports into the POC.  Given we're in the midst of rolling out 1.0, I think we should see that through.  Then follow-up with a 1.1 (however quick it may be) after proving out test exports/shaking.

In the mean time, any stripes 1.0 app that needs `startMirage`, etc., can continue to import as it does today, and (temporarily) maintain a devDependency on stripes-core.

cherewaty:
Based on Oct. 4 stripes architecture call, we'll try the separate `@folio/stripes/core/test` interface in `@folio/stripes` `1.1`; this PR can proceed as-is for now

ttamj:
@cherewaty, I understood, we'd skip the export and rely on a devDependency on stripes-core in the few apps for the time being...

cherewaty:
:headslap: yes, you're right, Matt. It's been a long day already 😩

PR #460, opened 2018-10-09
Suppress translation key from UI by adding English default translation

julianladisch:


doytch:
I've heard as much @zburke. An alternative would be to [use the `defaultMessage` fallback](https://github.com/yahoo/react-intl/wiki/Components#message-formatting-fallbacks) for cases like this where showing an ID is particularly annoying.

zburke:
It's my understanding that we have Lokalise configured to overwrite missing and empty values in the non-English files with the values from `en.json`. Additionally, Lokalise is configured to be the authoritative source for changes to everything except `en.json` and will overwrite changes made directly to the repositories. 

The only reason a translation key would directly appear in the app, then, is if the value were missing in the `en.json` file.

doytch:
Yeah I think you're right @zburke. I believe @julianladisch's intent is to fix the current annoyance of developing on Folio while using a non-English locale. I don't think he cares that it'll eventually be overwritten with the _actual_ translated values.

dltj:
All of the above is correct.  Good point about developing in a non-English locale.  This stuff gets trickier and trickier.

dltj:
...and if we do want to use the defaultMessage fallback, we should make a new ticket for that and relate it to https://issues.folio.org/browse/STCOR-190

PR #464, opened 2018-10-10
UIU-590: Change error handling. Add tests.

OleksandrAntonenko1:


src/loginServices.js
```diff
@@ -110,15 +115,50 @@ export function getBindings(okapiUrl, store, tenant) {
...
+  ));
+}
+
+function parseErrors(json) {
+  const { errorMessage } = json;
+  try {
+    const { errors } = JSON.parse(errorMessage);
+
+    return errors.length > 1 ? (
+      <ul style={styles.multipleErrorsUl}>{getPasswordValidationMessages(errors)}</ul>
```
cherewaty: Could we move the style information to a `.css` file and use `className=""` instead of inline styles?

test/bigtest/network/config.js
```diff
@@ -20,23 +30,110 @@ export default function configure() {
...
-        }
-      },
-      permissions: {
-        permissions: []
-      }
-    });
+  this.post('/bl-users/login', (schema, request) => {
+    const formData = JSON.parse(request.requestBody);
+
+    switch (formData.username) {
```
cherewaty: Mirage route definitions can be overridden in individual tests. I'd recommend redefining the `bl-users/login` endpoint in each individual test that needs a different response, instead of the `default` scenario relying on different inputs.

Here's an example: https://github.com/folio-org/ui-eholdings/blob/15dbfb7a1a73b720834fc12cc3bf8105bd30e406/test/bigtest/tests/resource-show-test.js#L190-L194

src/loginServices.js
```diff
@@ -110,15 +115,50 @@ export function getBindings(okapiUrl, store, tenant) {
...
+function getPasswordValidationMessages(errors) {
+  return errors.map(element => (
+    <li key={`${element.code}-${element.type}`}>
+      { getErrorMessage(element)}
+    </li>
+  ));
+}
+
+function parseErrors(json) {
+  const { errorMessage } = json;
```
VictorSoroka1: Please add space after defining variables.

test/bigtest/tests/login-test.js
```diff
@@ -19,13 +38,292 @@ describe('Login', () => {
...
+
+    it('error message should not be present', () => {
+      expect(login.message.isPresent).to.be.false;
+    });
+  });
+
+  describe('errors', () => {
+    describe('error for the wrong username', () => {
+      beforeEach(async () => {
+        const username = login.username;
```
VictorSoroka1: Please use destruction for such cases. Check other places.

```
    const { password, submit, username } = login;
```

test/bigtest/tests/login-test.js
```diff
@@ -19,13 +38,292 @@ describe('Login', () => {
...
+
+      it('error message should be present upon failed submit', () => {
+        expect(login.message.isPresent).to.be.true;
+      });
+
+      it('error message should have proper text upon failed submit', () => {
+        expect(login.message.text).to.equal(translations['loginServices.login.fifth.failed.attempt.blocked']);
+      });
+    });
+
```
VictorSoroka1: Remove excess empty line.

test/bigtest/tests/login-test.js
```diff
@@ -19,13 +38,292 @@ describe('Login', () => {
...
+      });
+
+      it('error message should be present upon failed submit', () => {
+        expect(login.message.isPresent).to.be.true;
+      });
+
+      it('error message should have proper text upon failed submit', () => {
+        expect(login.message.text).to.equal(parseMessageFromJsx(
+          translations['loginServices.login.username.invalid'],
+          { username:wrongUsername }
```
VictorSoroka1: Add space after the colon.

src/loginServices.js
```diff
@@ -110,15 +115,50 @@ export function getBindings(okapiUrl, store, tenant) {
     });
 }
 
+function getErrorMessage(error = {}) {
+  const { code = 'default.error', parameters } = error;
+  const values = parameters ?
+    parameters.reduce(
+      (res, key) => {
+        res[key.key] = key.value;
+        return res;
```
VictorSoroka1: Please add an empty line before`return` statement or even refactor it to the code below:
```
parameters.reduce((res, { key, value }) => { ...res, [key]: value }, {})
```

Try to avoid such name constructions like `key.key` whenever possible.

src/loginServices.js
```diff
@@ -110,15 +115,50 @@ export function getBindings(okapiUrl, store, tenant) {
     });
 }
 
+function getErrorMessage(error = {}) {
+  const { code = 'default.error', parameters } = error;
```
VictorSoroka1: What is the value of the parameters when they are empty? I am wondering whether it can also be assigned to the default value like:

```
parameters = {}
```

src/loginServices.js
```diff
@@ -2,8 +2,9 @@ import _ from 'lodash';
 import 'isomorphic-fetch';
 import localforage from 'localforage';
 import { change } from 'redux-form';
-import { addLocaleData } from 'react-intl';
+import { addLocaleData, FormattedMessage } from 'react-intl';
 import { translations } from 'stripes-config';
+import React from 'react';
```
VictorSoroka1: React import usually placed on the very top. Please move it there for consistency.

src/loginServices.js
```diff
@@ -23,6 +24,10 @@ import {
   setUserServicePoints,
 } from './okapiActions';
 
+const styles = {
+  multipleErrorsUl: { textAlign: 'left', margin: '0' }
```
maximdidenkoepam: Are you sure that we need to store styles in service? Maybe it should be moved to the error component?

src/loginServices.js
```diff
@@ -110,15 +115,50 @@ export function getBindings(okapiUrl, store, tenant) {
     });
 }
 
+function getErrorMessage(error = {}) {
+  const { code = 'default.error', parameters } = error;
+  const values = parameters ?
+    parameters.reduce(
+      (res, key) => {
+        res[key.key] = key.value;
```
maximdidenkoepam: res[key.key] looks pretty strange. Is key an object? If yes maybe it should be renamed and you also can use destructuring? Btw, use short notation. See comment below.

test/bigtest/tests/login-test.js
OleksandrAntonenko1: Done.

src/components/Login/Login.js
```diff
@@ -19,7 +19,7 @@ class Login extends Component {
     handleSubmit: PropTypes.func.isRequired,
     onSubmit: PropTypes.func.isRequired,
     submitting: PropTypes.bool,
-    authError: PropTypes.string,
+    authError: PropTypes.arrayOf(PropTypes.object),
```
maximdidenkoepam: Please change property name to authErrors, because it is an array

src/components/Login/components/authError/AuthError.js
```diff
@@ -0,0 +1,50 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import { FormattedMessage } from 'react-intl';
+import classNames from 'classnames';
+
+import styles from './AuthError.css';
+
+export default class AuthError extends Component {
```
maximdidenkoepam: As far as I understand this component is an error container. Maybe we should rename it? AuthErrorsContainer or ErrorsPlaceholder?

src/components/Login/components/authError/AuthError.js
```diff
@@ -0,0 +1,50 @@
...
+  static propTypes = {
+    errors: PropTypes.arrayOf(PropTypes.object),
```
maximdidenkoepam: Please add isRequired to errors property or defaultProps if this property is not required

src/components/Login/components/authError/AuthError.js
```diff
@@ -0,0 +1,50 @@
...
+
+  getErrorMessage(error = {}) {
+    const {
+      code = 'default.error',
+      parameters = [],
+    } = error;
+
+    const values = parameters.reduce((res, { key, value }) => ({ ...res, [key]: value }), {});
+
+    return <FormattedMessage id={`stripes-core.loginServices.login.${code}`} values={values} />;
```
maximdidenkoepam: Please put each component property to a separate row and wrap this component in a brackets

src/components/Login/components/authError/AuthError.js
```diff
@@ -0,0 +1,50 @@
...
+  parseErrors(errors) {
+    return errors.length > 1 ? (
+      <ul className={styles.errorsList}>{this.getPasswordValidationMessages(errors)}</ul>
+    ) : (
+      this.getErrorMessage(errors[0])
+    );
+  }
+
+  render() {
+    const { errors } = this.props;
```
maximdidenkoepam: Why do we need errors in render? All you do is just passing them to parseErrors method. Let's move it to parseErrors.

src/components/Login/Login.js
```diff
@@ -101,7 +101,7 @@ class Login extends Component {
               </Button>
             </div>
             <div className={authFormStyles.formGroup}>
-              { authError ? <div className={classNames(authFormStyles.formMessage, authFormStyles.error)}>{authError}</div> : null }
+              { authErrors ? <AuthErrorsContainer errors={authErrors} /> : null }
```
maximdidenkoepam: You can use the same construction as in line 107.

src/components/Login/components/AuthErrorsContainer/AuthErrorsContainer.js
```diff
@@ -0,0 +1,53 @@
...
+    return (
+      <FormattedMessage
+        id={`stripes-core.loginServices.login.${code}`}
+        values={values}
+      />
+    );
+  }
+
+  getPasswordValidationMessages(errors) {
+    return errors.map(element => (
```
maximdidenkoepam: Please rename element to error

src/components/Login/components/AuthErrorsContainer/AuthErrorsContainer.js
```diff
@@ -0,0 +1,53 @@
...
+
+import styles from './AuthErrorsContainer.css';
+
+export default class AuthErrorsContainer extends Component {
+  static propTypes = {
+    errors: PropTypes.arrayOf(PropTypes.object).isRequired,
+  };
+
+  getErrorMessage(error = {}) {
+    const {
```
maximdidenkoepam: If in Login.js you will set default value to an empty array for auth errors you will avoid default value for this method because it will newer invoked without error object

src/components/Login/components/AuthErrorsContainer/AuthErrorsContainer.js
```diff
@@ -0,0 +1,53 @@
...
+        {this.getErrorMessage(element)}
+      </li>
+    ));
+  }
+
+  parseErrors() {
+    const { errors } = this.props;
+
+    return errors.length > 1 ? (
+      <ul className={styles.AuthErrorsList}>{this.getPasswordValidationMessages(errors)}</ul>
```
maximdidenkoepam: Please move each property and component body to a separate line

src/components/Login/components/AuthErrorsContainer/AuthErrorsContainer.js
```diff
@@ -0,0 +1,53 @@
...
+    ) : (
+      this.getErrorMessage(errors[0])
```
maximdidenkoepam: Are these styles different for list of errors and single error? Why we can't use <ul> for both cases to avoid condition?

src/components/Login/Login.js
```diff
@@ -48,7 +48,7 @@ class Login extends Component {
     const {
       handleSubmit,
       submitting,
-      authError,
+      authErrors = false,
```
maximdidenkoepam: Are you sure that default value for an array should be false? Please use static field defaultProps.
`static defaultProps = {
   authErrors: [],
}`

src/components/Login/components/AuthErrorsContainer/AuthErrorsContainer.js
```diff
@@ -0,0 +1,67 @@
...
+        className={className}
+      >
+        {messages}
+      </ul>
+    );
+  }
+
+  render() {
+    return (
+      <div
```
maximdidenkoepam: In case when element has only one property please put it in one line `<div className="class">`
Move to new line only when you have more than one property. Fix all the issues in this file.

src/components/Login/Login.js
```diff
@@ -1,13 +1,14 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
-import classNames from 'classnames';
 import { connect } from 'react-redux';
 import { reduxForm, Field, Form, formValueSelector } from 'redux-form';
+import _ from 'lodash';
```
wwilsman: Instead of importing _all_ of lodash. Let's just import what we need.
```suggestion
import isEmpty from 'lodash/isEmpty';
```

test/bigtest/tests/login-test.js
```diff
@@ -19,13 +27,355 @@ describe('Login', () => {
     expect(login.submit.isDisabled).to.be.true;
   });
 
-  describe('with invalid credentials', () => {
-    it.always('does not log in');
-    it('displays an error');
+  it('error message should not be present', () => {
```
wwilsman: When we test that something _should not_ happen, there can be times where the test passes before the thing happens. We can use `it.always` here to ensure that this test passes for a short period of time (`200ms` by default).

You can learn more about these type of tests here: [@bigtest/mocha](https://github.com/bigtestjs/mocha/blob/master/README.md)

test/bigtest/tests/login-test.js
```diff
@@ -19,13 +27,355 @@ describe('Login', () => {
...
+
+      await username.fill('test');
+    });
+
+    it('has an active submit button', () => {
+      expect(login.submit.isPresent).to.be.true;
+      expect(login.submit.isDisabled).to.be.false;
+    });
+
+    it('error message should not be present', () => {
```
wwilsman:
```suggestion
    it.always('error message should not be present', () => {
```

wwilsman: In general, the same feedback applies everywhere we test that something `should not` happen. So I won't leave these comments all over the place.

PR #469, opened 2018-10-16
Fix maximum update depth exceeded. Fixes STRIPES-562. 

mkuklis:
This PR is an attempt to fix the "maximum update depth exceeded" when moving from the users settings module back to users module. More details can be found under [STRIPES-562](https://issues.folio.org/browse/STRIPES-562).

It looks like the `params` object was not constructed correctly in some cases for example when:

`locationQuery = { query: 'a', sort: 'Name' }`  
`stateQuery = { query '', sort: 'Name' }` 

the result of the previous implementation would return:

`{ sort: 'Name' }` 

instead of: 

`{ query: 'a', sort: 'Name' }` 

which was causing the infinitive loop.

The entire `locationService` implementation including `queryResource` looks like a hack. Hopefully we can transition to a better implementation at some point but for now I'm hoping this PR should fix STRIPES-562.

@aditya-matukumalli and @zburke would you mind giving it a try? The best way to reproduce it:

1. Navigate to users app and conduct a search, for example "a".
2. Navigate to settings > users
3. Navigate between two or three of the general categories (ex. patron groups > permission sets > profile pictures)
4. Navigate back to users app

Thanks!

### DIFF
--- a/src/locationService.js
+++ b/src/locationService.js
@@ -1,4 +1,4 @@
-import { snakeCase, isEqual, forOwn, isEmpty, unset } from 'lodash';
+import { snakeCase, isEqual, forOwn, isEmpty, unset, mergeWith } from 'lodash';
 import queryString from 'query-string';
 import { replaceQueryResource } from './locationActions';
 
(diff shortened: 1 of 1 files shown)


mkuklis:
@zburke thank you for testing this! You are right. My troubleshooting introduced a new bug...

zburke:
I _think_ the problem is that sometimes you want the location query to overwrite the state, e.g. when you click a cancel button, you want to transition from `...?layer=edit` to an empty query string. Using `merge` instead of `removeEmpty` that never happens. You're right that
```
locationQuery = { query: 'a', sort: 'Name' }
stateQuery = { query '', sort: 'Name' }
```
should reduce to 
```
{ query: 'a', sort: 'Name' }
```
but it's also true that 
```
locationQuery = { query: '', sort: 'Barcode' }
stateQuery = { query 'b', sort: 'Name' }
```
should reduce to 
```
{ query: '', sort: 'Barcode' }
```

mkuklis:
It looks like we can't handle it this way because in some cases like the one you mentioned @zburke we want to actually clear up / override some params located in the url with what we have in state. I'm going to close this PR and start looking at it again.

mkuklis:
@zburke I missed your last comment :) Yes exactly what you are describing. We need to find a different way to fix the original issue.

PR #470, opened 2018-10-16
STCOR-151 update quick start guide

ttamj:
This brings the Quick Start guide up to date with basic CLI operations for both app and platform development.

The New Development Setup Guide will need updating next.  That document will include more detail and alternative manual steps to accomplish the same tasks.


dcrossleyau:
Your changes make it so much better. Thanks.

A minor point: The generated stripes.config.js.local has "users" listed twice in the "modules" section.

ttamj:
Thank you for the edits @dcrossleyau.  I've created STCLI-110 to address the duplicate "users."

PR #471, opened 2018-10-16
Fix maximum update depth exceeded. Fixes STRIPES-562

mkuklis:
@zburke this is a second attempt to fix this :). This one is a bit cleaner. I think we originally were defaulting to empty `query` when navigating from the codex search to a specific inventory item. It seems like everything is still working fine without the query present right now. Wonder if you could give  it a try again.


zburke:
This replaces #469.

mkuklis:
It looks like the work in https://issues.folio.org/browse/STCOM-302 also fixed this issue.

PR #473, opened 2018-10-24
[STCOR-269] Fix dropdown behavior

anastasiaHahanova:
## Purpose
Regarding [STCOR-269](https://issues.folio.org/browse/STCOR-269) the dropdown with possible applications should not have a scroll till its height equals screen's height, the scroll should appear only if the dropdown's height more than screen's.

## Approach
- Remove set max-height, redefine paddings for the dropdown.
- Add constraints to tether props for fixing position of the dropdown.

### DIFF
--- a/src/components/MainNav/AppList/AppList.css
+++ b/src/components/MainNav/AppList/AppList.css
@@ -4,6 +4,10 @@
 
 @import '@folio/stripes-components/lib/variables.css';
 
+:root {
+  --padding-for-header: 66px;
+}
+
 .appList {
   display: flex;
 }
@@ -39,16 +43,19 @@
   z-index: 200;
 }
 
+.navListDropdown > div {
+  padding: 10px 5px;
+}
+
 .dropdownBody {
   width: 100%;
   outline: 0;
 }
 
-ul.dropdownList {
-  padding: 10px 5px;
+.dropdownList {
   margin: 0;
-  max-height: 225px;
   overflow-y: auto;
+  max-height: calc(100vh - var(--padding-for-header));
 }
 
 .dropdownListItem {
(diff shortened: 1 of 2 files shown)


src/components/MainNav/AppList/AppList.js
```diff
@@ -186,20 +190,20 @@ class AppList extends Component {
     return (
       <nav className={css.appList}>
         <ul className={css.navItemsList}>
-          { renderNavButtons() }
+          {renderNavButtons()}
         </ul>
         <div className={css.navListDropdownWrap}>
           <Dropdown tether={tether} dropdownClass={css.navListDropdown} open={this.state.open} id={dropdownId} onToggle={toggleDropdown}>
```
Godlevskyi: please stack such long lists of properties into a column

src/components/MainNav/AppList/AppList.css
```diff
@@ -39,16 +43,20 @@
...
+.navListDropdown > div {
+  padding: 10px 0;
+}
+
 .dropdownBody {
   width: 100%;
   outline: 0;
 }
 
 ul.dropdownList {
```
Godlevskyi: What is the reason to add extra selector ul?

src/components/MainNav/AppList/AppList.js
```diff
@@ -186,20 +190,20 @@ class AppList extends Component {
...
-              { this.focusTrap(this.focusDropdownToggleButton) }
+              {this.focusTrap(this.focusDropdownToggleButton)}
               <AppListDropdown
                 listRef={this.dropdownListRef}
                 apps={apps}
                 dropdownToggleId={dropdownToggleId}
                 toggleDropdown={toggleDropdown}
               />
-              { this.focusTrap(this.focusFirstItemInList) }
+              {this.focusTrap(this.focusFirstItemInList)}
```
MaksymDryha: Isn't it better to destructure this stuff? Especially considering that this is already done with other properties of 'this' object at the top of render method.

PR #476, opened 2018-10-26
STRIPES-573 move stripes docs

ttamj:
Many `stripes-core` documents have moved to the `stripes` repository.  The content of these documents has been replaced with a notice linking to their new location.  This is done to ensure the documentation remains accessible for anyone following external links into stripes-core.

STRIPES-573


dcrossleyau:
I merged both. Thanks Matt, good stuff.

PR #480, opened 2018-10-30
STCOR-273:  Add Create/Reset a Password Screen.

OleksandrAntonenko1:


src/components/CreateResetPassword/CreateResetPassword.js
```diff
@@ -0,0 +1,233 @@
...
+import FieldLabel from './components/FieldLabel';
+import OrganizationLogo from '../OrganizationLogo';
+import AuthErrorsContainer from '../AuthErrorsContainer/AuthErrorsContainer';
+
+import styles from './CreateResetPassword.css';
+
+class CreateResetPassword extends Component {
+  static propTypes = {
+    stripes: PropTypes.shape({
+      intl: PropTypes.shape({
```
Godlevskyi: Retrieving the intl object from this.context.intl or the stripes god object (stripes.intl) is a deprecated pattern.
Update it to only use react-intl constructs outlined at https://github.com/folio-org/stripes-core/blob/master/doc/i18n.md
Use <FormattedMessage>, <FormattedDate>, and <FormattedTime> as often as possible, only falling back to formatMessage(), formatDate(), and formatTime() (accompanied by injectIntl) when a bare string is absolutely necessary.

formatDateTime() should be replaced by <FormattedTime> or formatTime() with day/month/year props like https://github.com/folio-org/stripes-core/blob/d0d8f71ec322c9b89ae6ba92cb29179db1cc1704/util/dateUtil.js#L21

Util functions formatDate() and formatDateTime() defined in this repo shouldn't be necessary.

React Intl docs: https://github.com/yahoo/react-intl/wiki

src/components/CreateResetPassword/CreateResetPassword.js
```diff
@@ -0,0 +1,236 @@
...
+            <Form
+              className={styles.form}
+              onSubmit={handleSubmit(onSubmit)}
+            >
+              <div data-test-new-password-field>
+                <Row center="xs">
+                  <Col xs={6}>
+                    <FieldLabel
+                      htmlFor="new-password"
+                      text={<FormattedMessage id={`${this.translateNamespace}.newPassword`} />}
```
maximdidenkoepam: I think we should use intl.formatMessage instead of FormattedMessage in this case.

src/components/CreateResetPassword/CreateResetPassword.js
```diff
@@ -0,0 +1,236 @@
...
+                    />
+                  </Col>
+                </Row>
+              </div>
+              <div data-test-confirm-password-field>
+                <Row center="xs">
+                  <Col xs={6}>
+                    <FieldLabel
+                      htmlFor="confirm-password"
+                      text={<FormattedMessage id={`${this.translateNamespace}.confirmPassword`} />}
```
maximdidenkoepam: Please take a look at comment on line 127

src/components/CreateResetPassword/CreateResetPasswordControl.js
```diff
@@ -0,0 +1,73 @@
...
+        onSubmit={this.handleSubmit}
+        errors={authFailure}
+        stripes={this.props.stripes}
+      />
+    );
+  }
+}
+
+function mapStateToProps(state) {
+  return {
```
maximdidenkoepam: In this case you can use arrow function without return statement

src/components/CreateResetPassword/CreateResetPasswordControl.js
```diff
@@ -0,0 +1,73 @@
...
+      }).isRequired,
+    }).isRequired,
+    stripes: stripesShape.isRequired
+  };
+
+  static defaultProps = {
+    authFailure: []
+  };
+
+  handleSuccessfulResponse() {
```
maximdidenkoepam: Please use the same syntax as in handleSubmit function

src/components/CreateResetPassword/CreateResetPassword.js
```diff
@@ -0,0 +1,234 @@
...
+      passwordMasked: !passwordMasked,
+    }));
+  };
+
+  confirmPasswordFieldValidation = (value, { newPassword, confirmPassword } = {}) => {
+    const isConfirmPasswordInvalid = newPassword && confirmPassword && newPassword !== confirmPassword;
+    const { errors } = this.props;
+
+    if (!isConfirmPasswordInvalid) {
+      _.remove(errors, this.passwordMatchError);
```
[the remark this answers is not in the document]
zburke: The trend is toward importing only individual methods. We are not doing much with [tree shaking](https://webpack.js.org/guides/tree-shaking/) now but I know there are plans to incorporate it into the build process.

PR #491, opened 2018-11-06
Release v2.16.0

cherewaty:
* Set `textComponent` on `<IntlProvider>`
* Fixed `MainNav` not setting the correct app as `selected`.
* Add Create/Reset a Password Screen, STCOR-273
* Fix memory leaks, UIEH-570

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,12 @@
 # Change history for stripes-core
 
-## 2.16.0 (IN PROGRESS)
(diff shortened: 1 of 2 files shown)


cherewaty:
Blocked on some regressions introduced in https://github.com/folio-org/stripes-core/pull/480

cherewaty:
Unblocked by https://github.com/folio-org/stripes-core/pull/493

PR #494, opened 2018-11-08
UIU-689: Add password strength meter. Apply validation rules.

OleksandrAntonenko1:


src/components/CreateResetPassword/CreateResetPassword.js
```diff
@@ -13,6 +13,10 @@ import isEmpty from 'lodash/isEmpty';
 import remove from 'lodash/remove';
 import some from 'lodash/some';
```
maximdidenkoepam: I suggest to combine these imports into single one
`import {
  remove, 
  some, 
  isEmpty
} from lodash`

src/components/CreateResetPassword/CreateResetPassword.js
```diff
@@ -60,28 +71,44 @@ class CreateResetPassword extends Component {
...
     const { errors } = this.props;
+    const error = {
+      code: errorCode,
+      translationNamespace,
+    };
 
-    if (!isConfirmPasswordInvalid) {
-      remove(errors, this.passwordMatchError);
+    if (valid) {
+      remove(errors, error);
```
maximdidenkoepam: It is a bad practice to mutate props. The same for line 111. In this case we should have an action that triggers in particular cases and changes errors array in application state.

src/components/CreateResetPassword/CreateResetPassword.js
```diff
@@ -10,30 +10,37 @@ import {
   formValueSelector,
 } from 'redux-form';
 import isEmpty from 'lodash/isEmpty';
-import remove from 'lodash/remove';
-import some from 'lodash/some';
 
+import {
+  PasswordStrength,
+  PasswordValidationField
```
ViktorSoroka: Please add comma.

src/components/CreateResetPassword/CreateResetPassword.js
```diff
@@ -46,12 +53,15 @@ class CreateResetPassword extends Component {
...
+    this.translationNamespaces = {
+      module: 'stripes-core',
+      smartComponents: 'stripes-smart-components',
+      page: 'stripes-core.createResetPassword',
+      errors: 'stripes-core.errors',
+      button: 'stripes-core.button',
     };
-    this.newPassword = React.createRef();
+    this.newPasswordField = props.stripes.connect(PasswordValidationField);
+    this.passwordMatchErrorCode = 'password.match.error';
```
ViktorSoroka: Let's move it above the connected field so translation codes stay grouped.

src/components/CreateResetPassword/CreateResetPassword.js
```diff
@@ -46,12 +53,15 @@ class CreateResetPassword extends Component {
...
     this.state = {
```
ViktorSoroka: Let's have `state` right after the 

`super(props)` call. So it would be easier to find what is a component state.

src/components/CreateResetPassword/CreateResetPassword.js
```diff
@@ -128,25 +158,43 @@ class CreateResetPassword extends Component {
...
                       hasClearIcon={false}
-                      autoComplete="new-password"
-                      ref={this.newPassword}
-                      withRef
+                      errors={errors}
+                      autoFocus
+                      marginBottom0
+                      fullWidth
+                      token={token}
+                      inputColProps={{
```
ViktorSoroka: Please define this `props` in `constructor` so the new object is not being created upon every render. Pay attention to spacing issue inside an object.

src/components/CreateResetPassword/CreateResetPassword.js
```diff
@@ -128,25 +158,43 @@ class CreateResetPassword extends Component {
...
+                        xs:12,
+                        sm:8
+                      }}
+                      passwordMeterColProps={{
```
ViktorSoroka: The same as above.

webpack/serve.js
```diff
@@ -54,11 +54,20 @@ module.exports = function serve(stripesConfig, options) {
     const port = options.port || process.env.STRIPES_PORT || 3000;
     const host = options.host || process.env.STRIPES_HOST || 'localhost';
 
-    app.use(express.static(`${serverRoot}/public`));
+
```
ViktorSoroka: Please, remove an excess empty line.

ViktorSoroka:
@OleksandrAntonenko1 please ask for a review from @ttamj for Webpack specific stuff.

src/components/CreateResetPassword/CreateResetPassword.js
```diff
@@ -96,13 +131,17 @@ class CreateResetPassword extends Component {
...
     const submissionStatus = submitting || submitSucceeded;
     const buttonDisabled = !isEmpty(errors) || submissionStatus || !(newPassword && confirmPassword);
-    const buttonLabelId = `stripes-core.${submissionStatus ? 'settingPassword' : 'setPassword'}`;
     const passwordType = passwordMasked ? 'password' : 'text';
-    const passwordToggleLabelId = `stripes-core.button.${passwordMasked ? 'show' : 'hide'}Password`;
+    const buttonLabelId = `${this.translationNamespaces.module}.${submissionStatus ? 'settingPassword' : 'setPassword'}`;
+    const passwordToggleLabelId = `${this.translationNamespaces.button}.${passwordMasked ? 'show' : 'hide'}Password`;
+
+    // Todo don't have a back-end yet, should be parsed from token
+    const username = 'diku_admin';
```
ttamj: What side-effects does this have for someone logged in with a user other than diku_admin?  I am weary of hard-coding a username like this even if it is expected to be temporary.  Not everyone may use diku_admin for development and other usernames are certainly valid in demo/live environments.

OleksandrAntonenko1: This variable currently is used for validation rule that checks that password doesn't includes username. So I think it doesn't have any side-effects.

webpack/serve.js
```diff
@@ -54,11 +54,20 @@ module.exports = function serve(stripesConfig, options) {
...
-    app.use(express.static(`${serverRoot}/public`));
+    const staticFileMiddleware = express.static(`${serverRoot}/public`);
+
+    app.use(staticFileMiddleware);
 
     // Process index rewrite before webpack-dev-middleware
     // to respond with webpack's dist copy of index.html
-    app.use(connectHistoryApiFallback({}));
+    app.use(connectHistoryApiFallback({
+      disableDotRule: true,
```
ttamj: Why are we disabling the dot rule?  It is not clear to me from this PR or the related JIRA ticket what the dev server change accomplishes.

OleksandrAntonenko1: Link to this page is consist of path and token. And token usually contains dot in its body(it is normal behavior).  So I decided to disable dot rule to handle this url.

ttamj: Are you saying the token is passed in the URL somewhere?  Could you provide an example request/response for this?

package.json
```diff
@@ -57,6 +57,7 @@
     "stylelint-junit-formatter": "^0.2.1"
   },
   "dependencies": {
+    "@folio/stripes-smart-components": "^1.8.0",
```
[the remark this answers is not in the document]
ttamj: Good catch @zburke.  For reference for you or anyone else trying to follow along, I found the details from MODUSERBL-40 and the sequence diagrams linked within to be very helpful.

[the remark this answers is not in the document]
cherewaty: I initially questioned why `PasswordStrength` and `PasswordValidationField` were built in `stripes-smart-components`: https://issues.folio.org/browse/UIU-516

- `PasswordStrength` appears to be purely presentational and could easily live in `stripes-components` instead.
- It wouldn't be that big of a deal if `PasswordValidationField` just gets repeated a few places. Duplication is always preferred over premature abstractions that back us into corners.

PR #501, opened 2018-11-23
Styles added to each forgot form component

anton-plotnikov:
Stylesheets added to both of Forgot Form components to make them more independent and extendable. Import paths fixed.

### DIFF
--- a/src/components/ForgotPassword/ForgotPasswordForm.css
+++ b/src/components/ForgotPassword/ForgotPasswordForm.css
@@ -1,0 +1,70 @@
+/**
+ * Styling for login and sign up
+ */
+@import "@folio/stripes-components/lib/variables.css";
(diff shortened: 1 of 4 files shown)


src/components/ForgotPassword/ForgotPasswordForm.css
```diff
@@ -0,0 +1,70 @@
+/**
+ * Styling for login and sign up
+ */
+@import "@folio/stripes-components/lib/variables.css";
```
mkuklis: @anton-plotnikov this is pretty cool. I didn't realize we could import these variables from `stripes-components`

anton-plotnikov: seemed convenient to me

PR #504, opened 2018-11-26
STCOM-357 Move IfPermission from stripes-components

cherewaty:
## Purpose
`IfPermission` relies on the `stripes` god object context. That coupling is essential to its abstraction, so it doesn't make sense to refactor to a purely presentational component that just takes a prop. Instead we can move it to `stripes-core`.

https://issues.folio.org/browse/STCOM-357

## Approach
I did a slight refactor to reference the `StripesContext` directly instead of using `withStripes`.


zburke:
Refs [STCOR-295](https://issues.folio.org/browse/STCOR-295)

PR #506, opened 2018-11-27
Fix incomplete interactor selector

cherewaty:
Causes test suite to fail in Safari: https://circleci.com/gh/frontside-folio/stripes-core/1028


cherewaty:
The `stripes-core` tests don't run on Safari in Jenkins - I only caught it because PRs from the `frontside-folio` fork run in Edge and Safari (in addition to Chrome and Firefox) in CircleCI.

PR #512, opened 2018-12-04
load region-specific translations if available

zburke:
react-intl expects locales to look like `pt-BR`, i.e. a locale, a hyphen,
and a region code. lokalise.com provides translation files with names
like `pt_BR.json`, i.e. a locale, an underscore, and a region code. In
other words, the two systems we use for managing translations DON'T KNOW
HOW TO TALK TO EACH OTHER. For reals.

Regardless, the old code threw away region codes altogether, so it
didn't really matter that they couldn't talk because they weren't even
trying. Now, we translate the dash to an underscore in order to look up
the locale + region and use that if available, and fall back to the
locale-only, which is what we used to do, otherwise.

Fixes [STCOR-261](https://issues.folio.org/browse/STCOR-261)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 ## 2.18.0 (IN PROGRESS)
 
 * Sort the settings. Fixes STCOR-286.
+* Load region-specific translations if available. Fixes STCOR-261.
 
 ## [2.17.0](https://github.com/folio-org/stripes-core/tree/v2.17.0) (2018-11-29)
 [Full Changelog](https://github.com/folio-org/stripes-core/compare/v2.16.0...v2.17.0)
--- a/src/loginServices.js
+++ b/src/loginServices.js
@@ -34,9 +34,18 @@ function getHeaders(tenant, token) {
 
 export function loadTranslations(store, locale) {
   const parentLocale = locale.split('-')[0];
+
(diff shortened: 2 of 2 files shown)


ttamj:
One thing the original code failed to guard against was a missing key in the translations object (or a 404).  I believe the list of languages to select from is maintained independently from those available on disk, so it may be worth logging a warning when translations couldn't be found, possibly in a .catch() after the .then()'s.

zburke:
Log a warning in a `catch` clause, or in the `then` clause where the translations are fetched and we can easily provide a default (English?) locale.

ttamj:
Could be either. The translation object _should_ represent all that is available on disk.  Therefore we can rely on checking `region` and `parentLocale` keys in the `.then` with a fair amount of confidence.  

A `catch` could guard against the translation object being wrong.  Come to think of it, in that case we'd see the 404 logged in the network anyway, so additional logging in that case may be redundant.

ttamj:
We may not even need to explicitly set a default here since (I think) the framework performs a fallback already.  Or at least I seem to recall that being the observation leading up to STCOR-261.

zburke:
I think you're right that the framework provides a fallback already, so adding additional error handling here would not solve an additional problem. I'll leave this as-is.

PR #516, opened 2018-12-10
STCOR-289: Move Forgot username and Forgot password links below the L…

OleksandrAntonenko1:
…og in button
# Purpose
When attempting to log in to FOLIO, the location of the "Forgot password?" link places it in the tab order between the username and password fields. This creates a clunky experience when using solely the keyboard to input the username / password.
# Screenshots
screen shot 2018-12-10 at 1 21 45 pm

### DIFF
--- a/src/components/Login/Login.css
+++ b/src/components/Login/Login.css
@@ -20,15 +20,16 @@
 }
 
 .formGroup {
-  margin-bottom: 10px;
+  margin-bottom: 1rem;
 
   &:last-child {
     margin-bottom: 0;
   }
 }
 
+.linksWrapper,
 .authErrorsWrapper {
-  margin-top: 10px;
+  margin-top: 1rem;
 }
 
 input.input {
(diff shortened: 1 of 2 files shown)


src/components/Login/Login.js
```diff
@@ -29,21 +29,21 @@ import styles from './Login.css';
 
 class Login extends Component {
   static propTypes = {
-    handleSubmit: PropTypes.func.isRequired,
-    onSubmit: PropTypes.func.isRequired,
-    submitting: PropTypes.bool,
+    ssoActive: PropTypes.any, // eslint-disable-line react/forbid-prop-types
```
maximdidenkoepam: Why do we use `any` type if default value if `bool`?

src/components/Login/Login.css
```diff
@@ -44,13 +44,16 @@ button.submitButton {
   font-size: var(--font-size-medium);
 }
 
+.linksWrapper {
+  margin-top: 10px;
```
cherewaty: This seems like a good spot for `1rem` or `var(--gutter)` instead of a hard-coded pixel value. Any reason for 10px specifically?

OleksandrAntonenko1: Yeah, you are right. Thank you. I'll fix this in a moment.

PR #518, opened 2018-12-11
fix the import classname

zburke:
`ForgotPassword` defines `ForgotPasswordInteractor`, not
`ForgotUsernameInteractor`.

### DIFF
--- a/test/bigtest/tests/forgotPassword-test.js
+++ b/test/bigtest/tests/forgotPassword-test.js
@@ -16,7 +16,7 @@ describe('forgot password form test', () => {
     this.visit('/forgot-password');
   });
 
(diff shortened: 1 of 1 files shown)


test/bigtest/tests/forgotPassword-test.js
```diff
@@ -16,7 +16,7 @@ describe('forgot password form test', () => {
     this.visit('/forgot-password');
   });
 
-  const forgotUsernamePage = new ForgotUsernameInteractor();
+  const forgotUsernamePage = new ForgotPasswordInteractor();
```
mkuklis: @zburke perhaps the const should be renamed too?

PR #538, opened 2018-12-26
STCOR-305: email hidden on Check email screen

anton-plotnikov:
<h1>Purpose</h1>
<li>To hide an email presented on Check email screen</li>
<h1>Approach</h1>
<li>To create a helper function which replaces all the required symbols with a default '*' sign</li>
<h1>Screenshots</h1>
<li>Expected behavior</li>

check_email

### DIFF
--- a/src/components/CheckEmailStatusPage/CheckEmailStatusPage.js
+++ b/src/components/CheckEmailStatusPage/CheckEmailStatusPage.js
@@ -25,9 +25,9 @@ const CheckEmailStatusPage = (props) => {
   const notificationText = isEmail
     ? `${labelNamespace}.sent.email`
     : `${labelNamespace}.your.email`;
-  const hiddenUserEmail = isEmail
-    ? hideEmail(userEmail)
-    : null;
+  const userEmailMessage = isEmail
+    ? { hiddenUserEmail: hideEmail(userEmail) }
+    : {};
 
   return (
     <div
@@ -52,7 +52,7 @@ const CheckEmailStatusPage = (props) => {
         >
           <FormattedMessage
             id={notificationText}
-            {...(isEmail && { values: { hiddenUserEmail } })}
+            values={userEmailMessage}
           />
         </Headline>
         <Headline
--- a/src/components/CheckEmailStatusPage/CheckEmailStatusPage.js
+++ b/src/components/CheckEmailStatusPage/CheckEmailStatusPage.js
@@ -1,4 +1,4 @@
-import React, { Fragment } from 'react';
+import React from 'react';
 import { withRouter } from 'react-router-dom';
 import { FormattedMessage } from 'react-intl';
 import PropTypes from 'prop-types';
@@ -7,6 +7,7 @@ import Headline from '@folio/stripes-components/lib/Headline';
 
 import OrganizationLogo from '../OrganizationLogo/OrganizationLogo';
 import validateEmail from '../../validators/validateEmail/validateEmail';
+import { hideEmail } from '../../helpers';
 
 import styles from './CheckEmailStatusPage.css';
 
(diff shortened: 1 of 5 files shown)


src/components/CheckEmailStatusPage/CheckEmailStatusPage.js
```diff
@@ -7,6 +7,7 @@ import Headline from '@folio/stripes-components/lib/Headline';
 
 import OrganizationLogo from '../OrganizationLogo/OrganizationLogo';
 import validateEmail from '../../validators/validateEmail/validateEmail';
```
oleksii-popov: Uppercase or lower case in imports? Please be consistent

src/helpers/hideEmail/hideEmail.js
```diff
@@ -0,0 +1,14 @@
+const signToShow = '*';
+
+const replacer = (string, pattern1, pattern2) => (
+  pattern1.concat(signToShow.repeat(pattern2.length))
+);
+
+const hideEmail = email => (
+  email
+    .replace(/(^.{2})(.+?)(?=@)/g, replacer)
```
oleksii-popov: Please add comments what is happening here.
And why do you do it in several steps?

oleksii-popov:
Do we consider tests for this feature? If yes - where can I check them?

src/components/CheckEmailStatusPage/CheckEmailStatusPage.js
OleksandrAntonenko1: I think this imports are fine. Because we are using CamelCase for class based components and lowerCamelCase for functions such as helpers or validators. Guys what do you think about this?

anton-plotnikov: Same point as Sasha's one. `validateEmail` is a validation function and the `OrganizationLogo` is a functional component that is why their namings are different

src/helpers/hideEmail/hideEmail.js
anton-plotnikov: It is possible to replace in a single `String.prototype.replace` step but in that case we should implement a  more complicated `replacer` function with more pattern groups, it's signature is likely to be as `function replacer (string, p1, p2, p3, p4, p5, p6, p7)`  which is, from my point of view, less understandable. What do you think about it?

anton-plotnikov:
> Do we consider tests for this feature? If yes - where can I check them?

Please, check `./test/bigtest/tests/forgotPassword-test.js` for that. 

For now we are checking whether we match a `<FormattedMessage>` text used for the label and, I believe that is enough for bigtest cause testing the function is more about unit tests. I am not sure that we now have them set up.

oleksii-popov:
> > Do we consider tests for this feature? If yes - where can I check them?
> 
> Please, check `stripes-core/test/bigtest/tests/forgotPassword-test.js` for that.
> 
> For now we are checking whether we match a `<FormattedMessage>` text used for the label and, I believe that is enough for bigtest cause testing the function is more about unit tests. I am not sure that we now have them set up. What's your opinion on that?

Ok, we can go with that.

src/components/CheckEmailStatusPage/CheckEmailStatusPage.js
oleksii-popov: Thanks

src/components/CheckEmailStatusPage/CheckEmailStatusPage.js
```diff
@@ -18,51 +19,53 @@ const CheckEmailStatusPage = (props) => {
...
+        <Headline
+          size="x-large"
+          tag="p"
+          bold={false}
+          faded
+          data-test-p-notification
+        >
+          <FormattedMessage
+            id={notificationText}
+            {...(isEmail && { values: { hiddenUserEmail } })}
```
VictorSoroka1: Please consider avoiding `isEmail` check here as you already did it for the `hiddenUserEmail `.

anton-plotnikov: Thank you for pointing that. The thing is that  `isEmail` results in different `id`s, that is why in one case we need to pass the params but in the other - we do not need to do that

PR #544, opened 2018-12-30
STCOR-306: RTL compatibility

rasmuswoelk:
**Ticket:** https://issues.folio.org/browse/STCOR-306

When a user switches language to a language that has right-to-left script direction the UI/layout should reverse. This can be done by adding `dir="rtl"` on the HTML element.

In this PR I made sure that the dir-attribute is updated on the HTML element once the user switches language to e.g. Arabic. Before, the lang-attribute was only updated on page refresh – now it's updated on both page refresh and language change.

## Before
dec-30-2018 21-51-27

## After
dec-30-2018 21-49-57


src/loginServices.js
```diff
@@ -41,6 +42,11 @@ export function loadTranslations(store, locale) {
   // don't know how to talk to each other. sheesh.
   const region = locale.replace('-', '_');
 
+  // Update dir- and lang-attributes on the HTML element
+  // when the locale changes
+  document.documentElement.setAttribute('lang', parentLocale);
+  document.documentElement.setAttribute('dir', isRTL(parentLocale) ? 'rtl' : 'ltr');
```
zburke: Hmmmm. I wonder if we should only update the `lang` and `dir` elements for supported locales, i.e. in the `then((response))...` block near the `store.dispatch(...)` calls rather than up above. In this implementation, somebody could submit a request with an unsupported locale and the display would flip to RTL but the labels would still remain English. That seems incorrect to me. 

What do you think?

src/helpers/isRTL/isRTL.js
```diff
@@ -0,0 +1,17 @@
...
+ */
+
+import rightToLeftLocales from './RTL-locales';
+
+const isRTL = locale => {
+  // Locales will in some cases be formed like e.g. "ar-AR"
+  // But we only need the first part to check for RTL locales
+  const cleanLocale = locale.indexOf('-') >= 0 ? locale.split('-')[0] : locale;
+
+  return rightToLeftLocales.indexOf(cleanLocale) >= 0;
```
zburke: Simple enough. Is there an advantage to maintaining our own function rather than relying on a package like [rtl-detect](https://www.npmjs.com/package/rtl-detect)?

rasmuswoelk: I looked at rtl-detect and actually got the array of locales from this package. I decided to create this simple helper because rtl-detect included extra code/functionality that we don't need.

To me, it doesn't really make a difference. What do you think? Should I import rtl-detect instead?

src/loginServices.js
rasmuswoelk: At first, I had the code within those blocks so that it only changed when that request resolved. 

I moved it since on page refresh it would load the English locale first and once the request resolved it would change the language and the direction. This caused the layout to flash which looked a bit buggy to me.

I was thinking that this approach would be fine since users probably wouldn't be able to change language to any non-existing languages (if they use the UI to change language).

src/helpers/isRTL/isRTL.js
JohnC-80: `rtl-detect` is pretty tiny and has no dependencies - getLangDir might not be useful here, but I feel like we might want it by the time we're through with all of this i18n process. I don't see a terrible downside to just using it instead of maintaining that list ourselves.

rasmuswoelk: Alright – I'll use rtl-detect instead 👍

src/components/MainNav/AppList/AppList.js
```diff
@@ -31,6 +32,9 @@ class AppList extends Component {
     dropdownId: PropTypes.string,
     dropdownToggleId: PropTypes.string.isRequired,
     selectedApp: PropTypes.object,
+    stripes: PropTypes.shape({
+      locale: PropTypes.string.isRequired,
```
cherewaty: We could avoid having to pass around the `stripes` god object here by referring to `locale` from `react-intl` instead. I did a similar thing with `<Datepicker>` where I switched to using `intl.locale` instead of `stripes.locale` (or worse, `stripes.intl.locale`).

Maybe could make use of the new `<IntlConsumer>`? https://github.com/folio-org/stripes-core/pull/535

rasmuswoelk: No problem! I'll make that change 👍

zburke: Good idea! We can apply `<IntlConsumer>` in `MainNav.js` and use it to pass `intl` to `<AppList>` as a prop.

rasmuswoelk:
@zburke @cherewaty 

I updated MainNav to pass intl to AppList. For consistency, I passed intl to ProfileDropdown as well in order to retrieve the current locale directly from react-intl instead of relying on the stripes god object. 

MainNav was already wrapped with the `injectIntl` HoC so I already had access to the intl-object. Do you guys prefer using the `IntlConsumer` or is the current solution sufficient?

PR #556, opened 2019-01-10
An attempt to fix STCOM-302

mkuklis:
Please don't ask questions...
https://issues.folio.org/browse/STCOM-302


mkuklis:
@JohnC-80 I'm hoping this will fix the issue Ann-Marie experienced when editing permission sets but it should also work in other cases when the 'x' is used.

mkuklis:
I ran into couple additional issues with this one so let's ignore it for now.

PR #559, opened 2019-01-14
provide default translations from en.json

zburke:
Include `en.json` as the base for all translations so that keys missing
from a locale file but present in the `en.json` file will show the value
from the `en.json` file rather than the key. This will allow developers
to add pairs to the `en.json` file and have the values immediately show
up in the UI regardless of locale.

Fixes [STCOR-310](https://issues.folio.org/browse/STCOR-310)


ttamj:
I like your solution.  It looks like the test errors are caused in part by the need to stub `fs.accessSync()` so the tests don't actually interact with the file system.

PR #561, opened 2019-01-15
provide default translations from en.json

zburke:
Include `en.json` as the base for all translations so that keys missing
from a locale file but present in the `en.json` file will show the value
from the `en.json` file rather than the key. This will allow developers
to add pairs to the `en.json` file and have the values immediately show
up in the UI regardless of locale.

And this time, do it without merging all translations into those English
translations, thus causing the values from the last locale in the
iteration to overwrite the values from all locales. Ahem. #559. * cough *

Fixes [STCOR-310](https://issues.folio.org/browse/STCOR-310)

### DIFF
--- a/webpack/stripes-translations-plugin.js
+++ b/webpack/stripes-translations-plugin.js
@@ -96,12 +96,20 @@ module.exports = class StripesTranslationPlugin {
   loadTranslationsDirectory(moduleName, dir) {
     logger.log('loading translations from directory', dir);
     const moduleTranslations = {};
+
+    let enTranslations = {};
+    const enPath = path.join(dir, 'en.json');
+    if (fs.existsSync(enPath)) {
+      const rawEnTranslations = StripesTranslationPlugin.loadFile(enPath);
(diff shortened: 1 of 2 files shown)


webpack/stripes-translations-plugin.js
```diff
@@ -96,12 +96,20 @@ module.exports = class StripesTranslationPlugin {
...
+      enTranslations = StripesTranslationPlugin.prefixModuleKeys(moduleName, rawEnTranslations);
+    }
+
     for (const translationFile of fs.readdirSync(dir)) {
       const language = translationFile.replace('.json', '');
       // When filter is set, skip other languages. Otherwise loads all
       if (!this.languageFilter.length || this.languageFilter.includes(language)) {
         const translations = StripesTranslationPlugin.loadFile(path.join(dir, translationFile));
-        moduleTranslations[language] = StripesTranslationPlugin.prefixModuleKeys(moduleName, translations);
+        moduleTranslations[language] = Object.assign({}, enTranslations, StripesTranslationPlugin.prefixModuleKeys(moduleName, translations));
```
VictorSoroka1: So `Object.assign({}` was the culprit. Thanks for managing this @zburke.

PR #563, opened 2019-01-18
UIOR-110 remove extra scrollbars

aliaksei-chumakou:
There is an issue when clicking on dropdown with Apps when extra scrollbars appear. Here screens with them and with the fix. Also it might help with @shaking@ issue from UIOR-110, but I was unable to reproduce all shaking, only partial moving elements due to extra scrollbars, so I suggest to apply this fix to improve UX. At the same time I need help to review it on its correctness, since it's a root element.
extra-scrollbars-at-the-right
no_extra_scrollbars

### DIFF
--- a/src/components/Root/Root.css
+++ b/src/components/Root/Root.css
@@ -1,5 +1,6 @@
 /* stylelint-disable */
 :global(#root) {
   height: 100%;
+  overflow: hidden;
 }
 /* stylelint-enable */


JohnC-80:
This fix creates issues for the login screen where you're unable to scroll down to click the login button. (see screenshot)

We need to investigate to see what's really causing the overflow there - it might be a missplaced margin/padding somewhere or just general misbehavior of Tether (which I'm hoping we can tear out soon)

image

PR #567, opened 2019-01-22
STCOM-411: Copy <AppIcon> to stripes-core

MaksymDryha:
In this PR, the `<AppIcon>` component from `stripes-components` was copied to the `stripes-core` repo. In the future, all ui-modules should use the `<AppIcon>` from `stripes-core`, as the `stripes-components`'s one is deprecated.
Also, the `<NavButton>` and `<AppList>` from `stripes-core` were changed to not use that deprecated `<AppIcon>` from `stripes-components`.
[JIRA issue link](https://issues.folio.org/browse/STCOM-411)


src/components/AppIcon/AppIcon.css
```diff
@@ -0,0 +1,78 @@
...
+
+/**
+ * Label
+ */
+
+.label {
+  margin: 0 0 0 0.35em;
+}
+
+:global([dir="rtl"]) .label {
```
JohnC-80: Probably need to copy https://github.com/folio-org/stripes-components/blob/master/.stylelintrc to this repo, or at least make an exception for `:global` in this repo as exists in `stripes-components` ... it's a weird instance for 'global' to be required, since it isn't a class name that it's wrapping - @rasmuswoelk any thoughts?

rasmuswoelk: I don't think it's needed actually. I think it can be removed.

rasmuswoelk:
I was wondering if we should update the AppIcon-component in stripes-components to not rely on context and simply receive a url/src prop for the icon.

And then the connected logic (accessing the context etc.) could be added here in stripes-core?

But maybe we should do a clean copy to begin with and once it's working as intended, we can make that change?

I think I have seen @JohnC-80 commenting something along those lines somewhere.. What do you think, John?

JohnC-80:
@rasmuswoelk yeah, that's sort of the plan here - 
The idea is to split the AppIcon component between the logic of pulling a ui-module's icon from context and the logic of supplying a static asset. The static asset component would remain in stripes-components under a different name with AppIcon moved to stripes-core since its context also lives there.

PR #574, opened 2019-01-25
support users-bl 4

zburke:
New features were added in users-bl that necessitated bumping its
interface, but without support for the new interface in the UI, we
couldn't build folio-snapshot. Providing parallel support for the old
interface (since no functionality here is changing) and the new
interface (so that backend dependencies can be satisfied) seems like the
least painful solution.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Change history for stripes-core
 
+## 3.1.0 (IN PROGRESS)
+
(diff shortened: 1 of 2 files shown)


JohnC-80:
3.0.2 was released yesterday for the hopeful resolution of multi-react workspaces. 3.0.3?

zburke:
Yeah, looks like it'll be 3.0.3. I'll make that version bump in its own PR and make a formal release.

PR #594, opened 2019-02-06
STCOR-313: Fixed settings translations

Godlevskyi:
## Purpose
This PR is  the fix for https://issues.folio.org/browse/STCOR-313.

Added hardcoded strings in settings to translations.

### DIFF
--- a/src/components/Settings/Settings.js
+++ b/src/components/Settings/Settings.js
@@ -88,20 +88,33 @@ class Settings extends React.Component {
 
     return (
       <Paneset>
-        <Pane defaultWidth="20%" paneTitle="Settings">
+        <Pane
+          defaultWidth="20%"
+          paneTitle={<FormattedMessage id="stripes-core.settings" />}
+        >
           <NavList>
-            <NavListSection activeLink={activeLink} label="Settings" className={css.navListSection}>
+            <NavListSection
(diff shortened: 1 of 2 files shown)


src/components/Settings/Settings.js
```diff
@@ -88,20 +88,33 @@ class Settings extends React.Component {
...
+            <NavListItem to="/settings/about">
+              <FormattedMessage id="stripes-core.front.about" />
+            </NavListItem>
           </NavListSection>
         </Pane>
         <Switch>
           {routes}
           <Route path="/settings/about" component={() => <About stripes={stripes} />} key="about" />
-          <Route component={() => <div style={{ padding: '15px' }}>Choose settings</div>} />
+          <Route component={() => <div style={{ padding: '15px' }}><FormattedMessage id="stripes-core.settingChoose" /></div>} />
```
Yurii-Danylenko: I believe, it's better to move this component and above(for path="/settings/about") to render methods.

zburke: This PR is only about fixing missing translations. If there is other refactoring that should be done to improve its structure, let's handle that in a separate PR.

Godlevskyi: @zburke agreed

PR #598, opened 2019-02-08
STCOR-309 Expose app context menu.

JohnC-80:
# `<AppContextMenu>`
FOLIO UI modules can populate a module-specific context menu by using the provided `<AppContextMenu>` component.

```
/* - index.js 
Ideally the AppContextMenu component is placed at the highest level within a module, so that the dropdown will render outside of any routing that's taking place.
*/
import { AppContextMenu } from '@folio/stripes/core';

    <AppContextMenu>
    {(handleToggle) => (
      <NavList>
        <NavListSection>
          <NavListItem to={packageInfo.stripes.home} onClick={handleToggle}>
            Users Application Home
          </NavListItem>
          <NavListItem onClick={() => { shortcutModalToggle(handleToggle); }}>
            Keyboard Shortcuts
          </NavListItem>
        </NavListSection>
      </NavList>
    )}
  </AppContextMenu>
```

The `handleToggle` renderProp is used to close the menu upon clicking an item.

## Approach

This is created using combination of context and react portals to render the children of the component within the dropdown menu. Despite being rendered as part of the universal navigation, the menu items will retain any of the variables, method calls, and state from its local ui-module implementation.

### Module without menu: 
image

### Module with menu:
image


src/components/MainNav/CurrentApp/README.md
```diff
@@ -0,0 +1,23 @@
...
+
+/* - index.js 
+Ideally the AppContextMenu component is placed at the highest level within a module, so that the dropdown will render outside of any routing that's taking place.
+*/
+import { AppContextMenu } from '@folio/stripes/core';
+
+    <AppContextMenu>
+      {(handleToggle) => (
+        <React.Fragment>
+          <button type="button" onClick={handleToggle}>User Application Shortcuts</button>
```
rasmuswoelk: I think we should use `<Button buttonStyle="dropdownItem">` in this example instead.

src/components/MainNav/CurrentApp/AppContextDropdown.js
```diff
@@ -0,0 +1,58 @@
...
+    const {
+      onToggle,
+      open,
+      icon,
+      displayName,
+    } = this.props;
+
+    const navButtonLabel = (
+      <React.Fragment>
+        <Icon iconRootClass={css.currentAppDropdownCaret} icon={open ? 'caret-up' : 'caret-down'} />
```
rasmuswoelk: Here you can do:

<Icon>{icon}</Icon>

JohnC-80:
@rasmuswoelk: 
The main meat of this PR is exposing a way for modules to create the items in this menu. I think we can move around the ingredients of the main nav rendering in future iterations since it's definitely a subject for further future discussion.

Later with breadcrumbs, would a home link be present within the crumbs or do we just rely solely on the link within the menu? 

Going home seems like it would be a common functionality - do we require 2 clicks to do it at all times? Or one click in some places, 2 clicks in others?

rasmuswoelk:
@JohnC-80 Yes, of course. And I think you solved that in a brilliant way!

I think the original idea was to have one single dropdown (for consistency) which would then have a "Go home"-link as well as other actions – either defined within the module or universally for all modules. You can see an example of this in the prototype: http://ux.folio.org/prototype

And yes – there's a good chance that the "Home"-link will be in the breadcrumbs like seen on this prototype: https://ux.folio.org/universal-navigation/uxi-1

But again; I am not sure if the plans were changed for the app menu – and how breadcrumbs will be structured. That's why I was curious about the two buttons instead of one.

JohnC-80:
@rasmuswoelk Updated this PR to reflect the requirements and have the dropdown's toggle button enclose the app name.

PR #626, opened 2019-02-15
Add ability to configure current user. Refs UICHKIN-53.

mkuklis:
In UICHKIN-53 I need to be able to switch between different service points as a logged in user. This PR allows for extending currentUser object by passing it to BigTest's `setupApplication` as a param.


mkuklis:
@aditya-matukumalli I used assign because it was already used in that file.

aditya-matukumalli:
@mkuklis Sounds good!

PR #638, opened 2019-03-13
make stripes dependencies more strict

zburke:
Caret dependencies are too loose in the context of releasing a platform
and allow new minor release to "leak" up into a platform where the
differing versions can cause duplication errors.

Refs [STRIPES-608](https://issues.folio.org/browse/STRIPES-608)


zburke:
Oh, I see what happened. GitHub's efforts to resolve the merge conflict with CHANGELOG.md resulted in merging master into this branch, subverting the entire purpose of branching this work from the v3.0 commit, i.e. to avoid incorporating all the work that's happened on master since. Thanks, GitHub! That was _super_ helpful.

PR #645, opened 2019-03-15
STCON-76 Option #2 - use ConnectContext Provider

JohnC-80:
# Circular Dependencies option 2
We don't like circ-deps. The changes in this PR are much smaller than that of #644 But the downside is the necessity that only one `stripes-connect` exist in a bundle.

On the stripes-core side, this merely means importing the ConnectContext from `stripes-connect`, and using its `Provider` in place of its own `RootContext.Provider`.

### DIFF
--- a/src/components/Root/Root.js
+++ b/src/components/Root/Root.js
@@ -9,7 +9,8 @@ import { ApolloProvider } from 'react-apollo';
 import ErrorBoundary from '@folio/stripes-components/lib/ErrorBoundary';
 import { metadata } from 'stripes-config';
 
-import { RootContext } from './RootContext';
+import { ConnectContext } from '@folio/stripes-connect';
+// import { RootContext } from './RootContext';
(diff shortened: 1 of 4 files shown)


JohnC-80:
These tests fail, of course, because the capabilities don't exist in `stripes-connect` yet....

src/components/Root/Root.js
```diff
@@ -9,7 +9,8 @@ import { ApolloProvider } from 'react-apollo';
 import ErrorBoundary from '@folio/stripes-components/lib/ErrorBoundary';
 import { metadata } from 'stripes-config';
 
-import { RootContext } from './RootContext';
+import { ConnectContext } from '@folio/stripes-connect';
+// import { RootContext } from './RootContext';
```
zburke: Either delete the commented-code altogether, or leave it in place BUT with a comment explaining what it did and why we don't do it that any longer.

PR #656, opened 2019-04-01
Load @folio/notes as both app and plugin STCOR-357

skomorokh:
This is some temporary hardcoding until we have STRIPES-148 providing
this in general for all modules.

### DIFF
--- a/webpack/stripes-module-parser.js
+++ b/webpack/stripes-module-parser.js
@@ -172,7 +172,7 @@ function parseAllModules(enabledModules, context, aliases) {
     }
 
     // Interim measure to allow ui-notes to offer both an app and a plugin
-    // until STRIPES-148 is complete. See STRIPES-357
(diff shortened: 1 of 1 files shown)


webpack/stripes-module-parser.js
```diff
@@ -170,6 +170,15 @@ function parseAllModules(enabledModules, context, aliases) {
     if (moduleParser.warnings.length) {
       warnings = warnings.concat(moduleParser.warnings);
     }
+
+    // Interim measure to allow ui-notes to offer both an app and a plugin
+    // until STRIPES-148 is complete. See STRIPES-357
```
ttamj: You should update the JIRAs in the comments.  I assume you mean STCOR-148 and STCOR-357 here.

skomorokh: True!

PR #658, opened 2019-04-10
Stcor 285 apply lodash plugins to build

ryandberger:
- Fixes [STCOR-285](https://issues.folio.org/browse/STCOR-285)
- Applied l[odash-webpack-plugin](https://github.com/lodash/lodash-webpack-plugin) and [babel-plugin-lodash](https://github.com/lodash/babel-plugin-lodash) to minimize the size of lodash in our builds
- Compared before and after footprint of lodash in a sample hello world UI app with [webpack-bundle-analyzer](https://www.npmjs.com/package/webpack-bundle-analyzer)

Before:

Screen Shot 2019-04-10 at 3 21 07 PM

After:

Screen Shot 2019-04-10 at 4 09 41 PM


package.json
```diff
@@ -54,16 +56,19 @@
     "@folio/eslint-config-stripes": "^3.2.0",
     "@folio/stripes-cli": "^1.5.0",
     "babel-eslint": "^9.0.0",
+    "babel-plugin-lodash": "^3.3.4",
     "babel-register": "^6.16.3",
     "chai": "^4.1.2",
     "eslint": "^5.5.0",
+    "lodash-webpack-plugin": "^0.11.5",
```
ttamj: The plugins will need to applied as dependencies, rather than devDependencies, otherwise they won't be installed when yarn installing a platform.

package.json
```diff
@@ -46,6 +46,8 @@
     ]
   },
   "devDependencies": {
+    "@babel/cli": "^7.4.3",
+    "@babel/preset-env": "^7.4.3",
```
ttamj: Are you intending to migrate to `@babel/preset-env` from `babel-preset-env` in this PR as well?

package.json
ryandberger: Got it, moved!

package.json
```diff
@@ -63,7 +65,8 @@
     "sinon-chai": "^2.14.0",
     "stylelint": "^8.2.0",
     "stylelint-config-standard": "^17.0.0",
-    "stylelint-junit-formatter": "^0.2.1"
+    "stylelint-junit-formatter": "^0.2.1",
+    "webpack-bundle-analyzer": "^3.3.0"
```
ttamj: You shouldn't need to add the bundle analyzer here; the CLI provides it.  If you want to upgrade it, I'd bump the version in the CLI.

ryandberger: Ah, okay will remove.

package.json
ryandberger: Wouldn't we normally want these as devDependencies since they aid in the Babel transpile process but aren't actually used in the bundle?

ttamj: In a typical app that manages its own build, yes. FOLIO, being distributed across multiple repos, has its build logic and dependencies consolidated in stripes-core.  However, devDependencies are only ever installed for the root package.json.  For FOLIO, the `platform-*` repos are the root of a build (and sometimes `ui-*` modules in a dev environment). Therefore in order to have Babel, Webpack, etc. centrally defined and installed, they must be declared as regular dependencies of stripes-core.

ryandberger: Makes sense. Thanks for explaining!

doytch:
What's the recommended way to import lodash to benefit from these results? Are the following statements identical from a performance/size POV?

`import { get } from 'lodash'`
`import get from 'lodash/get'`

ryandberger:
> What's the recommended way to import lodash to benefit from these results? Are the following statements identical from a performance/size POV?
> 
> `import { get } from 'lodash'`
> `import get from 'lodash/get'`

Both statements would be considered identical since https://www.npmjs.com/package/babel-plugin-lodash will automatically optimize imports when transpiling.

PR #660, opened 2019-04-15
UX-282: Minor updates for 'My Profile' avatar

rasmuswoelk:
Minor style updates for 'My profile' icon as requested in UX-282

Ticket: https://issues.folio.org/browse/UX-282

## Before
image

## After
image


rasmuswoelk:
@JohnC-80 Quick note: I removed the `hideIcon`-prop because it produced a linting error. I thought that since the prop wasn’t being used we might as well remove it altogether.

But I can re-add it and add the conditional again if you think we need it. I think we would do something more simple for breadcrumbs so we would probably not use the NavButton here in my opinion. I can’t currently think of a use-case for a text-only nav button.

PR #663, opened 2019-04-18
Updates STRIPES-611

skomorokh:
Updating lodash, mocha and babel-core to patch some vulns. And sinon because it was really out of date and I was sanity checking the webpack tests anyhow after the mocha update.


zburke:
It isn't clear to me why these builds failed in CI. Does it make any sense to you, @skomorokh?

ryandberger:
> It isn't clear to me why these builds failed in CI. Does it make any sense to you, @skomorokh?

I'm seeing the following error when `yarn install` is running:

`error Couldn't find any versions for "node-environment-flags" that matches "1.0.5"`

zburke:
Right, I saw that in the error log too, but [v1.0.5 was released 11 days ago](https://github.com/boneskull/node-environment-flags/releases) and I can't reproduce the problem locally.

ryandberger:
> Right, I saw that in the error log too, but [v1.0.5 was released 11 days ago](https://github.com/boneskull/node-environment-flags/releases) and I can't reproduce the problem locally.

I can't reproduce locally either. Is there caching of node_modules that can occur on the Jenkins slaves?

zburke:
@funkymalc, I know Jenkins works off a cache, but node-environment-flags v1.0.5 was released on 2019-04-11, which is like a decade ago in JavaScript years, and seems like a long time to wait for a cache to expire. Any idea what's up here?

skomorokh:
I was hoping it was caching but it's still broken now that I'm back after being away from FOLIO for a few days. However, the delay fixed it anyhow because one of our peers at Leipzig ran into the same issue in another context and found out that it mocha 6.x moves node-environment-flags to a peer dep and so we just need to add it alongside mocha. Though I could swear I tested this outside the workspace...

skomorokh:
...I did test it independently of the workspace. I saw that when I went to test it again and found I'd already cloned my branch where I would do that. And it isn't a peer dependency as I'd heard, it very much is a [normal dep](https://github.com/mochajs/mocha/blob/master/package.json#L521). 

And it does work for me locally. I'm kinda stumped. Any ideas? Is it just @funkymalc maintaining CI these days? Might be for the best if another of us had a good grasp on what goes on in that environment...

funkymalc:
@zburke @skomorokh - I manually invalidated the NPM cache on repository.folio.org (which is a bit troubling) and now npm-environment-flags v1.0.5 is available.    Re-ran the jobs in Jenkins and now they are green.

ryandberger:
> @zburke @skomorokh - I manually invalidated the NPM cache on repository.folio.org (which is a bit troubling) and now npm-environment-flags v1.0.5 is available. Re-ran the jobs in Jenkins and now they are green.

@funkymalc, thank you! Is there documentation (or some that could be created) on how to invalidate our NPM cache in case we run into this issue again? Ideally we shouldn't need to, but I'm sure it will necessary again at some point.

PR #664, opened 2019-04-18
Stripes 564 verify tree shaking

ryandberger:
- Fulfills [STRIPES-564](https://issues.folio.org/browse/STRIPES-564) and [STRIPES-581](https://issues.folio.org/browse/STRIPES-581).
- Turned off sideEffects for all non-CSS files and switched bundle output from commonJS to ES Modules to support tree-shaking.
- Since ES Modules don't allow overriding or stubbing imports, added `defaultTranslations` as a prop to allow BigTests to continue to pass.
- Added /dist directory to .gitignore.
- Ran bundle analyzer on a hello world UI app to compare bundle sizes.

Before:

Screen Shot 2019-04-18 at 9 43 24 AM

After:

Screen Shot 2019-04-18 at 9 43 41 AM

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,4 +1,4 @@
 {
   "plugins": ["lodash"],
-  "presets": ["env", "stage-2", "react"]
+  "presets": [["env", { modules: false }], "stage-2", "react"]
 }
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ OVERVIEW.html
 yarn.lock
 yarn-error.log
 artifacts
+dist
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 * Resolve memory leaks and improve performance in dev re-builds. Refs STCOR-296.
 * Reduce lodash footprint using [lodash-webpack-plugin](https://www.npmjs.com/package/lodash-webpack-plugin) and [babel-plugin-lodash](https://www.npmjs.com/package/babel-plugin-lodash). Refs STCOR-285.
+* Turned off sideEffects to enable tree-shaking for production builds. Refs STRIPES-564 and STRIPES-581.
 
 ## [3.3.0](https://github.com/folio-org/stripes-core/tree/v3.3.0) (2019-03-28)
 [Full Changelog](https://github.com/folio-org/stripes-core/compare/v3.2.0...v3.3.0)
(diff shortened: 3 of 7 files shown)


doytch:
This doesn't completely resolve [STRIPES-565](https://issues.folio.org/browse/STRIPES-565) does it? Ie, with this PR, I still can't just `import { setupStripesCore } from '@folio/stripes/core';`, right?

ryandberger:
> STRIPES-565

You're right. I'll drop STRIPES-565 from the description.

ryandberger:
> This doesn't completely resolve [STRIPES-565](https://issues.folio.org/browse/STRIPES-565) does it? Ie, with this PR, I still can't just `import { setupStripesCore } from '@folio/stripes/core';`, right?

You're right. I'll drop STRIPES-565 from the description.

ryandberger:
Based on some concerns from @skomorokh on whether this will be compatible with code splitting, I'm going to plan to hold off on merging this unless I can verify that both work in harmony. https://folio-project.slack.com/archives/CAN13SWBF/p1555624406025700

skomorokh:
I think you can merge this, I don't think it'll make it _worse_, but we just might not benefit from tree-shaking much. Your commit still tidies up some things and none of what I saw said using ES6 vs. commonjs will preclude code-splitting, only that tree-shaking might be disabled for trees included via dynamic import.

What we hope might happen with tree-shaking may not be possible eg. if we want code splitting to only put the input field and button from stripes-components in the bundle loaded at the login page and not include the calendar and kitchen sink, we may have to import things directly from the specific files implementing them and not from an index.js that loads everything and re-exports it.

ryandberger:
> I think you can merge this, I don't think it'll make it _worse_, but we just might not benefit from tree-shaking much. Your commit still tidies up some things and none of what I saw said using ES6 vs. commonjs will preclude code-splitting, only that tree-shaking might be disabled for trees included via dynamic import.
> 
> What we hope might happen with tree-shaking may not be possible eg. if we want code splitting to only put the input field and button from stripes-components in the bundle loaded at the login page and not include the calendar and kitchen sink, we may have to import things directly from the specific files implementing them and not from an index.js that loads everything and re-exports it.

That makes sense. I'll proceed with merging this PR and will evaluate the state of tree-shaking after implementing code-splitting.

PR #666, opened 2019-04-19
sort MainNav apps by name

zburke:
Previously, apps were listed in whatever order they happened to be
returned by the discovery process, which is unpredictable, which is bad.
Manual ordering would be great, but we don't have that yet. Now, we at
least have predictable ordering.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,10 @@
 # Change history for stripes-core
 
-## 3.3.1 (IN-PROGRESS)
+## 3.4.0 (IN-PROGRESS)
 
 * Resolve memory leaks and improve performance in dev re-builds. Refs STCOR-296.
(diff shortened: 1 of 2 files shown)


zburke:
@filipjakobsen, The reasoning you describe makes sense, but I didn't see evidence of it in the code, although, true confessions, I didn't look too hard. Do you have any idea who might have done this work?

filipjakobsen:
@zburke, I am not sure, but I suspect John C, Khalilah or Cate might be able to provide more information.

rasmuswoelk:
@filipjakobsen @zburke 

I can confirm this. 

The current order is determined by the stripes-config so that the order can be controlled manually for each installation. This is, of course, temporary until we get some kind of ordering UI.

zburke:
Closing per @rasmuswoelk.

PR #667, opened 2019-04-19
Added setupStripesCore and startMirage as default exports so they don't need to be addressed by path.

ryandberger:
- Resolves [STRIPES-565](https://issues.folio.org/browse/STRIPES-565).
- Exposes setupStripesCore and startMirage into /test so that full paths aren't required on import. For example, import now looks like `import { setupStripesCore } from '@folio/stripes-core/test';`
- Organized and commented main index.js file in stripes-core.


index.js
```diff
@@ -1,16 +1,24 @@
...
 export { default as TitleManager } from './src/components/TitleManager';
 export { default as HandlerManager } from './src/components/HandlerManager';
-export { default as coreEvents } from './src/events';
 export { default as IntlConsumer } from './src/components/IntlConsumer';
 export { default as AppIcon } from './src/components/AppIcon';
-export { ConnectContext as RootContext, withConnect as withRoot } from '@folio/stripes-connect';
 export { Route, Switch, Redirect } from './src/components/NestedRouter';
+
+/* test scaffolding */
+export { default as setupStripesCore } from './test/bigtest/helpers/setup-application';
```
ttamj: For organization and clarity of the interface, it may be cleaner to group the test scaffolding under a common export like `test` rather than having them commingled with components.

ryandberger: That makes sense, I'll move those to `test`.

PR #670, opened 2019-04-23
heave ho! hoist non-react statics

zburke:
Hoist static assignments from the wrapped component to the exported one.
Huh? Hoist the whatnow? Now hear this:

https://reactjs.org/docs/higher-order-components.html#static-methods-must-be-copied-over

> Sometimes it’s useful to define a static method on a React component. For example, Relay containers expose a static method getFragment to facilitate the composition of GraphQL fragments.
> 
> When you apply a HOC to a component, though, the original component is wrapped with a container component. That means the new component does not have any of the static methods of the original component.…
>
> You can use [hoist-non-react-statics](https://github.com/mridgway/hoist-non-react-statics) to automatically copy all non-React static methods:

Refs [UIREQ-256](https://issues.folio.org/browse/UIREQ-256)

### DIFF
--- a/package.json
+++ b/package.json
@@ -98,6 +98,7 @@
     "graphql": "^0.11.7",
     "hard-source-webpack-plugin": "^0.12.0",
     "history": "^4.6.3",
+    "hoist-non-react-statics": "^3.3.0",
     "html-webpack-plugin": "^3.2.0",
     "isomorphic-fetch": "^2.2.1",
     "jwt-decode": "^2.2.0",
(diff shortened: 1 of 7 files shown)


doytch:
👌 Bellissimo!

I assume this does in fact fix the bugs with `ui-plugin-find-user`?

zburke:
It does indeed.

doytch:
Do you think it's worth handling the rest of the HOCs in stripes-core with this PR? `withModule[s]`, `withAppCtxMenu` and `withLastVisited` or should we just push this through and try to minimise unintended consequences?

zburke:
> Do you think it's worth handling the rest of the HOCs in stripes-core with this PR? 

No time like the present. Stay tuned.

PR #684, opened 2019-05-31
Allow modules to be of more than one type. (STCOR-148)

skomorokh:
Allow modules to be ingested by stripes-core as more than one type -- the `type` property of the `stripes` section is now replaced by `actsAs` which can be a module type string or an array of them. Also removes redundant copies of the icons, Okapi interfaces, and permissions in the `modules.<type>` structure exposed by the `ModulesContext` as they are kept elsewhere.

### DIFF
--- a/webpack/stripes-module-parser.js
+++ b/webpack/stripes-module-parser.js
@@ -49,8 +49,16 @@ class StripesModuleParser {
     if (!Array.isArray(actsAs)) {
       if (typeof actsAs === 'string') actsAs = [actsAs];
       else if (typeof stripes.type === 'string') {
+        this.warnings.push(`Module ${this.moduleName} uses deprecated "type" property. Prefer "actsAs".`);
         actsAs = [stripes.type];
-        if (stripes.hasSettings) actsAs.push('settings');
+        if (stripes.hasSettings) {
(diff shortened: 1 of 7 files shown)


webpack/stripes-module-parser.js
```diff
@@ -36,9 +39,26 @@ class StripesModuleParser {
...
+    const stripes = this.packageJson.stripes;
+    if (!_.isObject(stripes)) {
+      throw new StripesBuildError(`Included module ${this.moduleName} does not have a "stripes" key in package.json`);
+    }
+
+    // Upgrade string actsAs to singleton array and provide compatibility for deprecated options
+    let actsAs = stripes.actsAs;
+    if (!Array.isArray(actsAs)) {
+      if (typeof actsAs === 'string') actsAs = [actsAs];
+      else if (typeof stripes.type === 'string') {
```
ttamj: Looks like this is still backwards compatible with `stripes.type`, nice!  Would we want to maintain at least one test to verify that `stripes.type` is still supported (until some future major version when its not)?

Also this may be a good place to log/warn that folks should upgrade their package.json to `actAs`.

skomorokh: Thanks Matt! You're right, a deprecation warning is in order. Added some.

I think the stripes.type -> actsAs compatibility code is pretty explicitly that and not in danger of being inadvertently removed so I'm comfortable not adding a test for it. While I very much think tests are well worth maintaining I think there is a point of diminishing returns and am more skeptical of the super granular ones. By that token I also didn't include a check to ensure each of the deprecation warnings is appropriately triggered.

PR #687, opened 2019-06-06
provide setCurrency function on stripes

zburke:
Refs [UIU-1040](https://issues.folio.org/browse/UIU-1040)


skomorokh:
This fits with what's there but it kinda seems repetitive at this point, why not just surface `state.okapi`? And, indeed, since it's supposed to represent the tenant config as recorded on the backend, not really have explicit setters for each of these but just a method to refresh it (and an `overrideOkapiState` or whatever for development purposes eg. temporarily setting a different locale locally until per-user settings exist, etc.)

zburke:
> This fits with what's there but it kinda seems repetitive at this point
That's exactly how I felt as I was writing it, but reimplementing it with API changes that would require a new major version this close to a release deadline (2019-06-12) seemed like a bad idea :) Shall I file that as Jira?

skomorokh:
Yes please. Perhaps link it as related to the per-user settings issue, I could see this tying in as something like a `tenantConfig` context distinct from `currentUserConfig` and blended together for most purposes as simply `config`. We can safely enough use context for this as it's not changing a lot so we're not at risk of incurring a bunch of of unnecessary re-rendering or anything...

PR #689, opened 2019-06-14
Include okapiInterfaces in stripes config. Part of STRIPES-634

mkuklis:
This PR should fix the issue found in STRIPES-634. It looks like `okapiInterfaces` was filtered out from stripes config for each module.

https://issues.folio.org/browse/STRIPES-634

### DIFF
--- a/webpack/stripes-module-parser.js
+++ b/webpack/stripes-module-parser.js
@@ -5,7 +5,7 @@ const StripesBuildError = require('./stripes-build-error');
 const logger = require('./logger')('stripesModuleParser');
 
 // These config keys do not get exported with type-specific config
-const TOP_LEVEL_ONLY = ['permissions', 'icons', 'okapiInterfaces', 'permissionSets', 'actsAs', 'type', 'hasSettings'];
+const TOP_LEVEL_ONLY = ['permissions', 'icons', 'permissionSets', 'actsAs', 'type', 'hasSettings'];
 
 function appendOrSingleton(maybeArray, newValue) {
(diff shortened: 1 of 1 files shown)


webpack/stripes-module-parser.js
```diff
@@ -5,7 +5,7 @@ const StripesBuildError = require('./stripes-build-error');
 const logger = require('./logger')('stripesModuleParser');
 
 // These config keys do not get exported with type-specific config
-const TOP_LEVEL_ONLY = ['permissions', 'icons', 'okapiInterfaces', 'permissionSets', 'actsAs', 'type', 'hasSettings'];
+const TOP_LEVEL_ONLY = ['permissions', 'icons', 'permissionSets', 'actsAs', 'type', 'hasSettings'];
```
zburke: @skomorokh, we used to [remove only `type`](https://github.com/folio-org/stripes-core/blob/4fccc75bfbfd10153b55cf8335a0133862261260/webpack/stripes-module-parser.js#L66). We did we add all these other values?

skomorokh: As mentioned in my PR and in the changelog, I attempted to keep things in one place so that eg. you didn't access icons or permissions via `modules.app.foo` as those are consumed elsewhere and made available through other means that should be used instead. Plus there'd be redundant copies now that the metadata not excluded there is placed in the array for each type.

The `okapiInterfaces` I'd considered as being used to generate module descriptors and didn't take into account the about page, total oversight on my part, apologies for the hassle.

PR #693, opened 2019-07-10
Add svg optimization

VictorSoroka1:


ryandberger:
@funkymalc, can we exclude these configuration files from code coverage as there is no way to unit test them?

PR #695, opened 2019-07-12
UIU-1099: Change reset password flow.

OleksandrAntonenko1:
# Purpose:
UI permission [Reset passwords: user can reset password by link] is confusing. If a user is able to create a password by a link then s/he should be able to also reset a password.
# Description:
The previous behaviour of sending a token from the link and obtaining  the new token with username and permissions were changed and currently, we have just one token which we validate and then send with a password to create a new password


OleksiiMaksymov:
The interface version was raised due to a change in the API contract in the 'mod-users-bl' module

src/RootWithIntl.js
```diff
@@ -117,14 +116,9 @@ class RootWithIntl extends React.Component {
...
                     <Switch>
-                      <TitledRoute
-                        name="changePassword"
-                        path="/reset-password/:token"
-                        component={<CreateResetPasswordLoader stripes={stripes} />}
-                      />
                       <TitledRoute
                         name="CreateResetPassword"
-                        path="/change-password/:token/:resetPasswordActionId"
+                        path="/change-password/:token"
```
dmytrotur: Back-end generates link with `/reset-password` path. Could this be changed?

OleksandrAntonenko1: Done.

zburke:
Technically, removing support for interface versions `3.0` and `4.0` means this is a breaking change, requiring a major version update, which would mean a major version update to stripes as well, and therefore to every single UI app. 

Are you _certain_ these changes cannot be made in such a way that allows them to degrade gracefully if only the `3.0` or `4.0` API interface is available?

PR #702, opened 2019-08-09
UIIN-678 receive react via peerDependencies

zburke:
We should only provide react in one place (stripes) and every other
library and app built with stripes should recieve it as a peer. This
reduces the chance of different modules using different minor version of
react with a `~` dependency and pulling multiple versions of it into the
bundle. Multiple versions of React === Bad Things.

Refs [UIIN-678](https://issues.folio.org/browse/UIIN-678)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Change history for stripes-core
 
+## 3.8.0 (IN PROGRESS)
+
+* Receive React as a peerDependency (provide by `stripes`). One dep to rule them all. Refs UIIN-678.
+
 ## [3.7.0](https://github.com/folio-org/stripes-core/tree/v3.7.0) (2019-07-22)
 [Full Changelog](https://github.com/folio-org/stripes-core/compare/v3.6.0...v3.7.0)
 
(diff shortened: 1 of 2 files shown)


skomorokh:
Yes, I agree it shouldn't be in core. But due to hoisting etc. reasons, we unfortunately have to provide react in TWO places --- framework AND platform.

zburke:
Gotta investigate why _framework_ though; seems like we should be able to get rid of that one, right?

skomorokh:
You'd hope. I forget the details, I think it was something like peers had to be satisfied by the immediate parent, but ya, it does seem more likely that if it is at all possible to do it in a single place it'd be at the root.

skomorokh:
> The peerDependencies object guarantees that, for each entry, any package that requires you will be requested to provide a copy of the dependency listed here, ideally matching the version you requested. (from [here](https://yarnpkg.com/blog/2018/04/18/dependencies-done-right/))

...hmm, so it sounds like, theoretically, we could just make it an explicit peer dep in the framework also and then the platform really will get to be the sole arbiter of react version?

skomorokh:
Do we really want to use `*` here though? I think it might be beneficial for things to explicitly list which major versions of React they expect to be compatible with so we know which framework versions and platforms are able to include them. It'll be one more thing to do every major release but seems like useful information to be explicit about.

PR #711, opened 2019-09-04
Calling PRUNE Redux reducer

ryandberger:
whenever switching modules so that application doesn't continuously fill the store (which can lead to eventual degraded performance).

This is to address https://issues.folio.org/browse/UIIN-687 where performance slowly degrades over long periods of time.

NOTE: this PR relies on https://github.com/folio-org/stripes-connect/pull/102 being merged first.

The approach here is to reset stored records when moving between modules since they do take up space in memory and can grow unbounded as you visit more modules and run additional searches. By calling PRUNE whenever switching modules, we ensure that at most only 1 set of records is stored in Redux at any given time, thus making memory usage more reasonable and more predictable.

We will still be retaining search queries and other light pieces of metadata in Redux, so when moving back to a previously accessed module, the search is quickly re-run adding only a slight pause in user experience.

In testing, I observed considerably more memory being released when switching modules vs. the current Chalmers prod environment. That put memory usage closer to around 200MB-300MB on average vs. growing upwards of 700MB and holding.

### DIFF
--- a/src/components/LastVisited/withLastVisited.js
+++ b/src/components/LastVisited/withLastVisited.js
@@ -29,12 +29,25 @@ function withLastVisited(WrappedComponent) {
       });
 
       this.cachePreviousUrl = this.cachePreviousUrl.bind(this);
+      this.previousModule = '';
       this.lastVisited = {};
       this.previous = {};
 
       history.listen((location) => {
         const module = this.getCurrentModule(location);
         if (!module) return;
+
+        if (this.previousModule && this.previousModule !== module.module) {
+          this.props.stripes.store.dispatch({
+            type: '@@stripes-connect/PRUNE',
+            meta: {
+              resource: 'records',
+              module: this.previousModule,
+            }
+          });
+        }
+
+        this.previousModule = module.module;
         const name = module.module.replace(/^@folio\//, '');
         this.previous[name] = this.lastVisited[name];
         this.lastVisited[name] = `${location.pathname}${location.search}`;
--- a/src/components/LastVisited/withLastVisited.js
+++ b/src/components/LastVisited/withLastVisited.js
@@ -37,6 +37,7 @@ function withLastVisited(WrappedComponent) {
         const module = this.getCurrentModule(location);
         if (!module) return;
 
+        // When switching modules, clear the record cache to keep memory from growing unbounded. Refs UIIN-687.
         if (this.previousModule && this.previousModule !== module.module) {
           this.props.stripes.store.dispatch({
             type: '@@stripes-connect/PRUNE',
--- a/src/components/LastVisited/withLastVisited.js
+++ b/src/components/LastVisited/withLastVisited.js
@@ -29,7 +29,7 @@ function withLastVisited(WrappedComponent) {
       });
 
       this.cachePreviousUrl = this.cachePreviousUrl.bind(this);
-      this.previousModule = '';
+      this.previousModules = [];
       this.lastVisited = {};
       this.previous = {};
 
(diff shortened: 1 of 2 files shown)


src/components/LastVisited/withLastVisited.js
```diff
@@ -29,12 +29,31 @@ function withLastVisited(WrappedComponent) {
...
         const module = this.getCurrentModule(location);
         if (!module) return;
+
+        // When switching modules, clear the record cache older than previous module to keep memory from growing unbounded. Refs UIIN-687.
+        if (this.previousModules.length > 1) {
+          if (this.previousModules.indexOf(module.module) < 0) {
+            this.props.stripes.store.dispatch({
+              type: '@@stripes-connect/PRUNE',
+              meta: {
+                resource: 'records',
```
VictorSoroka1: I would add that `ui-data-import` in settings for file extensions, job profiles, mapping profiles uses other record keys. It is done to maintain more than one resource and thus S&S functionality per module. So, in this case, their data won't be pruned.

https://github.com/folio-org/ui-data-import/blob/master/src/settings/FileExtensions/FileExtensions.js#L45

doytch: I don't actually believe the `meta` is doing anything, but fwiw, the `records` in this case are the [`props.resources.foo.records`](https://github.com/folio-org/stripes-connect/pull/102/files#diff-0750d38c98cb1579ed9274950cd696edR51). So I believe that anything that uses `RESTResource` will be pruned correctly, not just those which are explicitly keyed in the manifest as `records`.

src/components/LastVisited/withLastVisited.js
```diff
@@ -29,12 +29,31 @@ function withLastVisited(WrappedComponent) {
...
       this.cachePreviousUrl = this.cachePreviousUrl.bind(this);
+      this.previousModules = [];
       this.lastVisited = {};
       this.previous = {};
 
       history.listen((location) => {
         const module = this.getCurrentModule(location);
         if (!module) return;
+
+        // When switching modules, clear the record cache older than previous module to keep memory from growing unbounded. Refs UIIN-687.
```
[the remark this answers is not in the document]
doytch: Yeah I agree with Victor here. Maybe it's just me but I kinda struggled to keep up with the logic going on in this listener and had to put it down with pen and paper across a few iterations to convince myself that it actually worked. The tricky bit for me to understand was the nomenclature flipping between current and previous modules. 

Like, it took me way too long to realise that the current module will **not** always be in the list of previousModules.

zburke:
@VictorSoroka1, @doytch, I will merge this as-is because, despite concerns about legibility, there is agreement that functionally it is correct and we need to get these changes into the current release cycle. We can always revisit and adjust the implementation details later.

ryandberger:
> @VictorSoroka1, @doytch, I will merge this as-is because, despite concerns about legibility, there is agreement that functionally it is correct and we need to get these changes into the current release cycle. We can always revisit and adjust the implementation details later.

Thanks for merging this @zburke! I will plan to follow-up with a PR to implement the suggested changes (such as moving this code into a method and improving the comments so the logic is clear).

zburke:
The symptoms and comments at [STCOR-392](https://issues.folio.org/browse/STCOR-392) make a strong case for reverting this PR. Do you think that's the right decision, or should we try to debug this, @ryandberger, @doytch, @VictorSoroka1, @mkuklis?

mkuklis:
@zburke I missed this PR but from your discoveries under STCOR-392 it seems like a right decision.

PR #712, opened 2019-09-05
Allow platforms to provide index.html overriding Stripes default

MikeTaylor:
Many thanks for @zburke for providing me with all the information I needed to make this simple change.

### DIFF
--- a/webpack.config.base.js
+++ b/webpack.config.base.js
@@ -1,4 +1,5 @@
 // Common Webpack configuration for building Stripes
+const fs = require("fs");
 const webpack = require('webpack');
 const path = require('path');
 const HtmlWebpackPlugin = require('html-webpack-plugin');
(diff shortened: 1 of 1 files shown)


ttamj:
Nice solution!  @MikeTaylor would you mind documenting?  This and your recent PRs (#709 and #710) all seem well suited for [The Branding Guide](https://github.com/folio-org/stripes/blob/master/doc/branding.md).

MikeTaylor:
Thanks, @ttamj. I added the relevant documentation to the Stripes developer guide, which is a document I'm familiar with, in https://github.com/folio-org/stripes/pull/58

Perhaps someone else, who's familiar with the branding guide, could handle that part of it? (Ideally someone who has allocated time to work on Stripes, which I do not.)

MikeTaylor:
FAO @dcrossleyau ------^

PR #717, opened 2019-09-23
STCOM-581, STCOR-390 react-router updates

zburke:
Update `react-router` and `react-router-dom` to v5 and move them from
dependencies to peerDependencies in order to guarantee that all modules
in a platform are on the same version. [STCOM-581](https://issues.folio.org/browse/STCOM-581)

Updating from v4 to v5 causes warnings in a few components that ask for
`router` in the context but find it to be undefined. In fact, in every
instance where `router` is explicitly pulled from the context, we don't
actually use it. I think it is likely just left over from previous
incarnations where we called `...router.history.push` and it was never
removed when that code was refactored away. [STCOR-390](https://issues.folio.org/browse/STCOR-390)


package.json
```diff
@@ -150,6 +148,8 @@
   },
   "peerDependencies": {
     "react": "*",
-    "react-dom": "*"
+    "react-dom": "*",
+    "react-router": "^5.0.1",
```
NikitaSedyx, line 152: `5.2.0` is already available

zburke:
Replaced by #881. The `react-router` `v5` update should be handled separately.

PR #719, opened 2019-09-25
STCOR-377: Adjust <MainNav> visible nav items based on the width of the viewport

rasmuswoelk:
## Ticket
https://issues.folio.org/browse/STCOR-377

## Changes
- Some structural changes for the MainNav
- Merged List and AppList into one component
- Added ResizeContainer that checks if a NavItem fits in the NavBar and hides it away in the app dropdown if it doesn't

## Manual testing
- Tested and working for RTL

I uploaded the changes here on the URL below if you want to check it out.

http://folio-styling.testing.samhaeng.com

## GIF
Sep-25-2019 16-55-16

Sep-27-2019 14-06-37


JohnC-80:
Nice! I like the fade transition. A little strange that they disappear from the 'beginning' of the list than at the end closest to the menu.. but I'd be willing to wait and hear what users think about it.

rasmuswoelk:
@JohnC-80 Yeah, I agree.. I think I want to change that logic so the apps closest to the app switcher is hidden instead.

src/components/MainNav/NavButton/NavButton.js
```diff
@@ -82,14 +84,25 @@ const NavButton = React.forwardRef(({
 
   let Element = 'span';
   let clickableProps = {};
+  const isInteractable = href || onClick || to;
```
zburke: `isInteractive`, maybe?

PR #721, opened 2019-09-27
STCOR-391 Keep and use existing token

hjiebsco:
Store SSO token as Okapi token and reuse token if needed. 

Fixes [STCOR-391](https://issues.folio.org/browse/STCOR-391)

### DIFF
--- a/src/loginServices.js
+++ b/src/loginServices.js
@@ -248,8 +248,8 @@ function handleLoginError(dispatch, resp) {
   dispatch(setOkapiReady());
 }
 
-function processOkapiSession(okapiUrl, store, tenant, resp) {
-  const token = resp.headers.get('X-Okapi-Token');
+function processOkapiSession(okapiUrl, store, tenant, resp, origToken) {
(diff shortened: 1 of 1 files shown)


src/loginServices.js
```diff
@@ -248,8 +248,8 @@ function handleLoginError(dispatch, resp) {
   dispatch(setOkapiReady());
 }
 
-function processOkapiSession(okapiUrl, store, tenant, resp) {
-  const token = resp.headers.get('X-Okapi-Token');
+function processOkapiSession(okapiUrl, store, tenant, resp, origToken) {
+  const token = resp.headers.get('X-Okapi-Token') || origToken || store.getState().okapi.token;
```
zburke: I don't think we need ` || store.getState().okapi.token` here. This function is called from two places, `requestLogin` and `requestUserWithPerms`. In the former, the `X-Okapi-Token` header will be present. In the latter your change explicitly provides the value provided to `origToken` and that function is only called from `src/components/SSOLanding` when the token value is truthy.

PR #724, opened 2019-10-02
Release v3.10.2

zburke:
Exactly the same as `v3.10.1`, _exactly_, but we published a bug-fix
release in the `v3.7` line after we published `v3.10.1` and that seems
to have gotten npm's `latest` dist-tag. This is causing repositories
that have `^` dependencies to pull that non-latest "latest" release in
addition to the actual latest release that is included in the
`@folio/stripes` package with a `~` dependency.

So really this is just a retag of the existing `v3.10.1` release so that
tools like npm, yarn, and Jenkins will see the semantically-latest
release as the actual latest release. It sorta makes you wonder what the
point of semantic versioning is if `yarn` is going to ignore that in
favor of the `latest` dist-tag, doesn't it?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Change history for stripes-core
 
+## [3.10.2](https://github.com/folio-org/stripes-core/tree/v3.10.2) (2019-10-02)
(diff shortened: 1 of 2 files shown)


zburke:
Nope. Nope nope nope. This was supposed to be opened against the `v3.10` branch, not `master`.

PR #731, opened 2019-10-04
IfPermission: Support render-props form

doytch:
Good UX generally requires an "else" version of the thing you want to conditionally-render if a user has a permission. The only way to do this currently is using `stripes.hasPerm` directly. That's a shame because it means our render functions are more unwieldy and we have to have `stripes` passed to a component.

This adds an extra way of specifying `IfPermission`'s `children`. If `children` are a function, they're called with props that include a `hasPermission` boolean. This allows you to keep both the "if" and "else" cases right next to one another.


src/components/IfPermission/IfPermission.js
```diff
@@ -4,14 +4,20 @@ import { StripesContext } from '../../StripesContext';
...
 const IfPermission = ({ children, perm }) => (
   <StripesContext.Consumer>
-    {stripes => (
-      stripes.hasPerm(perm) ? children : null
-    )}
+    {stripes => {
+      const hasPermission = stripes.hasPerm(perm);
+
+      if (typeof children === 'function') {
+        return children({ hasPermission });
```
MikeTaylor: We don't need to pass the other props down through here?

doytch: Which other props? `perm`? Whoever is using this form will have access to everything they would have passed to the children, since they'll be rendering the children in the same spot they normally would've.

e.g,
```
renderContacts = () => (
  <IfPermission perm="view-user">
    {({ hasPermission }) => hasPermission ? 
      <Pluggable type="find-user" onSelectUser={this.props.onAddContact} />
      : 'You do not have the ability to add internal contacts.'
    }
  </IfPermission>
```

PR #734, opened 2019-10-11
Refactor login form to final form. Part of STCOR-395

mkuklis:
https://issues.folio.org/browse/STCOR-395

This is not what @skomorokh wants but I think it's still an improvement.

Unfortunately we can't completely remove redux-form from `stripes-core` because of: https://github.com/folio-org/stripes-core/blob/master/src/initialReducers.js#L1


skomorokh:
...also isn't confirmation when someone navigates away the only thing stripes-final-form does so I don't really see that it's helping? And I'm still confused why we can't just move to vanilla JS for these two text boxes?

mkuklis:
Good catch @zburke! I think we can probably just rely on react-final-form. I will try that.

I appreciate the article you shared @skomorokh but it felt a bit like a science fiction story to me :) Of course these stories come to life sometimes :). You don't think we can trust the guy behind the final-form?

mkuklis:
@zburke the stripes-final-form should be gone now.

zburke:
p.s. Shhhhh, SonarCloud, it's OK, it's OK. The superficial similarities between these two forms is insignificant and the contortions necessary to abstract the shared code would make it 10k times more difficult to understand. Go back to sleep now. Shhhhhhhhhh.

mkuklis:
Also regarding the security implications I agree with @skomorokh and the article was terrifying. I think we should get there eventually. My primary goal with this PR was to get rid of redux-form warnings.

mkuklis:
@zburke I tried to get the wrong message but I can't reproduce it. The only time I see a different message is when I enter both username and password.

mkuklis:
@zburke never mind I was finally able to reproduce it. I may create another PR with the fix.

PR #735, opened 2019-10-14
Revert "Calling PRUNE Redux reducer"

zburke:
Reverts folio-org/stripes-core#711

Details of the symptoms caused by #711 are in the description at [STCOR-392](https://issues.folio.org/browse/STCOR-392), which I slowly, haltingly tied to this PR at the end of the comments. Ultimately, I think there is a strong case for reverting #711. 

What do you think, merge this to revert 👍 or close this and actually try to debug things 👎?


mkuklis:
Thank you for spending time on this @zburke! Based on your discoveries and the memory graph I say we should roll this back. I think the primary issue with the performance degradation was related epics in stripes-connect. 

It looks like `PRUNE` will clear the query resources which will affect the navigation (causing issues). I think we can review this again but let's fix the navigation first (get rid of query resource as a way to navigate).

ryandberger:
I also support rolling this back, especially given that we were thinking of removing it anyway since the performance gain was minimal and arguably made user experience a little worse by having to refetch search results when moving back to modules. I also recommend that we revert https://github.com/folio-org/stripes-connect/pull/102 for a full cleanup.

PR #743, opened 2019-11-02
Upgrade style-loader to 1.0.0

JohnC-80:
## Overview
As we approach solutions on treatment of global styles within the system, this will be a great upgrade with its treatment of the API (much nicer than pre-1.0.0)

## breakage
Our current webpack.config.cli.dev.js sets a `sourcemaps: true` option for the `style-loader` entry. This has been removed. We've narrowly used the API for this thus far, github search says the `!loader` statements present among the modules leave it at standard usage (no options)

### DIFF
--- a/package.json
+++ b/package.json
@@ -139,7 +139,7 @@
     "rtl-detect": "^1.0.2",
     "rxjs": "^5.4.3",
     "serialize-javascript": "^1.4.0",
-    "style-loader": "^0.22.1",
+    "style-loader": "^1.0.0",
     "svgo": "^1.2.2",
     "svgo-loader": "^2.2.1",
     "tapable": "^1.0.0",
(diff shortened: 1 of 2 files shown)


MikeTaylor:
Sorry, I don't follow why you removed `options: { sourceMap: true }`.

JohnC-80:
It isn't supported by this version of style-loader and throws errors.
Since style-loader is typically the end of the assembly line of other CSS-related loaders, the developers just had it use sourcemaps generated by previous loaders rather than need everyone to have `options: {sourceMap: true}` applied a 2nd time in the chain. Our `css-loader` entry already has `sourceMaps: true` set.

MikeTaylor:
I see. That makes perfect sense.

rasmuswoelk:
👏🏼

Sorry for the late approval.

PR #745, opened 2019-11-11
STCOR-398: Fixed "last visited" state for apps in nav bar

rasmuswoelk:
Updated `<ResizeContainer>` to receive an array of items and return an object of refs mapped by ID which is then applied on the items to determine the widths of the items.

This fixes an issue where the "last visited" app states weren't being preserved because the items were cached by `<ResizeContainer>`. Instead, the original array will be used to map out the items so that any changes to that array will be rendered out.

Updated the tests to work with the new changes.


rasmuswoelk:
@zburke Np! Thank you for bringing this to me attention :)

PR #761, opened 2019-12-02
patch FakeXMLHttpRequest and remove response.

mkuklis:
This ugly PR patches `FakeXMLHttpRequest` based on the discoveries from @VictorSoroka1 in 

https://github.com/pretenderjs/FakeXMLHttpRequest/issues/54

which were introduced by:

https://github.com/pretenderjs/FakeXMLHttpRequest/pull/53/files

The changes introduced here together with the PR from @VictorSoroka1  in:

https://github.com/folio-org/stripes-smart-components/pull/677

should hopefully resolve the test issues. 

This is a temporary fix until we can get the Mirage dependency updated.


mkuklis:
I ended up removing the resolutions. I think we should get this in place now. If the pretenderjs team can change their dependency to `"fake-xml-http-request": "~2.0.0"` as @zburke proposed we should be able to remove it completely.

VictorSoroka1:
@mkuklis, great that you were able to overcome it without specifying deps.

PR #767, opened 2019-12-05
onHome function calls toggleDropdown

MikeTaylor:
When the `showHomeLink` configuration item is set, as is often the
case in development setups, the top right menu with the "Logout" link
also has a "Home" link that goes back to the FOLIO home page. This
link works, but a while back developed the wrinkle that clicking it
did not dismiss the dropdown menu that the link is on. The present
one-line commit fixes that by having the navigate-to-home function
also toggle the menu.


MikeTaylor:
This has been very very slightly irritating me for some time now.

MikeTaylor:
@zburke I've lost track of how the release procedures work these days — is there any problem in going ahead and merging this now?

PR #768, opened 2019-12-09
UIU-1373: Accessibility input fields

annamelnyk:
### Description:
Add proper labels for username, password fields. Fix `autocomplete` attribute value

---

### Issues:
https://issues.folio.org/browse/UIU-1373
https://issues.folio.org/browse/UIU-1374

---

### Screenshots
image


src/components/CreateResetPassword/CreateResetPassword.js
```diff
@@ -187,7 +187,7 @@ class CreateResetPassword extends Component {
                     <Field
                       id="new-password"
                       name="newPassword"
-                      autoComplete="new-password"
+                      autoComplete="on"
```
zburke: Why did this change? `new-password` seems correct to me. Explanation from https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete: 
> When creating a new account or changing passwords, this should be used for an "Enter your new password" or "Confirm new password" field

src/components/CreateResetPassword/CreateResetPassword.js
```diff
@@ -234,7 +234,7 @@ class CreateResetPassword extends Component {
                         inputClass={styles.input}
                         validationEnabled={false}
                         hasClearIcon={false}
-                        autoComplete="confirm-password"
+                        autoComplete="on"
```
zburke: As above, I think this should be `new-password`.

src/components/Login/Login.js
```diff
@@ -135,7 +135,7 @@ class Login extends Component {
                             inputClass={styles.input}
                             validationEnabled={false}
                             hasClearIcon={false}
-                            autoComplete="password"
+                            autoComplete="on"
```
zburke: should this be `current-password`?

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Change history for stripes-core
 
+## [3.11.1] (In Progress)
```
VictorSoroka1: I believe the next version should really be 3.12.0. 

cc @zburke

zburke: True, but the value in the CHANGELOG doesn't have any practical effect so I'm not worried about it. This could merge to a `b3.11` branch if we publish any patches, and/or it might also merge to master and be released as part of the next feature release, i.e. `v3.12.0`.

src/components/CreateResetPassword/CreateResetPassword.js
annamelnyk: @zburke, thanks for explanation link, I agree with your comments, already fixed

PR #780, opened 2020-01-03
STCOR-381: Upgrade Babel from v6 to v7.8

doytch:
This upgrades Babel from v6 to v7. In and of itself, that upgrade doesn't do much for modules. However, as part of the upgrade, we now have the ability to add more Babel plugins. 

As part of the upgrade, I've added two new plugins: [plugin-proposal-nullish-coalescing-operator](https://babeljs.io/docs/en/next/babel-plugin-proposal-nullish-coalescing-operator) and [plugin-proposal-optional-chaining](https://babeljs.io/docs/en/next/babel-plugin-proposal-optional-chaining.html). They add (unsurprisingly), [nullish coalescing](https://github.com/tc39/proposal-nullish-coalescing) and [optional chaining](https://github.com/TC39/proposal-optional-chaining). This means that we should be able to drop a lot of our lodash usage which was generally `_.get()`.

Changes should be fairly self-contained within stripes-core if not using workspaces. If using workspaces, consider upgrading other modules to Babel 7 to avoid polluting the workspace with old copies of Babel that yarn may play and break things with. stripes-connect and stripes-components use Babel to run their tests, and they have matching `babel7` branches if you have those in your workspace.


doytch:
The stripes-core tests rely on stripes-cli which relies on....stripes-core. That dependency is bringing in a _second_ copy of stripes-core that contains the _old_ Babel 6 dependency which breaks the tests.

ttamj:
> The stripes-core tests rely on stripes-cli which relies on....stripes-core. 

Huh... I don't think I ever realized that!  The CLI includes stripes-core for instances like building a UI module stand-alone, but the CLI [makes an effort](https://github.com/folio-org/stripes-cli/blob/master/lib/cli/stripes-core.js) to prefer the stripes-core defined by the app or platform (a pseudo-peer dependency of sorts).  I suppose this logic could be extended to consider that stripes-core might actually be the current working directory.

doytch:
> > The stripes-core tests rely on stripes-cli which relies on....stripes-core.
> 
> Huh... I don't think I ever realized that! The CLI includes stripes-core for instances like building a UI module stand-alone, but the CLI [makes an effort](https://github.com/folio-org/stripes-cli/blob/master/lib/cli/stripes-core.js) to prefer the stripes-core defined by the app or platform (a pseudo-peer dependency of sorts). I suppose this logic could be extended to consider that stripes-core might actually be the current working directory. (If core in fact needs the CLI to run tests... I'm not so sure it really would.)

@ttamj I'm not sure if it _needs_ the CLI to run the tests, but it does need it to run it's `start` script. From the package.json:
```
    "start": "stripes serve",
    "test": "stripes test karma",
    "test:core": "stripes test karma",
```

In any case, it seems like we'll need to either expand that logic or tweak these test commands. Having not looked at either, I'd assume the former is easier?

JohnC-80:
Hmm... looks like `stripes-core` does try to use itself rather than the transitive core for local tests : 
This entry here:
https://github.com/folio-org/stripes-core/blob/master/.stripesclirc.js#L39
which is picked up at
https://github.com/folio-org/stripes-cli/blob/master/lib/cli/stripes-core.js#L43

Looks as though that one bad dep spoils the barrel, but I'll keep sniffing around...

ttamj:
> Having not looked at either, I'd assume the former is easier?

@doytch - Yes and as @JohnC-80 has pointed out, a `.stripesclirc` configuration already exists to handle this. Apparently it is not working though :(  

We may be able to confirm if the alias configuration is getting picked as expected in a stand-alone checkout of stripes-core (master with no changes) and running tests with the [debug setting enabled](https://github.com/folio-org/stripes-cli/blob/master/doc/dev-guide.md#debugging).  For instance: 

```
$ export DEBUG=stripes-cli,stripes-core
```

JohnC-80:
@ttamj it appeared to be working as it should when I debugged it.

The tests seemed to be tripping over an incorrect babel-preset-flow dependency. Seems the syntax 'flow' was automatically resolving to that, so I modified the presets in `.babelrc` to more explicit package names and tests ran!
```
"presets": [
    ["@babel/preset-env", { "modules": false }],
    ["@babel/preset-flow", { "all": true }],
    "@babel/preset-react",
    "@babel/preset-typescript"
  ]
```

doytch:
Pushed on two new commits that:

- Remove the unused `.babelrc` file. It's hanging around there and isn't used for anything AFAIK and is just confusing since we configure babel via `babel-loader-rule.js` and not the config file.
- Removed the nullish coalescing and optional chaining plugins since [they're built-into preset-env as of Babel 7.8](https://babeljs.io/blog/2020/01/11/7.8.0).
- That made me wonder what other duplication we have going on. It wasn't as much as I'd hoped, just json-strings and dynamic-import. The key seems to be the proposal hitting Stage 4 in TC39. At that point it looks like Babel devs move it into preset-env.

PR #783, opened 2020-01-11
STCOR-407: Add miragejs dep and ability to choose mirage server implementation

mkuklis:
This PR adds a new `miragejs` dependency and introduces a new `mirageOptions` option called `serverType` which allows for configuring which mirage server implementation will be used. The default implementation will be set to BigTest mirage for backwards compatibility (the current testing setup for existing ui modules will work as is - it's not a breaking change).

This PR also migrates `stripes-core` tests to `miragejs`.

In order to opt-in for the new miragejs implementation the `serverType` can be set to `miragejs` via `setupApplication` for example:  

https://github.com/folio-org/ui-users/pull/1127/files#diff-afc65168f7c0334e511aba82e26e9977R4

### Migration steps:

1.  `import { ... } from '@bigtest/mirage'` becomes `import { ...} from 'miragejs'`
2. The `faker` has to be installed as a separate dependency and imported directly:  `import faker from 'faker'` instead of `import { faker } from '@bigtest/mirage'`
3. Singular names should be used for factories and models and serializers otherwise miragejs will complain about not being able to find them
4. Add `mirageOptions.serverType = 'miragejs'` in order to opt-in for miragejs implementation:

https://github.com/folio-org/ui-users/pull/1127/files#diff-afc65168f7c0334e511aba82e26e9977R4

### Migration Example:

ui-users has been migrated so far and things seem to be stable there:

https://github.com/folio-org/ui-users/pull/1127

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## 3.12.0 (IN PROGRESS)
 
+* Update Mirage library. Part of STCOR-407.
+
 ## [3.11.1](https://github.com/folio-org/stripes-core/tree/v3.11.1) (2019-12-10)
 [Full Changelog](https://github.com/folio-org/stripes-core/compare/v3.11.0...v3.11.1)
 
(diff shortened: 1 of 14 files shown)


VictorSoroka1:
@mkuklis, this looks great and without a lot of changes.

PR #787, opened 2020-01-13
Rollback miragejs changes

mkuklis:
This PR rolls back the recent changes related to miragejs in case we run into some issues.


zburke:
We had this ready to go in case there were any unforeseen problems with the non-bigtest version of mirage. There were not, so, yay!

PR #789, opened 2020-01-16
FOLIO-2097: Cleanup redux store during app unmount

mkuklis:
When running BigTest tests we are mounting and unmounting `App` for each test loading translations and resources configured via `stripes-connect` into redux store. The store sits in the memory but it's not cleaned up after each unmount. This PR tries to address that.


VictorSoroka1:
@mkuklis, great stuff, I tried just to fire just RESET_STORE, and did not see the difference =)

mkuklis:
I think this is still just a small improvement @VictorSoroka1. What you did in ui-users together with @zburke PR related to `before` should help us even more.

ViktorSoroka:
This is good to have as well.

PR #793, opened 2020-01-23
STCOR-409: Show Change Password link visible up to user's permission

annamelnyk:
# Description
Show/hide `Change Password` link in header `ProfileDropdown` 
# Issue
https://issues.folio.org/browse/STCOR-409
# Screenshot
### Without permission
image
### With permission
image


src/components/MainNav/ProfileDropdown/ProfileDropdown.js
```diff
@@ -89,8 +89,15 @@ class ProfileDropdown extends Component {
 
   createLink(link, index, module) {
     const { stripes } = this.props;
-    const { check } = link;
-    const checkfn = !check ? undefined : (module.getModule()[check] || validations[check]);
+    const { check, route } = link;
+    const checkingFoo = module.getModule()[check] || validations[check];
```
maximdidenkoepam: Please change variable name. I can't understand what is it and why do we need it?

src/components/MainNav/ProfileDropdown/ProfileDropdown.js
```diff
@@ -89,8 +89,15 @@ class ProfileDropdown extends Component {
...
+    let checkfn;
+
+    if (route === '/settings/myprofile/password') {
```
maximdidenkoepam: As I understand this dropdown is always visible, why we check this route?

src/components/MainNav/ProfileDropdown/ProfileDropdown.js
```diff
@@ -89,8 +89,15 @@ class ProfileDropdown extends Component {
...
+      checkfn = !check ? undefined : (stripes.hasPerm('ui-myprofile.settings.change-password') && checkingFoo);
```
maximdidenkoepam: Please try to avoid negative if statement like this. Use check ? (stripes.hasPerm....) : null; Also it is a bad practice assign undefined, I would suggest use null in this cases.

src/components/MainNav/ProfileDropdown/ProfileDropdown.js
annamelnyk: Current method creates links, and with this route checking we can define if current link is "Change password" and here we should check if user has necessary permission. If there would be created another link this permission won't be check.

PR #797, opened 2020-01-28
STCOR-376: Add tests

annamelnyk:
# Description
Increase test coverage of `<SSOLogin>` component to 80%
# Story
https://issues.folio.org/browse/STCOR-376
# Screenshot
image


src/components/SSOLogin/SSOLogin.js
```diff
@@ -24,7 +25,7 @@ function SSOLogin(props) {
       >
         <FormattedMessage id="stripes-core.loginViaSSO" />
       </Button>
-      <form id="ssoForm" />
+      <div id="ssoForm" />
```
VictorSoroka1: Could you please explain the reason why it cannot be a form anymore?

annamelnyk: As I understand SSOLogin button always exists inside `<Login>` component, which already contains <form> tag, therefore in DOM tree we've got <form> tag nested inside <form> tag and that is not valid.

VictorSoroka1: Then I think it makes sense to rename the class `ssoForm` to something else as it is no longer attached to the `form` but to the `div`.

annamelnyk: It looks like I made a mistake removing `<form>` tag, and I already returned it back. This is service tag, and it needs for SSO Login process, for making saml authorization request, so inside this form tag hidden inputs with needed parameters for SAML protocol are creating and then form is submitting.
Thanks for your attentive review by reason of I did not break authorization process.

PR #798, opened 2020-02-03
STCOR-415 replace deprecated webapp-webpack-plugin

zburke:
We replaced `favicons-webpack-plugin` with `webapp-webpack-plugin`
because the former had a memory leak, but the latter is now deprecated
and has been folded back into the former.

Refs [STCOR-415](https://issues.folio.org/browse/STCOR-415), [STCOR-232](https://issues.folio.org/browse/STCOR-232), [STCOR-296](https://issues.folio.org/browse/STCOR-296)

### DIFF
--- a/package.json
+++ b/package.json
@@ -94,6 +94,7 @@
     "debug": "^4.0.1",
     "duplicate-package-checker-webpack-plugin": "^3.0.0",
     "express": "^4.14.0",
+    "favicons-webpack-plugin": "^3.0.1"
     "file-loader": "^1.1.11",
     "final-form": "^4.18.2",
     "graphql": "^0.11.7",
@@ -154,7 +155,6 @@
     "tapable": "^1.0.0",
     "typescript": "^2.8.1",
     "uuid": "^3.0.0",
-    "webapp-webpack-plugin": "^2.6.1",
     "webpack": "^4.10.2",
     "webpack-dev-middleware": "^3.1.3",
     "webpack-hot-middleware": "^2.22.2",
--- a/package.json
+++ b/package.json
@@ -94,7 +94,7 @@
     "debug": "^4.0.1",
     "duplicate-package-checker-webpack-plugin": "^3.0.0",
     "express": "^4.14.0",
-    "favicons-webpack-plugin": "^3.0.1"
+    "favicons-webpack-plugin": "^3.0.1",
     "file-loader": "^1.1.11",
     "final-form": "^4.18.2",
     "graphql": "^0.11.7",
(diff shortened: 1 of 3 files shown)


zburke:
`yarn test:webpack` fails with
```
  1) The stripes-branding-plugin
       apply method
         applies the FaviconsWebpackPlugin:
     TypeError: Cannot read property 'tapPromise' of undefined
      at FaviconsWebpackPlugin.apply (node_modules/favicons-webpack-plugin/src/index.js:111:33)
      at Function.invoke (node_modules/sinon/lib/sinon/spy.js:302:51)
      at FaviconsWebpackPlugin.apply (node_modules/sinon/lib/sinon/spy.js:90:30)
      at StripesBrandingPlugin.apply (webpack/stripes-branding-plugin.js:45:47)
      at Context.<anonymous> (test/webpack/stripes-branding-plugin.spec.js:63:11)
      at processImmediate (internal/timers.js:439:21)
```
and I have no idea what to do with that.

package.json
```diff
@@ -94,6 +94,7 @@
     "debug": "^4.0.1",
     "duplicate-package-checker-webpack-plugin": "^3.0.0",
     "express": "^4.14.0",
+    "favicons-webpack-plugin": "^3.0.1"
```
ttamj:
```suggestion
    "favicons-webpack-plugin": "^3.0.1",
```

zburke: 🤦‍♂

zburke:
> `yarn test:webpack` fails with...

Oh, nevermind. This is all coming back to me. It's a compatibility issue with `html-webpack-plugin` `v3.x`. It wants `v4`, which has been in beta for something like a decade. [There's an issue from October](https://github.com/jantimon/html-webpack-plugin/issues/1291) asking when v4 will be tagged stable, but no response, despite a half-dozen commits since then. 

I wish I at least understood what the hold up was :/

zburke:
Alas, even upgrading `html-webpack-plugin` to `^4.0.0-beta.10` doesn't resolve the test failure.

ttamj:
At the time we replaced favicons-webpack-plugin, we on version `0.0.9`.  Now we are switching back three major versions later at `3.0.1`, so it seems plausible this is no longer the simple library swap it once was.  @zburke Does everything run as expected locally despite the failing test?

zburke:
> Does everything run as expected locally despite the failing test?

Everything appears to run just fine, e.g. integration tests pass and the unit tests (`yarn test`) pass too. It's just the `apply` test in `yarn test:webpack` test that fails.

ttamj:
@zburke you were so close! The test's stubbed compiler object needed to be extended a bit more to accommodate a hook that the prior plugin must not have been referencing.

PR #799, opened 2020-02-04
Gather translations from dependencies listed under stripesDeps STCOR-414

skomorokh:
See https://issues.folio.org/browse/STCOR-414


webpack/stripes-translations-plugin.js
```diff
@@ -40,36 +40,47 @@ module.exports = class StripesTranslationPlugin {
...
-    const fileData = this.generateFileNames(allTranslations);
-    const allFiles = _.mapValues(fileData, data => data.browserPath);
 
     // Hook into stripesConfigPlugin to supply paths to translation files
-    compiler.hooks.stripesConfigPluginBeforeWrite.tap('StripesTranslationsPlugin', (config) => {
+    // and gather additional modules from stripes.stripesDeps
+    compiler.hooks.stripesConfigPluginBeforeWrite.tap({ name:'StripesTranslationsPlugin', context: true }, (context, config) => {
+      Object.values(context.config).forEach(moduleType => moduleType.forEach(moduleConfig => {
+        if (Array.isArray(moduleConfig.stripesDeps)) {
+          moduleConfig.stripesDeps.forEach(dep => {
```
ttamj: It feels like this stripes dep extraction ought to be handled by [parseAllModules](https://github.com/folio-org/stripes-core/blob/62440df2491f1099a145455d7ecf4bf681076ec3/webpack/stripes-module-parser.js#L175-L202), then passed (tapped?) into the translation plugin.

skomorokh: StripesModuleParser is doing 90% of the work still and isn't a plugin so the hook will have to come from StripesConfigPlugin regardless as that's where it's invoked. I suppose it could be its own hook but I kind of like it as context on stripesConfigPluginBeforeWrite as it could plausibly be of use in other things looking to enrich the final config and seemed cleaner than nesting these.

I'm inclined to leave it here for now and move stripesDeps aggregation to the module parser later when adding support for icons. Unless we want to consider them full-on Stripes Modules which I'm a bit uneasy about. Please do join in our somewhat stalled and async #stripes-architecture thread on that...

zburke:
Updates to `.gitignore` notwithstanding`, it looks like there's a bunch of test artifact files included in this PR.

skomorokh:
Nope, those files are part of the tests--rather than stubs and spies I just have some very empty modules with zero byte icons.

PR #803, opened 2020-02-12
Update stripes-react-hotkeys dependency

JohnC-80:


zburke:
I don't understand why this is necessary. Can you explain? 

If we can't import `HotKeys` directly from stripes-components, does that mean it should actually be a peerDependency, i.e. so stripes-core and stripes-components (and any app) are guaranteed to have the same version?

JohnC-80:
You're right, this probably won't be necessary if we can publish stripes-components swimmingly. The dependency chain there (that we discovered yesterday) was concerning...

stripes-components -> stripes-cli -> stripes-core -> stripes-components -> stripes-react-hotkeys ^2.0.0

PR #807, opened 2020-02-24
Added CalloutContext

doytch:
Created a new `CalloutContext` that's fed the ref to a top-level `Callout`. The context can be imported by apps so that they can render callouts that may otherwise be unmounted such as cross-route callouts. I wrote a page in the docs as part of the PR so that should explain it.

Feedback is welcome!


doytch:
> Is there an obvious way to provide a unit test for this, i.e. that creates a `Callout` toast on one route, changes routes, and validates the continued presence of the toast?

Hmmm...maybe? There is an interactor available for Callout that I think I could hook into. I'll take a pop at it tomorrow and see what happens.

doytch:
>  Is there an obvious way to provide a unit test for this, i.e. that creates a Callout toast on one route, changes routes, and validates the continued presence of the toast?

Done, done, done, done, @zburke

PR #809, opened 2020-02-26
Workflow for forcing a user to log in again if their token expires

ryandberger:
- I am still waiting on exact messaging from Khalilah, but wanted to get a review on the approach now.
- The basic idea is that when a user's Okapi token expires (i.e. they haven't called any endpoints within the token TTL), all requests to Okapi return a 401 with body "Invalid token". This is centrally handled in `connectErrorEpic.js` and a dispatch occurs to be handled by `MainNav.js` which contains methods to handle logout. I slightly branched their method to skip resetting navigation, so the user's last action is remembered when they log back in.
- Tests are in progress.


ryandberger:
> This approach looks solid to me, though it's worth noting that "invalid token" is actually different than "expired token". A token will be rejected as "invalid" when the server's private key changes. Imagine changing the locks on your house; your old key no longer works but it hasn't become corrupted or anything. I'm not arguing about the approach taken here, but I just want to be clear about what we're doing because we may, eventually, _have_ an `expires_at` field in our tokens, and we'll have to handle that case separately/additionally.
> 
> Can you add a test to validate that `connectErrorEpic.js` dispatches a `SET_AUTH_FAILURE` when it receives a 401 response with `Invalid token` in the body?

It sounds like in that case, if there is a refresh token endpoint, that would be the appropriate action to take to hopefully keep the user logged in. I'll keep digging in regards to token expiration. 

And yes, I will add that test :)

src/components/MainNav/MainNav.js
```diff
@@ -83,6 +83,13 @@ class MainNav extends Component {
     this._unsubscribe = this.store.subscribe(() => {
       const { history, location } = this.props;
       const module = this.curModule;
+      const state = this.store.getState();
+
+      // If user has timed out, force them to log in again.
+      if (state.okapi && state.okapi.authFailure
+        && filter(state.okapi.authFailure, { type: 'error', code: 'user.timeout' }).length) {
```
mkuklis: @ryandberger would `find` be more appropriate here? You wouldn't need to use `length`.

src/components/MainNav/MainNav.js
```diff
@@ -83,6 +83,13 @@ class MainNav extends Component {
     this._unsubscribe = this.store.subscribe(() => {
       const { history, location } = this.props;
       const module = this.curModule;
+      const state = this.store.getState();
+
+      // If user has timed out, force them to log in again.
+      if (state.okapi && state.okapi.authFailure
```
mkuklis: I think we can now do `state?.okapi?.authFailure`

PR #810, opened 2020-03-02
Added react-router and react-router-dom as devDeps

ryandberger:
since these dependencies are now provided by the platform and tests run outside that context.

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,6 +53,8 @@
     "mocha-junit-reporter": "^1.17.0",
     "react": "^16.8.6",
     "react-dom": "^16.8.6",
+    "react-router": "^5.1.2",
+    "react-router-dom": "^5.1.2",
     "sinon": "^7.3.2",
     "sinon-chai": "^3.3.0",
     "stylelint": "^8.2.0",


zburke:
Oooh, I didn't realize this snuck in the v5 update. I think we need to revert this as it pushes two copies of react-router into workspace builds and all I get is stacktraces everywhere everywhere.

ryandberger:
> Oooh, I didn't realize this snuck in the v5 update. I think we need to revert this as it pushes two copies of react-router into workspace builds and all I get is stacktraces everywhere everywhere.

What is the recommended path forward for `stripes-core` after reverting this? Without this addition, running tests in CI fails since it can't resolve `react-router` and `react-router-dom` because `stripes-components` adds them in as devDeps.

ryandberger:
> > Oooh, I didn't realize this snuck in the v5 update. I think we need to revert this as it pushes two copies of react-router into workspace builds and all I get is stacktraces everywhere everywhere.
> 
> What is the recommended path forward for `stripes-core` after reverting this? Without this addition, running tests in CI fails since it can't resolve `react-router` and `react-router-dom` because `stripes-components` adds them in as devDeps.

We talked about this offline. The v5 versions just need to be reduced to v4. Opened PR #812 to address.

PR #819, opened 2020-03-10
Okapi 509  optional dependencies

MikeTaylor:
This is the Stripes part of OKAPI-509, and allows a UI module to declare optional dependencies on Okapi interfaces.

### DIFF
--- a/util/package2md.js
+++ b/util/package2md.js
@@ -28,12 +28,18 @@ fs.readFile(filename, 'utf8', (err, data) => { // eslint-disable-line consistent
   const json = JSON.parse(data);
   const stripes = json.stripes || {};
   const interfaces = stripes.okapiInterfaces || [];
+  const optional = stripes.optionalOkapiInterfaces || [];
   const md = {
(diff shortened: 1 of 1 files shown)


MikeTaylor:
Wayne writes (on Slack):

> I guess the question is if it will work in CI when the module descriptor is built and published... @malc might be a better choice to review it from that perspective.

The obvious test would be to use it to generate a ModuleDescriptor that has an optional dependency, and check that it loads into Okapi correctly. For UI modules that have no `optionalOkapiInterfaces` (i.e. currently all of them) it will emit exactly the same output as previously.

MikeTaylor:
The emerging picture, from having discussed this with @wafschneider and @funkymalc, is:
 the emerging picture is:

* There’s no reason not to merge this PR.
* But I will need to make another analogous PR against `stripes-cli`, which is now used to do most (probably all) our builds.
* `package2md` is _probably_ not used any more but we are not 100% certain.
* So we should have it emit a noisy message on startup saying “Use stripes-cli instead”. I will file a separate PR to do that.

PR #821, opened 2020-03-16
STCOR-412 correct stripes.hasPerm proptype

zburke:
#817 added many proptypes (yay!) but one was incorrect (whoops).

[STCOR-412](https://issues.folio.org/browse/STCOR-412)


ryandberger:
Oops! Thanks for fixing that!

PR #827, opened 2020-03-19
history location is PropTypes.object, not PropTypes.string

MikeTaylor:
I guess this is the danger we run into when adding prop-types to satisfy ESLint: we do need to also check that the added prop-types are actually correct, otherwise we get big, ugly red messages on the JavaScript console:

> Warning: Failed prop type: Invalid prop `history.location` of type `object` supplied to `Root`, expected `string`.


MikeTaylor:
No problem, it's easily enough done!

PR #828, opened 2020-03-19
By experiment, computedMatch is an object, not an element

MikeTaylor:
I am not 100% happy about this, because when I have `<TitledRoute>` log the value of `computedMatch` is seems to be undefined every time — so I have never need it show us as either an element _or_ an object. Nevertheless, I get console errors when the proptype specifies it as element, and not when I specify it as object.


src/components/TitledRoute/TitledRoute.js
```diff
@@ -12,7 +12,7 @@ import { withStripes } from '../../StripesContext';
 class TitledRoute extends React.Component {
   static propTypes = {
     component: PropTypes.element,
-    computedMatch: PropTypes.element,
+    computedMatch: PropTypes.object,
```
ryandberger: Could you instead say `PropTypes.oneOfType([PropTypes.element, PropTypes.object])`?

MikeTaylor: Sure! (I have no idea what this was trying to achieve, so I'm fine by it.)

PR #844, opened 2020-04-30
throw and log errors that prevent module load

doytch:
This PR logs the errors that are caught when parsing a module during `getModule()`. In tandem with https://github.com/folio-org/stripes-connect/pull/141, these errors become more easily fixable.

**Before:** Useless error message that doesn't say what the issue was.
Screen Shot 2020-04-30 at 2 44 07 PM

**After:** A better chance at debugging the issue based on the error message.
Screen Shot 2020-04-30 at 3 09 08 PM

Throwing **and** logging seems excessive, but:
- We can't just `throw` because React swallows some of the exceptions for reasons I don't fully understand.
- We _shouldn't_ just log when there's a major problem like the module fails to parse.


MikeTaylor:
Love this! Thank you!

It's so, so easy for React to lose track of useful context and report errors which, despite being very verbose, amount to "Something went wrong somewhere, no idea what or wherem, sorry LOL".

PR #845, opened 2020-05-01
Upgrade to react-intl v4

doytch:
- Upgrades to react-intl 4.5.1
- Removes dependency on @folio/react-intl-safe-html in favour of standard FormattedMessage
- Updates all translations to omit `class={className}` that breaks the formatting.
  - @dltj, AFAIK, the bot will overwrite these changes eventually, right? But should I do this at all or just leave it? Should I do something else?
- Removes `intlShape` usage.
- Removes `addLocaleData` usage.

**Read More**
[Upgrade guide from 2 -> 3](https://formatjs.io/docs/react-intl/upgrade-guide-3x)
[4.0.0 breaking changes](https://github.com/formatjs/formatjs/blob/master/packages/react-intl/CHANGELOG.md#400-2020-03-05)


src/loginServices.js
```diff
@@ -59,12 +58,7 @@ export function loadTranslations(store, locale, defaultTranslations = {}) {
     });
   }
 
-  return import(`react-intl/locale-data/${parentLocale}`)
```
VictorSoroka1, line 62: Simplifying this step looks really great.

PR #846, opened 2020-05-01
provide useCurrentApp hook

zburke:
This will be nice to have once we have [react-router v5](https://reacttraining.com/blog/react-router-v5-1/) [in place](https://issues.folio.org/browse/STRIPES-674).


src/moduleRoutes.js
```diff
@@ -68,3 +69,13 @@ function getModuleRoutes(stripes) {
...
 export default getModuleRoutes;
+
+export const useModules = () => useContext(ModulesContext);
+
+export const useCurrentApp = () => {
+  const modules = useModules();
+  const location = useLocation();
+
+  const { app, settings } = modules;
+  return app.concat(settings).find(m => isQueryResourceModule(m, location));
```
mkuklis, line 80: This looks great @zburke. Could we wrap this in `useMemo`?

zburke: Hmmm. I did that, and then I read https://kentcdodds.com/blog/usememo-and-usecallback, and now I don't know if I should have done that 😕

ViktorSoroka:
@zburke, I understand the idea, but could you please clarify the usages for this hook?

zburke:
I have to be honest, @ViktorSoroka07, I don't remember the specific use case I had in mind when I wrote this back in May. It was likely seeing a bunch of functional components elsewhere, and needing to tweak something with routing here in stripes-core and thinking, "I can't get the current app in a functional component, and that's a shame."

Given that there is no specific use case, however, I'm fine closing this to avoid bloat, or merging it commented-out so we have a record of the idea but avoiding bloat.

ViktorSoroka07:
I am fine with either option really. Thank you for clarification @zburke.

PR #856, opened 2020-05-14
[DO NOT MERGE] Fix send callout

aliaksei-chumakou:
so it works with that fix, but it's a huge breaking change, I suppose. Maybe we can come up with a plan to update, since check for every exception when you call `sendCallout()` makes this change lesser evil.

### DIFF
--- a/test/bigtest/tests/callout-context-test.js
+++ b/test/bigtest/tests/callout-context-test.js
@@ -82,8 +82,8 @@ describe('CalloutContext', () => {
   });
 
   describe('navigating to the Context app that shows an error Callout', () => {
-    beforeEach(async () => {
-      await app.nav('Context').click();
+    beforeEach(function () {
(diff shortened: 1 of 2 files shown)


test/bigtest/tests/callout-context-test.js
```diff
@@ -30,7 +30,7 @@ class ContextApp extends Component {
 
 const CalloutFreeApp = () => <h1>No Callouts!</h1>;
 
-describe('CalloutContext', () => {
+describe.only('CalloutContext', () => {
```
zburke, line 33: Remove `.only`

aliaksei-chumakou: @zburke it's on purpose here, this PR is not for merge, just to show what could be the working solution. Or you want to check all tests? I've tried before - there are some fake timeout test failures, not very useful

PR #857, opened 2020-05-14
fix candidate of `null` CalloutContext

aliaksei-chumakou:
Let's discuss this fix candidate, we introduce new context, so we can move to a proper usage step-by-step.
link to PR with failing test https://github.com/folio-org/stripes-core/pull/855
link to PR with example of CalloutContext containing `ref` value instead of `ref.current` https://github.com/folio-org/stripes-core/pull/856


aliaksei-chumakou:
cc @doytch @JohnC-80

PR #862, opened 2020-05-18
STCOR-430: Fix accessibility issue

annamelnyk:
# Description
Fix of issue: `Error message is not read by the screenreader`
# Issue
https://issues.folio.org/browse/STCOR-430
# Screenshot
image


src/components/AuthErrorsContainer/AuthErrorsContainer.js
```diff
@@ -21,7 +21,12 @@ export default class AuthErrorsContainer extends Component {
     const values = parameters.reduce((res, { key, value }) => ({ ...res, [key]: value }), {});
 
     return (
-      <li key={`${code}-${type}`}>
+      <li
+        key={`${code}-${type}`}
+        aria-invalid
```
rasmuswoelk, line 26: To my knowledge, `aria-invalid` is primarily used for form elements and should not have any effect in this case. But you might have found otherwise so I'm curious to hear if you found this attribute to make a difference here?

src/components/AuthErrorsContainer/AuthErrorsContainer.js
```diff
@@ -21,7 +21,12 @@ export default class AuthErrorsContainer extends Component {
...
+        aria-live="assertive"
+        role="alert"
```
rasmuswoelk, line 28: Instead of adding these two attributes for each `<li>`, you can simply add them onto the surrounding `<ul>`. This will automatically announce the messages as they get rendered inside the `<ul>`.

src/components/AuthErrorsContainer/AuthErrorsContainer.js
annamelnyk: No, you are right, thats my mistake, I will remove `aria-invalid` attribute

src/components/AuthErrorsContainer/AuthErrorsContainer.js
annamelnyk: I've tested and if these two attributes are setted to `li` than screen reader automatically read the error message when it appears, and when attributes are setted to `ul` - screen reader does not read it automatically, only if hover to error message. I will leave attributes in `li` elements. Do you agree?

rasmuswoelk: Hmm. Interesting. It works in the Mac screen reader. Which screen reader are you using?

annamelnyk:
@rasmuswoelk, I've tested it with NVDA on Windows only. Great that you have tested it on Mac screen too.

src/components/AuthErrorsContainer/AuthErrorsContainer.js
annamelnyk: I test with NVDA on Windows. I already checked again, the same behavior. Attributes should be in `li`

rasmuswoelk: I just checked out the various implementations of the `<AuthErrorsContainer>` and it's only rendered when there are errors. So that's why it's not working when the attributes are applied to the `<ul>` because it needs to stay in the DOM and then once it's children change (`<li>`), it will read out the updates.

But if it works on the `<li>`'s in both NVDA and JAWS then I think it's fine to leave them on each `<li>`.

However, I would like hear @JohnC-80's opinion on the matter too. I will approve it for now.

PR #866, opened 2020-05-20
Package Dupe Checker updates

doytch:
- turned off warnings
- added more packages that throw errors when duped

behold:
Screen Shot 2020-05-20 at 9 22 25 AM


VictorSoroka1:
Could you please clarify why it is allowed to have duplicated for stripes-smart-components?

doytch:
> Could you please clarify why it is allowed to have duplicated for stripes-smart-components?

[It's not. ](https://github.com/folio-org/stripes-core/blob/f38399f5d470f8bfedfc8d4c4974ddeaf25115b8/webpack/stripes-duplicate-plugin.js#L21)The Github inline diff view is a bit confusing for this PR.

VictorSoroka1:
Yep, just found out that and was typing about it. Great stuff.

webpack/stripes-duplicate-plugin.js
```diff
@@ -7,25 +7,23 @@ const DuplicatePackageCheckerPlugin = require('duplicate-package-checker-webpack
...
-  'stripes-smart-components',
-  'stripes-form',
   'stripes-connect',
-];
-
-// Module names that are acceptable to have duplicates
-const duplicatesAllowed = [
+  'stripes-final-form',
+  'stripes-form',
+  'stripes-smart-components',
```
VictorSoroka1, line 21: Does it makes sense to also add the package about react-intl-safe-html?

zburke: I don't think that one matters. Do you have reason to believe it does?

doytch: I don't think the issue there is dupes of that package, but that it _currently_ implies that somebody is expecting an older version of react-intl. It doesn't even bring in react-intl itself so that's not an issue. And future versions could very well coexist peacefully.

MikeTaylor:
This is great. Honestly, the amount of console crap that I have to ignore (from Yarn, Maven, Stripes, Brew and many others) is a blight. No wonder we miss real errors. I strongly strongly support the suppression of console crap. Thanks, @doytch!

doytch:
Just so it doesn't get missed, this PR does require that the Redux be deduped:
Screen Shot 2020-05-20 at 9 53 46 AM

webpack/stripes-duplicate-plugin.js
VictorSoroka1: Answering to Zak's questions, I just had a concern that multiple versions will not coexist peacefully, but it seems like yeah, the root cause is multiple versions of react-intl, which is already addressed.

zburke:
> Just so it doesn't get missed, this PR does require that Redux be deduped:

Oooh, I don't think we want that. stripes-connect redux-form are both picky about their versions (and they're pretty old versions), but I think in a way that is completely internal. IOW, I'm pretty sure we need to allow `redux` as a dupe, but I don't expect it to be a problem.

MikeTaylor:
> I'm pretty sure we need to allow `redux` as a dupe, but I don't expect it to be a problem.

That's what I thought, too. We may end up with multiple separate Redux stores in the running app, but since each of them supplies its own `dispatch` function, it should all just work out.

doytch:
> Remove `redux` from the list of dis-allowed dupes.

Done.

Also, I added the ability to set `config.warnAboutAllDuplicatePackages: true` in your Stripes config file to show _all_ duplicates, for performance/debugging/masochism reasons.

doytch:
@zburke @mkuklis @skomorokh should we add `rxjs` to the list of disallowed duplicates [given the mess it can cause](https://folio-project.slack.com/archives/C210UCHQ9/p1590099879139600)?

MikeTaylor:
Absolutely! `rxjs` is uncomfortably magical at the best of times. When you have two slightly different flavours of magic happening simultaneously, nothing good can come of it.

mkuklis:
@doytch please go ahead and add rxjs too.

doytch:
Done! I think we're clear to merge this now. Rerequested a review from @zburke based on the rxjs change and if nobody else has objections he's free to merge :)

PR #868, opened 2020-05-21
show app icons in settings navigation

zburke:
Nobody asked for this but I think its helpful. Say, _totally
hypothetically_, you are browsing folio and the locale has been set to
an unfamiliar language and you would like to change it back. Wouldn't it
be convenient if you could see the app icons so you don't have to mouse
over every single option and look at the URL in order to find
`settings/tenant-settings`?

At least, that was the promise. It turns out the settings-only apps
universally don't provide app-icons, so they all show the same empty
grey box for an icon. **sigh** I still think this is pretty though.

Screen Shot 2020-05-21 at 12 18 22 PM (2)


doytch:
This would be cool...can we fallback to a gear icon or something somehow?

rasmuswoelk:
Nice!

I went ahead and tweaked a few things – I hope that's okay!

1. I moved the label as a child of AppIcon. This automatically adds horizontal margin and flips it on RTL.
2. Changed the icon size to small (and added a small CSS tweak to align it properly)

image

zburke:
> I went ahead and tweaked a few things – I hope that's okay!

That's _perfect_!

PR #872, opened 2020-05-25
STCOR-424 stop filtering react-intl permissions warnings

zburke:
`react-intl` is pretty verbose about missing translation values, even
when using its own built-in system for providing default values, which
seems odd. We suppressed those warnings for permissions-related values
for a while to give folks time to update their apps and start providing
entries in their translation files, rather than just displaying the
English-only values pulled from `package.json` files as part of the
permission description. No longer.

I was supposed to do this in time for the 5.0 release on/before May 15. Alas. 
There will likely be a `v5.0.x` patch release and we should include it there. 

Refs [STCOR-424](https://issues.folio.org/browse/STCOR-424)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 * Show app icons in the settings navigation. It's pretty.
 * Make accessible errors at `Login` page. Refs STCOR-430.
+* Stop filtering "missing permission" warnings from `react-intl`. Refs STCOR-424.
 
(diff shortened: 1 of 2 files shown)


doytch:
Any chance you've tested with platform-complete? I wonder how many errors we'll have. 🤔 😆

zburke:
I have not tested this extensively. I'm assuming there will be zillions of errors, even in core. 

Copying the strings out of `package.json` and transforming the keys with the `permission.` prefix for the translations isn't hard but I just lost track of it among the other tasks that blew up in my face.

PR #875, opened 2020-05-28
STCOR-430: Screenreader does not read the list of errors

annamelnyk:
# Issue
When one error in `ul -> li` appears screenreader read error (already works).
When multiple errors appear in `ul -> li` (the bullet for each error message appears as well). And screenreader reads `marker` only, without error message. So I put the message into `span` element
# Fix of task
https://issues.folio.org/browse/STCOR-430
# Screenshot
image


annamelnyk:
Also tested and working on Windiws when using JAWS or NVDA screen readers.

PR #886, opened 2020-07-23
STCOR-440: Update location only if resourceQuery actually changes

mkuklis:
This PR fixes an issue with occasional double `history.push`.

While working on https://issues.folio.org/browse/UIU-1779 I noticed that sometimes the `history.push` executes twice.

Once after we apply search:

https://github.com/folio-org/ui-users/blob/0a4b7257bcbc36450955c9a88c51713c047fefba/src/routes/UserSearchContainer.js#L172

and the second time in:

https://github.com/folio-org/stripes-core/blob/master/src/locationService.js#L75

This currently happens because when the first `history.push` is executed the location is being changed which triggers resource refresh in stripes-connect:

https://github.com/folio-org/stripes-connect/blob/master/connect.js#L149

The refresh fails in:

https://github.com/folio-org/stripes-connect/blob/master/RESTResource/RESTResource.js#L654-L660 (potentially another bug in stripes-connect)

The `abortFetch` updates the redux store causing:

https://github.com/folio-org/stripes-core/blob/master/src/components/MainNav/MainNav.js#L79

to execute again.

I think @JohnC-80 has seen these issues before too during the ui-users refactoring.

### DIFF
--- a/src/components/MainNav/MainNav.js
+++ b/src/components/MainNav/MainNav.js
@@ -15,7 +15,8 @@ import {
   getLocationQuery,
   updateLocation,
   getCurrentModule,
-  isQueryResourceModule
+  isQueryResourceModule,
+  getQueryResourceState,
 } from '../../locationService';
 
 import css from './MainNav.css';
@@ -65,6 +66,8 @@ class MainNav extends Component {
 
   componentDidMount() {
     let curQuery = getLocationQuery(this.props.location);
+    const prevQueryState = {};
+
     this._unsubscribe = this.store.subscribe(() => {
       const { history, location } = this.props;
       const module = this.curModule;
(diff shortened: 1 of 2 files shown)


VictorSoroka1:
> The refresh fails in:
Is it correct that it only fails occasionally?

src/components/MainNav/MainNav.js
```diff
@@ -65,6 +66,8 @@ class MainNav extends Component {
 
   componentDidMount() {
     let curQuery = getLocationQuery(this.props.location);
+    const prevQueryState = {};
```
ryandberger, line 69: Is there a risk of this value getting reinitialized out when you would want the values to persist? If so, considered using `this.prevQueryState` instead.

mkuklis: The `componentDidMount` should only execute once when the `MainNav` is ready so we should be fine here.

mkuklis:
@VictorSoroka1 yes it seems like this happens sporadically. Another questionable thing we do is:

https://github.com/folio-org/stripes-connect/blob/master/connect.js#L149

This doesn't seem correct. I ran into a situation where the location changes and executes the refresh for resources which have nothing to do with location based `?` params. 

The logic there should be a bit smarter.

PR #888, opened 2020-07-29
About page displays version of Stripes platform

MikeTaylor:
As well as stripes-core, stripes-connect and stripes-components

WARNING: I am not currently set up to run this, so I have not tested it. All I know for sure is that it lints cleanly.

What do you think?

### DIFF
--- a/src/components/About/About.js
+++ b/src/components/About/About.js
@@ -2,6 +2,7 @@ import _ from 'lodash';
 import React from 'react';
 import PropTypes from 'prop-types';
 import { FormattedMessage } from 'react-intl';
+import stripesPlatform from '@folio/stripes';
 import stripesConnect from '@folio/stripes-connect/package';
 import stripesComponents from '@folio/stripes-components/package';
 import stripesLogger from '@folio/stripes-logger/package';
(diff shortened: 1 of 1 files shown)


zburke:
This creates a circular dependency between `@folio/stripes` and `@folio/stripes-core` so I don't think it's possible. 

Though it does nicely illustrate why I think we need to separate stripes-core the library from stripes-core the app. If `/settings/about` were displayed by the app, then it _could_ display the version of `@folio/stripes`. It really really bugs me that we don't have the ability to show this, TBH, nor the versions of stripes-smart-components, stripes-form, stripes-final-form or anything else above `stripes-core` in the `@folio/stripes` hierarchy.

MikeTaylor:
Ah yes, the circular dependency. It sucks big-time, but you're right.

BTW., as @skomorokh will attest I have arguing for breaking up stripes-core literally since it first existed. I think the conflation of the library and the app is the single biggest mistake in the system design.

Anyway, since there is no prospect of fixing this any time soon, I guess we just close this PR as WONTDO, right?

zburke:
Can't do, but would really like to do, shoobeedoobeedoo.

PR #889, opened 2020-07-30
Remove unnecessary use of isomorphic-fetch

skomorokh:


skomorokh:
We link to the commit history for that kind of detail, I like to keep the CHANGELOG limited to changes that are actually salient to consumers of the software so they can see at a glance what's interesting this version. Especially since isomorphic-fetch falls back to the browser's fetch implementation when available (eg. always) so this should be a no-op. If we put too much stuff in there it loses the benefit of being a curated selection of interesting changes (not necessarily just API changes, if this was a major refactoring I'd say it merits a mention).

PR #893, opened 2020-08-11
Moving tenant module info retrieval to post-login

ryandberger:
- Satisfies [STRIPES-671](https://issues.folio.org/browse/STRIPES-671)
- Confirmed only OKAPI call that happens before login now is /saml/check to determine SSO status, which has graceful handling for > 400 HTTP response.
- Determining locale occurs after login, so moving to config-based is out of scope for this story.


zburke:
See also #901 for additional necessary work: this here PR moves the tenant-module-retrieval calls to the right place (post-login); #901 updates those calls to use the correct headers.

PR #896, opened 2020-08-17
STCOR-445: Upgraded react-overlays to v.3.2.0.

rasmuswoelk:
We recently upgraded react-overlays from 0.8.x to 3.2.0 in stripes-components and created a replacement component for the now deprecated react-overlays`<RootCloseWrapper>` component.

However, stripes-core still depends on react-overlays@0.8.3 and this version is being pulled in instead, which is causing some bugs when using components (such as the Modal) that rely on a newer version of react-overlays.

This PR fixes that by removing react-overlays as a dependency stripes-core.

## Updates
- Replaced react-overlays `<Transition>` with the react-transition-group `<Transition>` component
- Replaced react-overlays `<Portal>` with React `<Portal>` component
- Imported `<RootCloseWrapper>` from `stripes-components`
- Removed `react-overlays` as a dependency.

## Ticket
https://issues.folio.org/browse/STCOR-445


ViktorSoroka07:
Thank you @rasmuswoelk for help on this.

rasmuswoelk:
@ViktorSoroka07 No problem! Thank you for making me aware of the issue 👍

zburke:
If we need `stripes-core` and `stripes-components` to share a common version of `react-overlays`, could/should we (a) make it a peer to both of them and supply it in `@folio/stripes`, or (b) re-export those components (i.e. `Portal`) from `react-overlays` from `stripes-components` and consume that re-exported/stripes-components-ified package in `stripes-core`? 

The PR resolves the immediate problem of the version conflict, but sets us up to encounter it again when we upgrade to the next major version of `react-overlays` in one package but not the other.

package.json
```diff
@@ -136,10 +136,11 @@
     "react-cookie": "^2.1.4",
     "react-final-form": "^6.3.0",
     "react-hot-loader": "^4.3.10",
-    "react-overlays": "^0.8.3",
+    "react-overlays": "3.2.0",
```
ViktorSoroka07, line 139: I see that in stripes-components, react-overlays is set to "react-overlays": "^3.2.0".

package.json
```diff
@@ -136,10 +136,11 @@
...
     "react-redux": "^5.0.2",
     "react-titled": "^1.0.0",
     "react-transform-catch-errors": "^1.0.2",
+    "react-transition-group": "2.9.0",
```
ViktorSoroka07, line 143: I see that in stripes-components, react-transition-group is set to "react-transition-group": "^2.2.1".

ViktorSoroka07:
Guys, is there any decision on it? Should then the stripes-components PR be reverted while there is no agreement on this one?

JohnC-80:
I think we need to just do the work here to remove stripes-core's direct dependency on react-overlays.

JohnC-80:
If this PR fixes the current issues we're seeing I'm fine with it... so long as we're staying with the new and not reverting...

JohnC-80:
It took us a handful of major versions to upgrade in the first place :D This PR does get us over the immediate issue hump, so I'm for migrating the dependency to a peer in a future development cycle.

rasmuswoelk:
> Guys, is there any decision on it? Should then the stripes-components work on react-overlay be reverted while there is no agreement on this one?

No, we can remove the react-overlays dependency from stripes-core. I just need to make a few changes. I will do this tomorrow.

ViktorSoroka07:
@rasmuswoelk, are you going to proceed with merging this?

rasmuswoelk:
> @rasmuswoelk, are you going to proceed with merging this?

Yes. I just need to figure out what to regarding the version of `react-transition-group`. I will figure this out today.

ViktorSoroka07:
@rasmuswoelk, thank you.

rasmuswoelk:
Merged this in for now. We might need to figure out how to deal with the different versions of react-transition-group at a later point.

PR #907, opened 2020-09-01
Allow for passing custom initial state to BigTest setupApplication

mkuklis:
Due to the change in: https://github.com/folio-org/stripes-core/pull/893/files

The discovery part of the store is only populated after a successful login. This caused some tests to fail because the `discovery` was missing.

This PR allows for passing a customized `initialState` via `setupApplication` so we can populate the `discovery`. 

For example in `ui-users` the `setupApplication` can pass `initialState` similar to:

````js

export default function setupApplication({
  scenarios,
  permissions = {},
  hasAllPerms = true,
  modules,
  translations,
  currentUser,
} = {}) {
  const initialState = {
    discovery: {
      modules: {
        'mod-circulation-16.0.0-SNAPSHOT.253': 'Circulation Module',
        'mod-users-16.0.1-SNAPSHOT.121': 'users',
        'mod-feesfines-15.8.0-SNAPSHOT.73': 'feesfines',
      },
    },
  };

  setupStripesCore({
    mirageOptions,
    scenarios,
    permissions,
    modules,
    translations,
    stripesConfig: {
      hasAllPerms,
    },
    currentUser,
    initialState,
  });
}

````

### DIFF
--- a/test/bigtest/helpers/setup-application.js
+++ b/test/bigtest/helpers/setup-application.js
@@ -29,10 +29,9 @@ export default function setupApplication({
   scenarios,
   currentUser = {},
   userLoggedIn = false,
+  initialState = {},
 } = {}) {
   beforeEach(async function () {
-    const initialState = {};
-
     // when auth is disabled, add a fake user to the store
     if (disableAuth) {
       initialState.okapi = {


K-Felk:
looks like exactly what I originally asked for: a way to tell the environment at startup to load specific modules/interfaces. :)
I assume to get it to load an interface, I'd need to populate the "interfaces"  object within discovery?

mkuklis:
@K-Felk yes the `interfaces` would be under discovery with a similar format to what the `discoveryReducer` now does:

https://github.com/folio-org/stripes-core/blob/master/src/discoverServices.js#L69-L77

ryandberger:
> Wicked smaht.
> 
> @ryandberger, do you think we could get some documentation about the structure of the store pre #893 so we can provide examples to apps that need to configure it with `initialState`?

Sure, I'd be happy to. Where should I document it?

PR #908, opened 2020-09-03
sort interfaces modules depend on

zburke:
Sort the list of interfaces a module depends on, because it's much
easier to find something in a sorted list than an unsorted one.

### DIFF
--- a/src/components/About/About.js
+++ b/src/components/About/About.js
@@ -56,7 +56,7 @@ const About = (props) => {
           <FormattedMessage id="stripes-core.about.moduleDependsOn" values={{ module: `${m.module} ${m.version || ''}` }} />
         </Headline>
         <List
-          items={Object.keys(okapiInterfaces)}
+          items={Object.keys(okapiInterfaces).sort()}
           itemFormatter={itemFormatter}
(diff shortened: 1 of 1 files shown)


mkuklis (approved):

src/components/About/About.js
```diff
@@ -80,9 +80,12 @@ const About = (props) => {
       default:
         headlineMsg = <FormattedMessage id="stripes-core.about.moduleTypeCount" values={{ count: list.length, type: caption }} />;
     }
+
+    list.sort();
+
     return (
       <div key={caption}>
-        <Headline>{headlineMsg}</Headline>
+        <Headline>headline: {headlineMsg}</Headline>
```
ryandberger, line 88: Should this be localized in `en.json`?

MikeTaylor: Why was this text added at all?

zburke: That was debugging; it should be removed. Good catch, guys; thank you.

MikeTaylor (approved):
Good to go!

ryandberger (approved):

PR #909, opened 2020-09-09
STCOR-447: Automatically import CSS variables from stripes-components into UI modules

rasmuswoelk:
## Ticket
https://issues.folio.org/browse/STCOR-447

## Changes
- Upgraded `postcss-custom-properties` to the latest version
- Added `importFrom`-setting to `postcss-custom-properties` to both development and production webpack configs to enable auto imports of CSS variables in UI modules

## Notes
With this change, you will be able to use all of the available CSS variables from `@folio/stripes-components/lib/variables.css` without manually importing them into your CSS file.

### DIFF
--- a/package.json
+++ b/package.json
@@ -124,7 +124,7 @@
     "postcss-calc": "^6.0.0",
     "postcss-color-function": "^4.0.0",
     "postcss-custom-media": "^6.0.0",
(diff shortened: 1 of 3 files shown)


JohnC-80 (approved):
SLIIIICK!! 👍

zburke (approved):
Yes please!

PR #910, opened 2020-09-11
[STCOR-441] Refactor CreateResetPassword to use final-form instead of redux-form

yelyzavetakhokhlova:
[[STCOR-441]](https://issues.folio.org/browse/STCOR-441) Refactor CreateResetPassword to use final-form instead of redux-form

## Purpose 
`CreateResetPassword` should be refactored to use final-from via stripes-final-form instead of redux-form. This will help reduce the number of deprecated libraries in our build, reduce console warnings, and likely provide performance improvements as well.


Vladyslav-Velytskyi (approved):

BogdanDenis (approved):

zburke (approved):
Thank you!

rasmuswoelk (approved):
LGTM. Tested locally but using the snapshot API endpoint.

I noticed some console noise so I pushed a tiny commit to the branch that removes the deprecated `bold`-prop from the `<Headline>` on the success page.

zburke:
Unfortunately, I did not notice when we merged this PR that it introduces a circular dependency between stripes-core and stripes-final-form. We'll either need to revert this PR or issue another one that uses bare-naked final-form/react-final-form without the `stripesFinalForm()` wrapper.

PR #913, opened 2020-09-21
Centralizing module scope regex

ryandberger:
allowing for other scope names besides 'folio' allowing for _ and - characters in name.

Fulfills [STCOR-456](https://issues.folio.org/browse/STCOR-456)


evaluk (commented):
I know that we encountered the - and _ characters not being supported, but urls allow other characters like
 "-" , "." , "_" , "~", so . and ~ could be added as possible characters if we wanted.  Would it be too much to add those other characters?

ryandberger:
> I know that we encountered the - and _ characters not being supported, but urls allow other characters like
> "-" , "." , "_" , "~", so . and ~ could be added as possible characters if we wanted. Would it be too much to add those other characters?

No problem. I'll add them as well.

zburke:
Thanks for cleaning up all those regexes to point to a single constant; that's much nicer!

As long as we're being thorough about the rules, the [nodejs docs](https://docs.npmjs.com/files/package.json#name) offer a little more precision. Can/should we update the comment in `packageName.js` about where these rules came from? 

> * The name must be less than or equal to 214 characters. This includes the scope for scoped packages.
> * The name can’t start with a dot or an underscore.
> * New packages must not have uppercase letters in the name.
> * The name ends up being part of a URL, an argument on the command line, and a folder name. Therefore, the name can’t contain any non-URL-safe characters.

mkuklis (approved):

ryandberger:
> Thanks for cleaning up all those regexes to point to a single constant; that's much nicer!
> 
> As long as we're being thorough about the rules, the [nodejs docs](https://docs.npmjs.com/files/package.json#name) offer a little more precision. Can/should we update the comment in `packageName.js` about where these rules came from?
> 
> > * The name must be less than or equal to 214 characters. This includes the scope for scoped packages.
> > * The name can’t start with a dot or an underscore.
> > * New packages must not have uppercase letters in the name.
> > * The name ends up being part of a URL, an argument on the command line, and a folder name. Therefore, the name can’t contain any non-URL-safe characters.

Done!

PR #914, opened 2020-09-21
Add okapiKy helpers

skomorokh:
Documented in doc/okapiKy.md, these provide an especially convenient
API for when you want to send HTTP requests to Okapi rather then let
stripes-connect manage a resouce. The previous alternative was an
explicit use of window.fetch where you'll need to feed it headers from
useStripes() and handle errors in at least two places (response with an
HTTP error code and an exception). Ky is a light wrapper that cleanly
lets us inherit the headers and streamlines error handling.


mkuklis (approved):
It looks great. Thank you for this contribution @skomorokh.

JohnC-80 (approved):

PR #916, opened 2020-09-21
Add Korean and Hindi (India)

dltj:
Ref [STCOR-458 Add Korean and Hindi languages - FOLIO Issue Tracker](https://issues.folio.org/browse/STCOR-458)


src/loginServices.js
```diff
@@ -58,8 +58,10 @@ export const supportedLocales = [
   'fr-FR',
   'de-DE',
   'he',
+  'hi_IN',
```
zburke, line 61: I think this should be `hi-IN`.

zburke (changes requested):

src/loginServices.js
dltj: Ack — you're right.  Thank you.  I'm presuming the language list in ui-developer is also needed since that list is still there.

zburke (approved):

src/loginServices.js
zburke: I forgot ui-developer still had its own list; I just pushed up a PR [to remove its local copy](https://github.com/folio-org/ui-developer/pull/164). We already [factored the local copy out](https://github.com/folio-org/ui-tenant-settings/pull/140) of ui-tenant-settings.

dltj: Makes sense—thank you so much!  I pulled [UID-28](https://issues.folio.org/browse/UID-28) into Core F Sprint 98 along with STCOR-458.  I'll update and close them.

PR #919, opened 2020-09-24
Upgrade apollo client

mkuklis:
It looks like version >= 3 changed the imports a bit.


MikeTaylor (approved):
_sigh_ Another day, another backwards-incompatible JavaScript library update.

PR #923, opened 2020-10-01
STCOM-452 actually validate the stored token

zburke:
Previous implementations tested the validity of a stored token by using
to conduct a dummy fetch. This was problematic for two reasons:

* It was an extra request and we just threw away the response as long as
it was a 200.
* Even if the token was valid, the request would still fail if the user
retrieved from the token didn't have permission on that endpoint.

Now, we parse the token and validate its shape, which is what this
function was supposed to be doing all along.

Refs [STCOR-452](https://issues.folio.org/browse/STCOR-452)


zburke:
@MikeTaylor, @skomorokh: this function used to do server two purposes: 
* validate the token (by making a dummy fetch request)
* validate that Okapi is alive (by catching exceptions from the fetch; it would redirect to the login unless it got a 200, but if the fetch actually threw it would dispatch `store.dispatch(setServerDown());` instead)

We are choosing to remove the fetch for the reasons detailed in the ticket, but also because the Okapi v4 upgrade means that all API requests require permissions. Well, almost all, since presumably authentication-related endpoints themselves need to be public. 

The point is, where do you think we _should_ be dispatching `setServerDown()`, if not here?

zburke:
@funkymalc, this should help resolve some of the Okapi v4 upgrade issues with stripes. 

The [fetch call to `saml/check`](https://github.com/folio-org/stripes-core/blob/513bd7cb71b49519a9513ec047722dcd76027ee4/src/loginServices.js#L240-L256) is still in place. LMK if we need to do something different to establish whether SAML is available.

MikeTaylor (commented):
Honestly, this change looks wrong-headed to me. The new approach (A) involves knowing details of implementation, (B) doesn't actually check whether the token is valid, only whether it's syntactically OK, and (C) can't detect certain classes of error — hence your question about where we should invoke `setServerDown`, to which the only possible answer is "when we have determined that the server is down, which we can only do my making a request to it".

Certainly fetching a ton of users, as we did previously, is far from optimial. But my inclination here would be to keep the same approach as before and just use a better endpoint for the probe. As you suggest, an authentication-based one seems like the way to go if as suggested it won't need permissions that the user might lack. And whatever request we make, we should include `limit=0` to avoid actually pulling records which we then discard.

zburke:
@adamdickmeiss, @craigmcnally, at present, stripes-core makes a dummy call to {{/users}} as a token-validation strategy. 

As @MikeTaylor points out, this approach is not without merit: it truly validates the legitimacy of the token rather than just making sure it parses; it validates that Okapi is up. But it does have some problems, namely, if the user in question doesn't have access to {{/users}} then this implementation fails (this was the original bug in [STCOR-452](https://issues.folio.org/browse/STCOR-452)). 

Is there a different endpoint than {{/users}} all authenticated users _do_ have access to that we could use here? Alternatively,  is there a different strategy you recommend for token-validation?

MikeTaylor:
Note that what we ideally want is an endpoint that all _authenticated_ users are authorized for, but a user who is not logged in at all cannot access. It may be that there are no such endpoints, or even that such an endpoint is not even possible; if so, then one that anonymous users can also access would be adequate, especially now we also have the code to syntactically validate the token.

zburke:
I know we don't use refresh and access tokens, but if we did, such an endpoint would be ideal for this purpose, i.e. globally accessible to authenticated users. Even if we don't fully engage that machinery here, is there some piece of it we could leverage? 

e.g. could we make a request to `/some/token/refresh/endpoint` the same way we currently due to `/users` and leave the existing implementation in place, throwing away the actual response (because we haven't implemented handling of refresh and access tokens yet) but taking action based on the status code?

mkuklis:
I kind of like what @zburke did here. It's simpler. I don't think we ever had what @MikeTaylor described in point B) even on the server.

The `setServerDown` could be executed if `loadResources` fail. That would indicate something is not right with the server.

If we have a separate clean endpoint for the token validation we could revisit this in the future but I would just go ahead with this.

mkuklis (approved):

mkuklis:
Or maybe if we hear back from @adamdickmeiss, @craigmcnally we should wait and see if that extra endpoint could be possible.

zburke:
That's an interesting point about dispatching `setServerDown` in `loadResources`, @mkuklis. Maybe the 
```
.catch(() => {
    store.dispatch(setServerDown());
  })
```
block properly belongs at the bottom of ` discoverServices()` [after the final `then` block](https://github.com/folio-org/stripes-core/blob/513bd7cb71b49519a9513ec047722dcd76027ee4/src/discoverServices.js#L51-L53)? (Would the nested `catch` clauses in their currently need to re-throw?)

mkuklis:
@zburke I would probably use `Promise.all` and add all 4 calls to it: `getLocale`, `getPlugins`, `getBindings`, `discoverServices` (last one only if that `if` statement is valid) and then attach single `catch` something similar to:

````js
const promises = [
  getLocale(okapiUrl, store, tenant);
  getPlugins(okapiUrl, store, tenant);
  getBindings(okapiUrl, store, tenant);
];

if (!store.getState().okapi.withoutOkapi) {
 promises.push(discoverServices(store));
}

Promise.all(promises).catch(() => store.dispatch(setServerDown());
````

zburke:
I like that idea, @mkuklis, but it also points to a need for a wider refactor than what I have aimed at here. A few things I noticed: 

* `getBindings()` dispatches `setOkapiReady()` but that should probably be handled at the end of `loadResources`. Agreed? 
* what is the use-case for `withoutOkapi`? 
* we could really use some comments in this file explaining what these functions do, and outlining the overall flow of things. You might think `getBindings()` is obvious, but there are no longer any default key-bindings shipped with the reference envs, so you can't inspect the response for a hint, and it might take you an embarrassingly long time to figure out, "Oh, right, _key_ bindings, not _React_ bindings, duh, how would that even work what were you possibly thinking?"

zburke:
WRT this particular pull request, where have we landed? 

* this change looks wrong-headed to me.
* I kind of like what @zburke did here

Round up some relevant colleagues and cast your votes please! 👍 👎

MikeTaylor:
It's option number one for me, Clive.

zburke:
I feel like the right option may be to replace the call to `/expensive-endpoint/we/might/not/have/permissions/for` with a call to something like `/authn/is-authenticated`, i.e. an endpoint that (a) requires authentication and (b) all users have access to. Is [`/bl-users/_self`](https://github.com/folio-org/mod-users-bl/blob/master/descriptors/ModuleDescriptor-template.json#L21-L35) that endpoint? 

@kurtnordstrom, it looks like you [added this feature](https://github.com/folio-org/mod-users-bl/commit/1dc2ed3f5316ffd752821eb1d653d6c43a9030f3), but I haven't been able to find documentation about this endpoint should be used, and I can't figure out which PR (and hence which Jira story) it's attached to.

MikeTaylor:
That looks like a good solution!

zburke:
As Mike as noted, parsing the token isn't the same as validating it.

PR #930, opened 2020-10-06
STCOR-466: Refactor CreateResetPassword to use vanilla react-final-form

yelyzavetakhokhlova:
[[STCOR-466]](https://issues.folio.org/browse/STCOR-466)

## Purpose

We inadvertently created a circular dependency between stripes-core and stripes-final-form in PR #910 for STCOR-441. This prevents the release of stripes-core v6.0.0 and stripes v5.0.0.


zburke (approved):
Thank you!

PR #931, opened 2020-10-07
STCOR-452 validate token without permissions

zburke:
The `validateUser()` function actually does three things:

* check if the given token is syntactically valid
* check whether Okapi is up
* check if the token can be used to make a request

It does this in one fell swoop by making a request, but in order for
this to function for _all_ users the endpoint it hits must be one that
does not require any permissions. `/users` is not such an endpoint, but
`/bl-users/_self` is.

Replaces #923, where I attempted to resolve the permissions problem by
checking for token validity but, ahem, also ignoring the other two tasks
this function accomplishes.

Refs [STCOM-452](https://issues.folio.org/browse/STCOR-452)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Change history for stripes-core
 
+## 6.1.0 IN PROGRESS
+
+* Validate token using a request that does not require permissions. Refs STCOR-452.
+
 ## [6.0.0](https://github.com/folio-org/stripes-core/tree/v6.0.0) (2020-10-06)
 [Full Changelog](https://github.com/folio-org/stripes-core/compare/v5.0.2...v6.0.0)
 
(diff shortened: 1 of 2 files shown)


MikeTaylor (approved):
Much better!

My only reservation is that it makes Stripes dependent on rather a high-level module being present. I know than in practice we're not likely to run a whole lot of FOLIO instances without mod-users-bl, but is there not something we can ask of Okapi itself that would suffice?

zburke:
It's a legit observation, @MikeTaylor, that maybe this should be a request directly to okapi itself rather than to a module okapi is routing requests for. Do you know how I can learn about such an endpoint? The [API documentation](https://dev.folio.org/reference/api/) doesn't list any of Okapi's endpoints 😕 

That said, given that `mod-users-bl` is what provides the login endpoint, I figured this was an acceptable tradeoff, i.e. if `mod-users-bl` isn't present, you can't authenticate in the first place, so you can't get a token, so this is all moot 😉

zburke:
Also, `stripes-core` has depended upon the `users-bl` Okapi interface since it [was possible to depend on okapi interfaces](https://github.com/folio-org/stripes-core/commit/b86e0c92ece103d95f950f24d18053c842dc7077) so this isn't changing anything WRT dependencies.

MikeTaylor:
Does [this](https://dev.folio.org/reference/api/#okapi) not give you the API documentation you need?

Your arguments that mod-users-bl will always be present are compelling, and I do agree that in practical terms there is no real objection to this (hence my review saying Accept). It just makes me a bit itchy.

But, yeah, let's just merge and get this done.

zburke:
Stupidly, I was searching the dev site for `proxy` rather than `okapi`. Let us never speak of this again.

mkuklis (approved):
Nice and clean @zburke! 👏 also thank you @MikeTaylor for all your input on this.

PR #932, opened 2020-10-08
STCOR-451 Settings > Software version: Remove references to color

uladislausamets:
## Purpose

https://issues.folio.org/browse/STCOR-451

## Pre-Merge Checklist
Before merging this PR, please go through the following list and take appropriate actions.
- [x] I've added appropriate record to the CHANGELOG.md
- Does this PR meet or exceed the expected quality standards?
  - [] Code coverage on new code is 80% or greater
  - [] Duplications on new code is 3% or less
  - [] There are no major code smells or security issues
- Does this introduce breaking changes?
  - [ ] If any API-related changes - okapi interfaces and permissions are reviewed/changed correspondingly
  - [x] There are no breaking changes in this PR.

If there are breaking changes, please **STOP** and consider the following:

- What other modules will these changes impact?
- Do JIRAs exist to update the impacted modules?
  - [ ] If not, please create them
  - [ ] Do they contain the appropriate level of detail?  Which endpoints/schemas changed, etc.
  - [ ] Do they have all they appropriate links to blocked/related issues?
- Are the JIRAs under active development?
  - [ ] If not, contact the project's PO and make sure they're aware of the urgency.
- Do PRs exist for these changes?
  - [ ] If so, have they been approved?

Ideally all of the PRs involved in breaking changes would be merged in the same day to avoid breaking the folio-testing environment.  Communication is paramount if that is to be achieved, especially as the number of intermodule and inter-team dependencies increase.

While it's helpful for reviewers to help identify potential problems, ensuring that it's safe to merge is ultimately the responsibility of the PR assignee.


src/components/About/About.js
```diff
@@ -97,6 +90,72 @@ const About = (props) => {
     );
   }
 
+  function renderWarningBanner(interfacesList) {
+    const modulesArray = Object.keys(props.modules).map(key => props.modules[key]);
+    const missingModules = _.flattenDeep(modulesArray.map(module => {
+      return module.map(item => {
+        const okapiInterfaces = item.okapiInterfaces;
+        return okapiInterfaces && Object.keys(okapiInterfaces).map(key => {
```
NikitaSedyx, line 98: map inside map inside map wrapped with flatten

booM!

please decrease code complexity

src/components/About/About.js
```diff
@@ -97,6 +90,72 @@ const About = (props) => {
     );
   }
 
+  function renderWarningBanner(interfacesList) {
```
NikitaSedyx, line 93: additionally please make this as a separate component

src/components/About/About.js
```diff
@@ -97,6 +90,72 @@ const About = (props) => {
...
+          if (!available) {
+            return `${key} ${required}`;
+          } else {
+            return null;
+          }
+        });
+      });
+    }));
+
+    const incompatibleModule = _.flattenDeep(modulesArray.map(module => {
```
NikitaSedyx, line 111: and seems all these maps are copy paste of missing modules, only difference is condition

uladislausamets:
@NikitaSedyx @rasmuswoelk @doytch review this one, please
CI checks fail might be fixed after https://github.com/terser/terser/issues/858 resolving

src/components/About/WarningBanner.js
```diff
@@ -0,0 +1,79 @@
...
+  const [missingModules, setMissingModules] = useState([]);
+  const [incompatibleModule, setIncompatibleModule] = useState([]);
+
+  useEffect(() => {
+    const modulesArray = _.flatten(_.values(modules));
+
+    modulesArray.forEach(module => {
+      const okapiInterfaces = module.okapiInterfaces;
+      if (okapiInterfaces) {
+        Object.keys(okapiInterfaces).forEach(key => {
```
NikitaSedyx, line 28: I suggest to use reduce to build array of missing and another reduce to build incompatible arrays - and move common logic to separate function

src/components/About/WarningBanner.js
```diff
@@ -0,0 +1,79 @@
...
+  interfaces,
+  modules,
+}) => {
+  const [missingModules, setMissingModules] = useState([]);
+  const [incompatibleModule, setIncompatibleModule] = useState([]);
+
+  useEffect(() => {
+    const modulesArray = _.flatten(_.values(modules));
+
+    modulesArray.forEach(module => {
```
NikitaSedyx, line 25: why not to build an array of all okapi interfaces?

the idea to find all missed etc interfaces and set them only once, instead of calling set function n*x times

src/components/About/WarningBanner.js
```diff
@@ -0,0 +1,75 @@
...
+  }, []);
+
+  const missingModulesMsg = <FormattedMessage id="stripes-core.about.missingModuleCount" values={{ count: _.compact(missingModules).length }} />;
+  const incompatibleModuleMsg = <FormattedMessage id="stripes-core.about.incompatibleModuleCount" values={{ count: _.compact(incompatibleModule).length }} />;
+
+  return (
+    <div className={css.warningContainer}>
+      <MessageBanner
+        type="warning"
+        show={_.compact(missingModules).length}
```
aliaksei-chumakou, line 45: `_.compact(missingModules).length` copy-paste

src/components/About/WarningBanner.js
```diff
@@ -0,0 +1,75 @@
...
+        <Headline>{missingModulesMsg}</Headline>
+        <List
+          listStyle="bullets"
+          items={missingModules}
+        />
+      </MessageBanner>
+
+      <MessageBanner
+        type="warning"
+        show={_.compact(incompatibleModule).length}
```
aliaksei-chumakou, line 57: the same

NikitaSedyx (approved):

src/components/About/WarningBanner.js
```diff
@@ -0,0 +1,77 @@
...
+  modules,
+}) => {
+  const modulesArray = _.flatten(_.values(modules));
+  const allInterfaces = modulesArray.reduce((prev, curr) => Object.assign(prev, curr.okapiInterfaces), {});
+
+  const missingModules = Object.keys(allInterfaces).reduce((prev, curr) => {
+    if (!interfaces[curr]) {
+      return [...prev, `${curr} ${allInterfaces[curr]}`];
+    } else {
+      return null;
```
aliaksei-chumakou, line 26: does it work right? From my perspective callback in reducer always returns accumulator, and I'n not sure that returning null would work here https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/reduce

uladislausamets:
@doytch @rasmuswoelk review this one, please

src/components/About/WarningBanner.js
uladislausamets: You are right, it's isn't work in some cases.

src/components/About/WarningBanner.js
```diff
@@ -0,0 +1,77 @@
...
+  const incompatibleModule = Object.keys(allInterfaces).reduce((prev, curr) => {
+    if (interfaces[curr] && !isVersionCompatible(interfaces[curr], allInterfaces[curr])) {
+      return [...prev, `${curr} ${allInterfaces[curr]}`];
+    } else {
+      return prev;
+    }
+  }, []);
+
+  const missingModulesCount = _.compact(missingModules).length;
+  const incompatibleModulesCount = _.compact(incompatibleModule).length;
```
mkuklis, line 39: @uladislausamets there is a lot going on here. This code will execute on every rerender. It would be nice to see if wrapping it in `useMemo` produces a better performance. I wonder if you tried to test it?

src/components/About/WarningBanner.js
```diff
@@ -0,0 +1,86 @@
...
+        return [...prev, `${curr} ${allInterfaces[curr]}`];
+      } else {
+        return prev;
+      }
+    }, []);
+  }, [allInterfaces, interfaces]);
+
+  const missingModulesCount = useMemo(() => _.compact(missingModules).length, [missingModules]);
+
+  const incompatibleModulesCount = useMemo(() => _.compact(incompatibleModule).length, [incompatibleModule]);
```
axelhunn, line 47: Why _.compact the arrays, aren't they supposed to be compact already?

src/components/About/WarningBanner.js
```diff
@@ -0,0 +1,86 @@
...
+    return Object.keys(allInterfaces).reduce((prev, curr) => {
+      if (!interfaces[curr]) {
+        return [...prev, `${curr} ${allInterfaces[curr]}`];
+      } else {
+        return prev;
+      }
+    }, []);
+  }, [allInterfaces, interfaces]);
+
+  const incompatibleModule = useMemo(() => {
```
axelhunn, line 35:
```suggestion
  const incompatibleModules = useMemo(() => {
```

src/components/About/WarningBanner.js
```diff
@@ -0,0 +1,86 @@
...
+
+  const missingModules = useMemo(() => {
+    return Object.keys(allInterfaces).reduce((prev, curr) => {
+      if (!interfaces[curr]) {
+        return [...prev, `${curr} ${allInterfaces[curr]}`];
+      } else {
+        return prev;
+      }
+    }, []);
+  }, [allInterfaces, interfaces]);
```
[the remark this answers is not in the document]
uladislausamets: @axelhunn  I don't think that these changes will decrease code complexity. IMHO one reduce looks better than map after filter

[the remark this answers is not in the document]
axelhunn: It's more self-explanatory for my taste.
Also, it could be faster - see https://www.measurethat.net/Benchmarks/Show/7249/0/filter-map-vs-reduce

Plus, you could DRY out the mapper for both `missingModules` and `incompatibleModules`

src/components/About/WarningBanner.js
```diff
@@ -0,0 +1,83 @@
...
+
+  const incompatibleModules = useMemo(() => {
+    return Object.keys(allInterfaces).reduce((prev, curr) => {
+      if (interfaces[curr] && !isVersionCompatible(interfaces[curr], allInterfaces[curr])) {
+        return [...prev, `${curr} ${allInterfaces[curr]}`];
+      } else {
+        return prev;
+      }
+    }, []);
+  }, [allInterfaces, interfaces]);
```
axelhunn, line 40: filter-map too? )

src/components/About/WarningBanner.js
```diff
@@ -0,0 +1,83 @@
...
+
+  const missingModulesCount = useMemo(() => missingModules.length, [missingModules]);
+
+  const incompatibleModulesCount = useMemo(() => incompatibleModules.length, [incompatibleModules]);
```
axelhunn, line 44: I think now, without `compact`, useMemo() is excessive here. No need to memoize just an array length.

axelhunn (approved):

src/components/About/WarningBanner.js
```diff
@@ -0,0 +1,78 @@
...
+
+const WarningBanner = ({
+  interfaces,
+  modules,
+}) => {
+  const allInterfaces = useMemo(() => {
+    const modulesArray = _.flatten(_.values(modules));
+
+    return modulesArray.reduce((prev, curr) => Object.assign(prev, curr.okapiInterfaces), {});
+  }, [modules]);
```
mkuklis, line 23: @uladislausamets thank you for trying out `useMemo`. Unfortunately this still won't work as expected because `useMemo` is not doing deep compare under the hood. If you put `console.log` right inside this `useMemo` you will notice that it's still being executed on every rerender. So this is actually worst that what we had before.

One way to solve it would be to use something similar to:

https://github.com/sandiiarov/use-deep-compare#usedeepcomparememo

another would be to do: `JSON.stringify(modules)` which sometimes beats the purpose of caching (because it's actually slower).

I think would introduce this new dependency to stripes core:

https://github.com/sandiiarov/use-deep-compare

and try to use `import { useDeepCompareMemo } from "use-deep-compare";`

I'm sure it will be helpful to have it around in other places.

uladislausamets: `useDeepCompareMemo` helped to decrease the count of functions executions

aliaksei-chumakou: useDeepCompareMemo looks like a hack to me, is there a way when props aren't updated on every render?

mkuklis: @aliaksei-chumakou another approach here could be to use `React.memo` on the entire component. I think deep comparing (maybe with https://lodash.com/docs/4.17.15#isEqual) would have to be used there too.

mkuklis: but it would clean up this code a bit because `useMemo` would not have to be used. Maybe that's the way to do it.

mkuklis: @uladislausamets sorry for raising these questions originally but I think it was a good exercise to see different approaches with these kind of issues.

mkuklis (approved):

src/components/About/WarningBanner.js
uladislausamets: I also tried `React.memo` and it isn't so effective as `useDeepCompareMemo`

PR #935, opened 2020-10-15
UIEH-976: Defined a mock for '/bl-users/_self' request

BogdanDenis:
## Description
eHoldings experiences test failures due to an update in stripes-core. These failures are caused by a PR that changed endpoint that validates a token (https://github.com/folio-org/stripes-core/pull/931).
This PR adds a new mock endpoint `bl-users/_self` so that the token validator receives a correct response

### DIFF
--- a/test/bigtest/network/config.js
+++ b/test/bigtest/network/config.js
@@ -18,7 +18,6 @@ export default function configure() {
     configs: []
   });
 
-  this.get('/users', {});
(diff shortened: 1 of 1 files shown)


zburke (changes requested):
Might as well remove the `/users` line since it is no longer used within the app.

zburke (approved):
Thank you! :shipit:

PR #944, opened 2020-11-11
Perms package2translation

MikeTaylor:


src/components/About/About.js
```diff
@@ -2,6 +2,7 @@ import _ from 'lodash';
 import React from 'react';
 import PropTypes from 'prop-types';
 import { FormattedMessage } from 'react-intl';
+import stripesPlatform from '@folio/stripes';
```
zburke, line 5: `@folio/stripes-core` sits below `@folio/stripes` in the hierarchy (i.e. `stripes` depends on `stripes-core`, not the other way around), so you can't import it here. This is one of my arguments that stripes-core the framework should be separated from stripes-core the app: if this page were in an app consuming the framework it would work fine. 

I feel like I must have filed a ticket for this but I can't find it.

zburke (changes requested):

zburke (changes requested):
If you separate the very-useful-looking `perms-package2translation` script from the broken/unrelated `About` changes, we could merge this so other folks can benefit from your helpful script.

MikeTaylor:
I have no idea how these two things got entangled in the first place.

PR #946, opened 2020-11-16
STCOR-481: Move CalloutContext to Root to avoid issues with React Intl

mkuklis:
https://issues.folio.org/browse/STCOR-481

It looks like every time we change the locale the `IntlProvider` is changed causing the `RootWithIntl` to be completely recreated (a new RootWithIntl instance is being created). This also recreates the callout ref and `CalloutContext.Provider`  causing issues similar to the one described in STCOR-481

This PR moves the context and callout to `Root`. This seems to help at least with the issue described in STCOR-481.

### DIFF
--- a/src/components/Root/Root.js
+++ b/src/components/Root/Root.js
@@ -7,6 +7,7 @@ import { IntlProvider } from 'react-intl';
 import queryString from 'query-string';
 import { ApolloProvider } from '@apollo/client';
 import ErrorBoundary from '@folio/stripes-components/lib/ErrorBoundary';
+import { Callout } from '@folio/stripes-components';
 import { metadata, icons } from 'stripes-config';
 
 /* ConnectContext - formerly known as RootContext, now comes from stripes-connect, so stripes-connect
@@ -23,6 +24,7 @@ import { getQueryResourceKey, getCurrentModule } from '../../locationService';
 import Stripes from '../../Stripes';
 import RootWithIntl from '../../RootWithIntl';
 import SystemSkeleton from '../SystemSkeleton';
+import CalloutContext from '../../CalloutContext';
 
 import './Root.css';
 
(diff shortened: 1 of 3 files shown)


zburke (approved):

src/components/Root/Root.js
```diff
@@ -130,22 +133,25 @@ class Root extends Component {
...
-            </IntlProvider>
+            <CalloutContext.Provider value={this.callout.current}>
+              <IntlProvider
+                locale={locale}
+                key={locale}
+                timeZone={timezone}
+                currency={currency}
+                messages={translations}
+                textComponent={Fragment}
+                onError={config?.suppressIntlErrors ? () => {} : undefined}
```
aliaksei-chumakou, line 144: isn't this `() => {}` treated as a new prop every render?

aliaksei-chumakou:
@mkuklis So it's intentionally, that you can't pass <FormattedMessage /> as `message` in this CalloutContext?

ViktorSoroka07:
I see that this PR causes errors in the whole FOLIO app when it comes to showing the callout.

aliaksei-chumakou:
Basically, @ViktorSoroka07  you're right - the main idea is to use your own callout context per module. I just want to be sure if this is the only way before changing all 13 modules to this...

mkuklis:
@aliaksei-chumakou and @ViktorSoroka07 thank you for looking at this. The change here was pretty minimal (moving the context to `Root` instead of `RootWithIntl`. I think we need more time to understand why we see these issues. I'm happy to revert it but let's spend some time understand why this is happening.

ViktorSoroka07:
Yeah, @mkuklis sounds good. I tried to handle it myself. I tried to move the Overlay container as a sibling to root as I thought that the might help because the overlay container is going to be rerendered but unfortunately it also did not help.

PR #947, opened 2020-11-17
STCOR-483 avoid render-props with FormattedMessage

zburke:
Using render-props with a `<FormattedMessage>` in `react-intl` `v5`
renders as an array rather than as a string as it did in `v4`. This
causes problems for attributes such as `aria-label`, which expect
strings.

Refs [STCOR-482](https://issues.folio.org/browse/STCOR-482), [STRIPES-694](https://issues.folio.org/browse/STRIPES-694)


rasmuswoelk (approved):

ryandberger (approved):
I think the syntax reads a bit better too. Nice work!

mkuklis (approved):

PR #948, opened 2020-11-17
STCOR-483 use correct HTTP responses codes

zburke:
The [mirage documentation
notwithstanding](https://miragejs.com/docs/main-concepts/route-handlers/#status-codes-and-headers),
to actually return a custom response code you must return a
custom `Response` object with the desired code _and_ include that same
code as the third parameter to the action (`put`, `post`, etc) method.
Only doing the first (as in the documentation) or the later (as in
previous code) doesn't work correctly.

With only the former, mirage is happy but the code under test receives a
201 instead of the custom code, causing tests of code that expect
specific response codes to fail.

With only the latter, mirage spews the warning but the code under test
receives the expected response.

Refs [STCOR-483](https://issues.folio.org/browse/STCOR-483)


mkuklis (approved):
Nice cleanup!

ryandberger (approved):

PR #949, opened 2020-11-18
fix tests - focus button before assertions

JohnC-80:
In a recent change to stripes-components' `<Dropdown>`, the focus moves into the menu only when the dropdown's trigger is in focus when the menu opens... this action isn't quite fast enough for the assertions and `click` alone doesn't appear to be setting a focus to the dropdown trigger.


mkuklis:
@JohnC-80 thank you for fixing these!

PR #952, opened 2020-11-20
STCOR-481: Move CalloutContext Provider back under IntlProvider

mkuklis:
After some refactoring related to STCOR-481 we started to see some issues reported by @ViktorSoroka07 and @aliaksei-chumakou related to missing `intl` when using `FormattedMessage` with the `callout`.

https://github.com/folio-org/stripes-core/pull/946#issuecomment-731106017

@zburke found a good solution for it under:

https://github.com/folio-org/stripes-smart-components/pull/962

This PR is moving the `CalloutContext.Provider` back under `IntlProvider` to address similar issues in other places.

The callout ref is kept in the `Root` so the issue reported in STCOR-481 should still work.


ViktorSoroka07 (approved):

aliaksei-chumakou (approved):
🔥

PR #955, opened 2020-11-23
New utility script util/perms-package2translation.js

MikeTaylor:


MikeTaylor:
As usual, trying to get this trivial script committed into the master branch has taken me an order of magnitude longer than it took my to create it in the first place. I won't bother next time, unless I am confident that it would take the next person to use it ten times as long to write as it did for me.

PR #961, opened 2020-12-15
Add support for generic package scope names

ryandberger:
- Fulfills [STCOR-490](https://issues.folio.org/browse/STCOR-490).
- Remove hard-coded requirement that modules must have prefix of `@folio/` scope and allow for other scope names.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 * Add support for building and consuming Webpack DLLs. Refs STCOR-471.
 * Provide default HTML formatters to `<IntlProvider>` so we can avoid `<SafeHTMLMessage>`. Fixes STCOR-477.
 * Settings > Software version > Display a loading indicator when querying for missing/incompatible modules, STCOR-479.
+* Add support for generic package scope names (instead of assuming @folio/something). Refs STCOR-490.
 
 ## [6.0.0](https://github.com/folio-org/stripes-core/tree/v6.0.0) (2020-10-06)
 [Full Changelog](https://github.com/folio-org/stripes-core/compare/v5.0.2...v6.0.0)
--- a/src/components/AppIcon/AppIcon.js
+++ b/src/components/AppIcon/AppIcon.js
@@ -39,8 +39,18 @@ const AppIcon = ({
      * If no icon is found we display a placeholder.
      *
      */
-    let appIcon = stripes?.icons?.[app]?.[iconKey];
-    if (!appIcon) appIcon = stripes?.icons?.[`@folio/${app}`]?.[iconKey];
+    const icons = stripes?.icons;
+    // First, see if the app name matches the package name exactly (no scope prefix).
+    let appIcon = icons?.[app]?.[iconKey];
+    if (!appIcon) {
+      for (const key in icons) {
+        // Otherwise, match the package name based on the name after the scope prefix.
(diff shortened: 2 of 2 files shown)


skomorokh:
I think the better solution would be to allow the scope to be included. And it may already be allowed? @folio is supported for backwards compatibility but we should be explicitly including the scope so that, in a contrived example, if another vendor makes an alternative users app @vendor/users for use most of the time but want to still enable @folio/users for some few admin users and some of the settings functionality they don't have a collision. eg. we should respect these namespaces and I think it's best not to futher entrench omitting them.

skomorokh:
eg. I think we should be calling `<AppIcon app="@scope/name" />` anyhow, including for `@folio` and gradually attempt to move away from implicit `@folio` anywhere.

ryandberger:
> "@scope/name

Yep, that's probably the best way to go. There's about 4-ish different entry points, so getting a handle on the full package name in those spots should be fun 😉

mkuklis (approved):

zburke:
I'm inclined to agree with @skomorokh here, though your point that un-namespaced use is already widely in use, and therefore complicates this, is duly noted. 

Can/should we delay this work until a next-major-version when we could change the API? (Alternatively, we merge this now, but also plan a next-major-version change that simplifies the implementation and throws out the namespace-ignoring portions?)

ryandberger:
> I'm inclined to agree with @skomorokh here, though your point that un-namespaced use is already widely in use, and therefore complicates this, is duly noted.
> 
> Can/should we delay this work until a next-major-version when we could change the API? (Alternatively, we merge this now, but also plan a next-major-version change that simplifies the implementation and throws out the namespace-ignoring portions?)

I'd prefer to have this done sooner. For those modules that don't have the `@folio/` scope, currently their icon doesn't appear in the UI and defaults to the black square.

mkuklis:
I think we should merge this because it's better than what we have now. We can always improve it. Small incremental steps are good :)

skomorokh:
This is a small incremental step in the wrong direction. Why not include the scope where stripes-core is currently failing to pass it through to AppIcon?

zburke:
Agreed again, @skomorokh, but I understand @ryandberger's point that folks writing non-`@folio` apps get screwed here. As above, how about we merge this now to get the bandaid fix in place, but also plan a next-major-version change that removes the namespace shenanigans altogether?

ryandberger:
> Agreed again, @skomorokh, but I understand @ryandberger's point that folks writing non-`@folio` apps get screwed here. As above, how about we merge this now to get the bandaid fix in place, but also plan a next-major-version change that removes the namespace shenanigans altogether?

I'm agreement on that. I'm going to proceed with the merge and plan on having the full module get passed in as a separate PR. The advantage there is that this PR both solves the issue, but then provides a better fallback in the future since the logic first checks for an exact match, but then allows for a generic scope rather than the hard-coded `@folio/` scope.

ryandberger:
Opened a separate PR with Jason's suggestion: https://github.com/folio-org/stripes-core/pull/964.

Haven't been able to fully vet that solution yet b/c workspaces are giving me trouble at the moment.

ryandberger:
Since I have verified correct behavior in https://github.com/folio-org/stripes-core/pull/964, I am proceeding with that approach and closing this PR.

PR #964, opened 2020-12-18
Stcor 490 full module name

ryandberger:
- Fulfills [STCOR-490](https://issues.folio.org/browse/STCOR-490).
- Passing in full module name to resolve icon so that `@folio/` scope prefix is not assumed.
- Tested with a non-folio-scoped module and confirmed that icon appears properly in menu bar, app drop-down list, current app in top left, and settings. Also verified folio-scoped app icons continue to work as expected. 
- NOTE: Icons within non-folio-scoped module page must make this same change in their codebase for the icon to display, but this handles the cases managed by `stripes-core`.


mkuklis (approved):

zburke:
Nice work-around!

PR #967, opened 2020-12-21
STCOR-496 conduct discovery after login

zburke:
Backport changes from STCOR-454 (PR #893, #901) to `v5` in order to
allow operators in a multi-tenant environment to update their okapi to
v4. This means conducting module-discovery _after_ login and
authenticated as the current tenant instead of `supertenant`.

In short, this change allows `@folio/stripes` `v4`
(`@folio/stripes-core` `v5`) to be compatible with `okapi` `v4`;
previously, it was compatible only with `v3`.

Refs [STCOR-496](https://issues.folio.org/browse/STCOR-496), [STCOR-454](https://issues.folio.org/browse/STCOR-454)

### DIFF
--- a/src/discoverServices.js
+++ b/src/discoverServices.js
@@ -1,6 +1,13 @@
 import 'isomorphic-fetch';
 import { some } from 'lodash';
 
+function getHeaders(tenant, token) {
+  return {
+    'X-Okapi-Tenant': tenant,
+    'X-Okapi-Token': token,
+    'Content-Type': 'application/json'
+  };
+}
 /*
  * This function probes Okapi to discover what versions of what
  * interfaces are supported by the services that it is proxying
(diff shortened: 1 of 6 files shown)


src/discoverServices.js
```diff
@@ -1,6 +1,13 @@
 import 'isomorphic-fetch';
 import { some } from 'lodash';
 
+function getHeaders(tenant, token) {
```
mkuklis, line 4: Nice cleanup :)

mkuklis (approved):

wafschneider (approved):
I tested this patch in our affected environments and it acts as expected: calls to `/_/proxy/tenant/{tenantId}/modules` happen after login and using the appropriate `x-okapi-tenant` header. It appears to resolve the issues with running Okapi 4.X in a Goldenrod environment.

PR #968, opened 2020-12-21
STCOR-455 webpack plugin for error logging config

zburke:
Read config values for error logging from the `stripes.config.js`
property `errorLogging`. Include the default keys `bugsnag`, `sentry`,
and `rollbar` with null values. The `errorLogging` object will be
present in `stripes-config` just as `branding` is, i.e. it can be
imported as
```
import { errorLogging } from `stripes-config`.
```

Refs STCOR-455

### DIFF
--- a/webpack/stripes-error-logging-plugin.js
+++ b/webpack/stripes-error-logging-plugin.js
@@ -1,0 +1,27 @@
+// This webpack plugin generates a virtual module containing the
+// error-handler configuration.
+
+const logger = require('./logger')('stripesErrorLoggingPlugin');
+
+module.exports = class StripesErrorLoggingPlugin {
+  constructor(options) {
+    logger.log('initializing...');
+
+    const defaultErrorLogging = {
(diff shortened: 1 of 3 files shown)


zburke:
As with its [stripes-components counterpoint|https://github.com/folio-org/stripes-components/pull/1392], this should be closed. I only opened the PR in order to link the POC to the ticket.

webpack/stripes-error-logging-plugin.js
```diff
@@ -0,0 +1,27 @@
...
+
+module.exports = class StripesErrorLoggingPlugin {
+  constructor(options) {
+    logger.log('initializing...');
+
+    const defaultErrorLogging = {
+      rollbar: null,
+      bugsnag: null,
+      sentry: null,
+    };
```
zburke, line 14: You know, aside from this assignment, which should just be `const defaultErrorLogging = {}`, this PR wasn't all bad!

PR #982, opened 2021-01-05
STCOR-497: Node.js TLS, HTTP and OpenSSL security vulnerabilities

julianladisch:
Node is used to compile the Stripes based UI.
If Node fetches dependencies (external packages) it may use TLS, HTTP and/or OpenSSL
and might be affected by this issue. Then the build crashes or the resulting
FOLIO runtime might contain malware.

https://nodejs.org/en/blog/vulnerability/january-2021-security-releases/ explains the vulnerabilities:

* use-after-free in TLSWrap (High) https://nvd.nist.gov/vuln/detail/CVE-2020-8265
* HTTP Request Smuggling in nodejs https://nvd.nist.gov/vuln/detail/CVE-2020-8287
* OpenSSL - EDIPARTYNAME NULL pointer de-reference https://nvd.nist.gov/vuln/detail/CVE-2020-1971

Zak Burke wrote in #stripes-architecture Slack channel on 2020-12-17:
"Additionally, as already noted in https://issues.folio.org/browse/FOLIO-2553,
most packages de facto require node v12 already due to dependencies which correctly state a
v12 minimum.
IOW, yes, we should update the minimum required version we state on our package"

Bumping the node version to the latest patch version of the LTS release lines will
fix the issues CVE-2020-8265, CVE-2020-8287, and CVE-2020-1971.

Refs [STCOR-497](https://issues.folio.org/browse/STCOR-497)

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
     "stylelint": "stylelint \"src/**/*.css\""
   },
   "engines": {
-    "node": "^15.5.1 || ^14.15.4 || ^12.20.1"
+    "node": ">=15.5.1 || ^14.15.4 || ^12.20.1"
   },
   "stripes": {
     "okapiInterfaces": {
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
     "stylelint": "stylelint \"src/**/*.css\""
   },
   "engines": {
-    "node": ">=10.0.0"
+    "node": "^15.5.1 || ^14.15.4 || ^12.20.1"
   },
   "stripes": {
     "okapiInterfaces": {


zburke:
It appears this relies on https://github.com/folio-org/folio-tools/pull/140 since CI is only running node `v12.20.0` and this PR bumps the minimum version to `v12.20.1`.

zburke (approved):

mkuklis (approved):

skomorokh (changes requested):
While I in no way question this is an issue and totally agree we should give our devops teams a heads-up, I don't think this is what that part of the package.json is for. That just indicates which versions of node Stripes is compatible with, it's not what pulls node in. That's up to who maintains the system running it. I think it's telling that we're not merging this until we get CI running on the patched version of node. So why not extend that courtesy to our partners rather than having Stripes potentially refuse to run on their unpatched systems?

It's not like we check that they're invoking us from a fully patched version of bash or on a fully up to date OS.

Also, it would mean that we then need to explicitly add every subsequent version of node to this || list, adding a maintenance burden with no clear benefit.

I think this situation is quite different than dependencies as those we are responsible for since package.json does determine which versions of those things are pulled in and is opaque to the system integrators.

julianladisch:
@skomorokh wrote:

> add every subsequent version of node to this || list

Thank you for pointing out this omission. I've changed "^15.5.1" to ">=15.5.1". We can safely assume that all future major Node versions are not affected by the current three vulnerabilities.

julianladisch:
> running on the patched version of node. So why not extend that courtesy to our partners rather than having Stripes potentially refuse to run on their unpatched systems?

1. Adding this Node version restriction to stripes-core requires to change a single file only, and it will fail the Stripes build on ALL installations. Adding this Node version restriction to the DevOp and SysOp scripts requires changes in all these scripts, requires much more work, and is error-prone. Some scripts are on https://github.com/search?q=org%3Afolio-org+apt-get+nodejs&type=Code but some are local.
2. `"node": ">=15.5.1 || ^14.15.4 || ^12.20.1"` is an easy way to set the version restriction for all major versions (15, 14, 13, 12, 11, ...). It is much more effort to accomplish the same in `apt-get`, `ansible` and other DevOp and SysOp tools, and to write documentation for DevOps and SysOps explaining what to do.
3. Security enforcement is a legal requirement for organisations in the EU. "GDPR requires you to put in place appropriate technical and organisational measures to implement the data protection principles effectively and safeguard individual rights. This is ‘data protection [by design and by default](https://ico.org.uk/for-organisations/guide-to-data-protection/guide-to-the-general-data-protection-regulation-gdpr/accountability-and-governance/data-protection-by-design-and-default/)’." "process personal data securely by means of ‘appropriate technical and organisational measures’ – this is the ‘[security principle](https://ico.org.uk/for-organisations/guide-to-data-protection/guide-to-the-general-data-protection-regulation-gdpr/security/)’." "[Who is responsible for complying with data protection by design and by default? (...) your software engineers, system architects and application developers](https://ico.org.uk/for-organisations/guide-to-data-protection/guide-to-the-general-data-protection-regulation-gdpr/accountability-and-governance/data-protection-by-design-and-default/#dpd5)"
4. No use case has been shown where FOLIO needs to ensure that Stripes can build on an unpatched system. Even if there were such a use case stripes-core can easily been forked and the Node version requirement can be changed locally without affecting the security of others.
5. This pull request currently fails because jenkins-slave-all runs vulnerable Node v12.20.0; this will be fixed by https://github.com/folio-org/folio-tools/pull/140. This failure shows that the change to stripes-core is needed and useful.

I understand that valid point of the Stripes developers that they don't want to take care for DevOps and SysOps that fail to use a fully-patched system. I will ask the Security Team to discuss this pull request and SCOR-497 and to make a suggestion to the Technical Council to decide (close as "Won't do" or commit).

PR #983, opened 2021-01-06
UX-419: Added FOLIO icon for <CurrentApp> on the home page

rasmuswoelk:
## Ticket
https://issues.folio.org/browse/UX-419

## Changes
Added FOLIO icon as a fallback when no app is selected.

## Screenshot
image


MikeTaylor (approved):
Changes look good. I have a couple of other concerns but I will mention one of them in the Jira issue and the other in a separate comment on this PR.

MikeTaylor:
My other comment: stripes-core has got very noisy under ESlint. I see 33 errors above, which is really not tolerable. It's easy to hide a leaf in a forest, and it's easy to miss a real error in screeds of output about things that "we'll worry about later". A lint-cleanup pass is badly needed. @zburke what is the procedure for getting such things to happen these days?

PR #984, opened 2021-01-08
STCOR-500: Avoid retrying of http requests using `useOkapiKy` hook by default

ViktorSoroka07:
## Purpose

Avoid retrying of HTTP requests using `useOkapiKy` hook by default in the scope of the [STCOR-500](https://issues.folio.org/browse/STCOR-500).


skomorokh (approved):
Good catch!

ViktorSoroka07:
Thanks. I actually found it when I tried `react-query` which also offers retrying of requests, so I wondered why it is doing so many requests.

PR #986, opened 2021-01-08
STCOR-536 Upgrade RxJS to v6

mkuklis:
https://issues.folio.org/browse/STCOR-536

There is a corresponding PR in stripes-connect which is tied to this work too:

https://github.com/folio-org/stripes-connect/pull/164

Also right after I upgraded this I realize that eholdings is also using epics:

https://github.com/folio-org/ui-eholdings/tree/master/src/redux/epics

They will have to be upgraded too before we can merge this work.


ViktorSoroka07 (approved):
Looks really nice with imports of operators from `rxjs/operators`.

zburke (commented):
Thanks for doing this important work to keep our dependencies current!

Your comments mention there is also an open [stripes-connect PR](https://github.com/folio-org/stripes-connect/pull/164) that deals with this same issue. If the versions need to be synced across these packages, we should also move `redux-observable` and `rxjs` to peers as part of this work, and supply them via the `@folio/stripes` package. 

Alternatively, if the versions need to be synced across apps as well (I'm looking at your [e-holdings PR](https://github.com/folio-org/ui-eholdings/pull/1233)), then I would argue for pulling these deps all the way up to the platform.

mkuklis:
@zburke moving them to `stripes` level sounds good to me. The problem with the platform level is that we would want to run e-holdings as a separate module (during development) so we may not reach the platform level. Does that make sense?

zburke:
> The problem with the platform level is that we would want to run e-holdings as a separate module (during development) so we may not reach the platform level. 

Thinking through this again, doesn't this just mean that here, `redux-observable` and `rxjs` should be moved to dev-deps (for running stand-alone, sans platform) and peer-deps (for running within a platform)? Alternatively, I would argue they should be peer-deps within `stripes-*`, hard-deps of `stripes` itself, _and_ exported by `stripes` for apps to import. Otherwise, we're just manually aligning all those versions with no assistance from yarn. Obviously, that's been working so far 😅 but it's the most error-prone way possible to handle this.

mkuklis:
Listing them as dev and peer makes sense to me. Good Idea Zak. (it still kind of sucks we have to do that everywhere).

I'm not sure how I feel about importing them (end re-exporting them) via `stripes`. I wonder if perhaps an alternative could be to create a `stripes-deps` package (or something similar) where we keep all those common deps and then just use that as a dependency.  Hmm.

skomorokh:
> I wonder if perhaps an alternative could be to create a `stripes-deps` package (or something similar) where we keep all those common deps and then just use that as a dependency. Hmm.

Unfortunately that won't help as that won't satisfy the peers when the app is installed standalone for CI testing. What would help is to require a platform for testing and that would provide the peers like we do in prod, could even have it in @folio/stripes so it's coordinated and versioned appropriately... something like that. But then you lose testability from a single package... but then we might not need to depend on stripes-cli everywhere too... :thinking: It's tradeoffs all the way down like everything.

doytch:
Or we change everything to instead `import { rxJSThingABob } from '@folio/stripes/rxjs'`. But..... 😬

zburke:
Peers can only be satisfied by looking up in the hierarchy. The only thing above an app is a platform. Thus, there are three options for satisfying deps shared among stripes libs and apps:

1. use peer-deps and dev-deps in `stripes-*` and apps; use a dep in platforms. This avoids re-exporting things, which some folks don't like, but, it means we have to add all the `stripes-*` peer-deps as peer-deps and dev-deps to every. single. UI app in order to be able to build them in isolation or run any tests. 
1. use peer-deps in `stripes-*`; use a dep and re-export in `stripes` proper; have apps import from `stripes` proper. 
1. create a new `stripes-deps` package that sits at the bottom of the hierarchy that contains deps and re-exports; have `stripes-*` and apps import from that. A module just for deps seems kinda lame, but it does mean we get to treat `stripes-*` and `ui-*` exactly the same way WRT shared deps so maybe there's some value there. 

Should we have this conversation in a less hidden corner of the world, either on Slack or in #stripes-architecture?

mkuklis:
That sounds good @zburke. Maybe we can think about what we are trying to solve here.

It would be great if we didn't have to list 3rd party dependencies in every UI module under both peer and dev. Another issue is the maintenance. It would be nice to just have one single entry under each ui-module.

skomorokh:
> Thus, there are three options for satisfying deps shared among stripes libs and apps:

That's a false trichotomy, I'm sure there are more ways to do it. Including what I described above, just include the ones you need as peer and, rather than needing the list of all peers in the dev deps for all modules, contrive a way to run tests with a platform. Probably involving stripes-cli.

> Should we have this conversation in a less hidden corner of the world,

Yes, probably on this week's call?

PR #989, opened 2021-01-14
STCOR-502 Expose module context

skomorokh:
Exposes the module context, enabling apps to consume build-time configuration.


mkuklis (approved):

zburke (approved):
LGTM. Thanks for cleaning up the `ModulesContext` export as well!

PR #990, opened 2021-01-15
STCOR-503 Move react-titled to peer dependency

skomorokh:
This enables the functionality to actually be used by apps as they need to be on the same version.


zburke (approved):
Technically, adding a peer-dep is a breaking change** so it is probably worth mentioning on the CHANGELOG. 

** which is fine; we already bumped the major version for this release. It's just sort of "extra" noteworthy, even though it doesn't seem like the feature is changing, because it does change the requirements to run this version of this module.

zburke (approved):

PR #999, opened 2021-01-22
Revert "STCOR-503 Move react-titled to peer dependency"

zburke:
Reverts folio-org/stripes-core#990

In fact this was exported here but we missed it. Sorry!


skomorokh (approved):
Sorry for the hassle, my bad :(

PR #1001, opened 2021-01-22
STCOR-507 Add prefix prop to TitleManager

skomorokh:


zburke (approved):

doytch (commented):
Looks good and just to nitpick, lint is at some point gonna complain about unsorted props, and I'd rather use `defaultProps` rather than `|| ''`.

skomorokh:
Thanks, that's excellent feedback and defaultProps is the way to go for sure. Though prop sorting is asinine so if we haven't already turned it off (it didn't seem to warn...) we should and leave that to developers to manage, or not, as they see fit. eg. in this case I'd say the order they appear in the title is more intuitive than a lexical ordering ...but I can live with it so I suppose if people are attached to it I'm not going to argue to disrupt that consensus.

PR #1002, opened 2021-01-24
STCOR-508: Setup react-query

ViktorSoroka07:
## Purpose

Following the decision during the stripes-architecture meeting, the purpose of this PR is to setup [react-query](https://react-query.tanstack.com/) in the scope of the [STCOR-508](https://issues.folio.org/browse/STCOR-508).

`react-query` is a library for fetching, caching, and updating data in React applications without touching any "global state". This PR also provides [default options](https://react-query.tanstack.com/guides/important-defaults) so individual applications will not need to override them every time. The library provides handy hooks such as `useQuery`, `useQueries`, `useQueryClient`, etc to work with queries and `useMutation` to work with mutations. This library can be treated as an alternative to `stripes-connect` for developers who want to write the code using hooks for querying/mutating the data and leverage the features of `react-query`.

I also checked that for the `Bigtest` tests no teardown logic is needed. The queries cache comes cleared in each test.

In case the PR is merged I will add documentation to the [stripes](https://github.com/folio-org/stripes) project about recipes and important aspects that should be followed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -144,6 +144,7 @@
     "redux-logger": "^3.0.6",
     "redux-observable": "^0.15.0",
     "redux-thunk": "^2.1.0",
+    "react-query": "^3.6.0",
     "regenerator-runtime": "^0.13.3",
     "rimraf": "^2.5.4",
     "rtl-detect": "^1.0.2",
--- a/src/createReactQueryClient.js
+++ b/src/createReactQueryClient.js
@@ -1,0 +1,16 @@
+import { QueryClient } from 'react-query';
+
+const cacheTimeMinutes = 5;
+
+const createReactQueryClient = () => new QueryClient({
+  // https://react-query.tanstack.com/guides/important-defaults
+  defaultOptions: {
+    queries: {
+      retry: false,
+      refetchOnWindowFocus: false,
+      cacheTime: 1000 * 60 * cacheTimeMinutes,
+    },
+  },
+});
+
+export default createReactQueryClient;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@
 * Avoid retrying of http requests using `useOkapiKy` hook by default. STCOR-500.
 * Expose module context via `useModules`/`withModules`/`withModule`. STCOR-502
 * Add `prefix` prop to TitleManager. STCOR-507
+* Setup `react-query`. Refs STCOR-508
 
 ## [6.0.0](https://github.com/folio-org/stripes-core/tree/v6.0.0) (2020-10-06)
 [Full Changelog](https://github.com/folio-org/stripes-core/compare/v5.0.2...v6.0.0)
--- a/src/components/Root/Root.js
+++ b/src/components/Root/Root.js
@@ -5,7 +5,9 @@ import { connect } from 'react-redux';
 import { createBrowserHistory } from 'history';
 import { IntlProvider } from 'react-intl';
 import queryString from 'query-string';
+import { QueryClientProvider } from 'react-query';
 import { ApolloProvider } from '@apollo/client';
+
 import ErrorBoundary from '@folio/stripes-components/lib/ErrorBoundary';
 import { Callout } from '@folio/stripes-components';
 import { metadata, icons } from 'stripes-config';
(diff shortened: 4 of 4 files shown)


src/createReactQueryClient.js
```diff
@@ -0,0 +1,16 @@
...
+
+const cacheTimeMinutes = 5;
+
+const createReactQueryClient = () => new QueryClient({
+  // https://react-query.tanstack.com/guides/important-defaults
+  defaultOptions: {
+    queries: {
+      retry: false,
+      refetchOnWindowFocus: false,
+      cacheTime: 1000 * 60 * cacheTimeMinutes,
```
ViktorSoroka07, line 11: Five seconds is the default value for the `cacheTime` option and I added it here for now to decide the value for the FOLIO. [caching article](https://react-query.tanstack.com/guides/caching) helps to understand what is the idea of the option. I think the `staleTime` option which is mentioned in the article should not be updated so I did not add it in the PR.

src/createReactQueryClient.js
```diff
@@ -0,0 +1,16 @@
+import { QueryClient } from 'react-query';
+
+const cacheTimeMinutes = 5;
+
+const createReactQueryClient = () => new QueryClient({
+  // https://react-query.tanstack.com/guides/important-defaults
```
ViktorSoroka07, line 6: Check this link to get the idea of the [important-defaults](https://react-query.tanstack.com/guides/important-defaults) which I did override here.

src/createReactQueryClient.js
```diff
@@ -0,0 +1,16 @@
...
+  defaultOptions: {
+    queries: {
```
ViktorSoroka07, line 8: It is possible to set defaults for the mutations as well, but I did not found use cases when the default ones do not fit folio needs. It is important that by default errored mutations won't be retried.

src/createReactQueryClient.js
```diff
@@ -0,0 +1,16 @@
+import { QueryClient } from 'react-query';
+
+const cacheTimeMinutes = 5;
+
+const createReactQueryClient = () => new QueryClient({
```
ViktorSoroka07, line 5: In order to access the client in the individual applications, the [`useQueryClient`](https://react-query.tanstack.com/reference/useQueryClient#_top) hook should be used so there is no need to export the client from the stripes-core.

mkuklis (approved):
Nice work @ViktorSoroka07!

package.json
```diff
@@ -57,6 +57,7 @@
     "react-intl": "^5.7.0",
     "react-router": "^5.2.0",
     "react-router-dom": "^5.2.0",
+    "react-query": "^3.6.0",
```
doytch, line 60: Should this be a `dependency` and not `devDep`?

src/createReactQueryClient.js
doytch: How smart is react-query about punching through the cache when something is mutated? Eg, if I mutate a `loan`, will the previously fetched and cached loan be refetched or served from cache? Will the collection of `loans` that I previously fetched be refetched?

Feel free to just throw me at docs that explain this :)

doytch (commented):
Exciting! I want to use this :D

aliaksei-chumakou (approved):
look promising :fireworks:

src/createReactQueryClient.js
ViktorSoroka07: `react-query` won't be refetching by default. For that, the [query invalidation](https://react-query.tanstack.com/guides/invalidations-from-mutations) should be used. If you received the updated information from the server after some mutation, there is also the possibility to update the cache without prefetching stuff.

skomorokh: Hm, it seems the query invalidation is done using the name of the query. For our use case that presents a problem as queries span apps and we have no idea what third party app might be using some endpoint we know our mutation has affected. One thing to look into would be a mechanism to invalidate queries by endpoint rather than query name and to raise an issue with the react-query team if that's not possible. Though... it should be possible to go the long way around and generate a list of query names based on an endpoint by looping through the queries.

skomorokh:
Thanks so much Viktor, I'm really glad to see this!

src/createReactQueryClient.js
doytch: Jason, I think you can use the more complex forms of [Query Keys](https://react-query.tanstack.com/guides/query-keys#array-keys) that would support some sort of unique parameters.

...I suppose this does raise the question of whether we should export a `useModuleQuery()` hook that appends the module name to the query key array (or array-ifies it if it's not already).

ViktorSoroka07: I think this is possible to use endpoint as the key. See the example for the [default query fn](https://react-query.tanstack.com/guides/default-query-function). 

I was more thinking about adding an app context to the beginning of the query. For example for data-export use, the key should start with ['data-export', ...other endpoint specific props].

doytch: I also do like the fact that it makes it possible to share queries if you explicitly _want_ to do that. As we split responsibilities across modules, things like controlled vocab lists are an easy candidate for cacheing in app groups like Agreements and Licenses.

Or things like user data for metadata components.

PR #1003, opened 2021-01-25
STRIPES-730: remove webpack from stripes-core

mkuklis:
https://issues.folio.org/browse/STRIPES-730

This PR moves webpack setup out of stripes-core and into [stripes-webpack](https://github.com/folio-org/stripes-webpack).

There is a corresponding PR in stripes-cli related to this work:

https://github.com/folio-org/stripes-cli/pull/244

The tests will fail until we merge the stripes-cli PR.


Jenkinsfile
```diff
@@ -5,6 +5,5 @@ buildNPM {
   runSonarqube = false
   runScripts = [
    ['test:core':'--karma.singleRun --karma.browsers ChromeDocker --karma.reporters mocha junit --coverage'],
```
ryandberger, line 7: This can just be reduced to `test` now and `test:core` can be removed from `package.json` since they do the same thing.

mkuklis: Thanks for catching this @ryandberger!

PR #1004, opened 2021-01-28
STCOR-455 publish ERROR events from RouteErrorBoundary

zburke:
Publish `ERROR` events from `<RouteErrorBoundary>` and provide a
callback to the `<ErrorBoundary>` component that allows its
`componentDidCatch` method to provide error details to the modules that
have registered to handle these `ERROR` events. (Modules of type
`handler` are basically subscribers to events published by stripes-core.)

While existing event-handlers return components to interrupt the render
process, error logging is a bit different: we _don't_ want to interrupt
that process; we just want to log the error. Hence the renaming of
methods such as `getHandlerComponents` to `getEventHandlers` since the
handlers may be components, or they may simply be functions.

Parallel `<ErrorBoundary>` PR in stripes-components: [#1500](https://github.com/folio-org/stripes-components/pull/1500).

Sample loggers: [bugsnag-logger](https://github.com/zburke/ui-bugsnag-logger), [rollbar-logger](https://github.com/zburke/ui-rollbar-logger), [sentry-logger](https://github.com/zburke/ui-sentry-logger). 

Refs [STCOR-455](https://issues.folio.org/browse/STCOR-455), [STCOR-493](https://issues.folio.org/browse/STCOR-493), [STCOM-753](https://issues.folio.org/browse/STCOM-753)


doytch (approved):
Nice. Is there any documentation for how to write these `errorLogging` objects yet?

zburke:
> Nice. Is there any documentation for how to write these errorLogging objects yet?

Fair point. I'll add some documentation before I merge. 

@MikeTaylor describes the general pub-sub architecture in the [comments of STCOR-240](https://issues.folio.org/browse/STCOR-240), and I did write a bunch of POCs ([bugsnag-logger](https://github.com/zburke/ui-bugsnag-logger), [rollbar-logger](https://github.com/zburke/ui-rollbar-logger), [sentry-logger](https://github.com/zburke/ui-sentry-logger)). I feel like we should have a generic "here's how events and `handler` modules work" document, maybe in stripes? I bashed my head against this for a while before I found STCOR-240 and figured out, oh, it's pub-sub; iterating through `getHandlerComponents()`, er, `getEventHandlers()` is how we publish, and a module's `static eventHandler(event)` is how we subscribe, but even that was confusing because existing `handler` modules all return a component from `eventHandler` because they want to interrupt render, which I don't care about here. 

FTR:
```
export default class ErrorLogger extends React.Component {
  static eventHandler(event) {
    if (event === coreEvents.ERROR) {
      const paulBunyan = new ThirdPartyLogger(...);
      // anticipate error and info as from componentDidCatch(error, info)
      return (error, info) => {
        paulBunyan.logThatShiz(error, info);
      };
    }

    return null;
  }

  render() {
    return <></>;
  }
}
```

doytch:
Yeah, there just looked like a bunch of things going on and given that this is intended to be an API surface for non-Folio people, just thought there should be _some_ info about how it all plugs in. Or even about the fact that the functionality _exists._

But I don't want to hold up this PR for that. In fact, I think the docs would be better in the `stripes` repo with other docs than here in stripes-core.

ryandberger:
Looks great! One sanity check I feel the need to ask: did you confirm that [STCOM-336](https://issues.folio.org/browse/STCOM-336) can still show the stack-trace in all its glory?

zburke:
Good thinking, @ryandberger. This doesn't change how `<ErrorBoundary>` renders at all. Merely, it provides the opportunity to publish the values received by `componentDidCatch` via a callback. IOW, yes, STCOM-336 continues to work exactly as-is.

mkuklis (approved):
Nice work @zburke 👏 👏 👏 !

ryandberger (approved):

PR #1006, opened 2021-01-28
STCOR-421 remove hard-source-webpack-plugin

zburke:
I want to build without warnings
I want to `yarn` with no old deps
I want to see naked, that terminal
That console, so true errors shine through
@skomorokh put this in here
So now what, so now what?

Wanting, needing, waiting
Will this decruftify my build? (my build)
Hoping, praying
Will this decruftify my build?

I run it local
Not like `--cache`
I don't want an ancient devdep
I don't want abandoned devdeps either
I just want to run `yarn` cleanly
exit zero, `yarn outdated`?
ZERO, that's right, zero.

Wanting, needing, waiting
Can this decruftify my build?
Yearning, burning
Please help decrutify my build.

Refs [STCOR-421](https://issues.folio.org/browse/STCOR-421), [STCOR-510](https://issues.folio.org/browse/STCOR-510), [STRIPES-675](https://issues.folio.org/browse/STRIPES-675)


skomorokh (approved):
Yay finally! Thank you. I do think I mentioned we should turf this a time or two, apologies for not just doing it. I think there's an option that passed this through in stripes-cli also.

mkuklis (approved):

zburke:
You're right, @skomorokh; `stripes-cli` accepts `--cache` and passes it along to `stripe-score`. Do you think we should remove that option from `stripes-cli` ([STCLI-164](https://issues.folio.org/browse/STCLI-164)) first, or just add a "you can pass `--cache` but stripes-core no longer cares about it" note and leave it alone as a do-nothing-for-now option?

Removing the option altogether would be cleaner, but also technically a breaking change, which is fine, and we may publish a new major version soon anyway in order to yank out the `nightmare` infrastructure.

skomorokh:
They both make sense. We plausibly will add back some similar functionality but no sense keeping it around when you're doing a major anyhow. I've no strong feelings on it, just glad it's not hanging around anymore taking up a bit of space and slightly slowing all the yarn installs and adding a line to yarn outdated.

Baroquem (approved):

PR #1010, opened 2021-02-02
Revert "STRIPES-730: remove webpack from stripes-core (#1003)"

mkuklis:
This reverts commit 6d2d1f1e8147adf79f3554814a6bf629a9b346f6.


mkuklis:
@aditya-matukumalli reported couple issues related to this:

1. When then older version of stripes-cli is installed globally and the new version of stripes-core is installed locally it will create issues since stripes-cli will try to use the local version of stripes-core to look for webpack config (which won't exist anymore).

2. After removing stripes-core from the list of stripes-cli dependencies the issues with a newer version of rxjs started popping up. The reason for this is that one of the dependencies in stripes-cli:

https://github.com/folio-org/stripes-cli/commit/723f0f9be8a86ea6512a2fa044306d1096f1a2d2#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519

depends on rxjs v6. 

This didn't come up before because we were protected by the version of rxjs set in stripes-core:

https://github.com/folio-org/stripes-core/blob/master/package.json#L94-L95

PR #1011, opened 2021-02-02
Goodnight lint

zburke:
In the great stripes-core
There was a bad prop-type
And a conflicting scope
And ESLint --
It totally went apeshit about this.
And there were legit uses of `for (const ... in ...)`
And there was a bogus ESLint warning about this
And there was an irritated developer whispering, "STFU ESLint. If I
could reach through the internet and stab you I would totally do that."

The End.


doytch (approved):
7ab

mkuklis (approved):

ryandberger (approved):
Love the story!

PR #1012, opened 2021-02-03
STCOR-515 refactor to avoid context-related error

zburke:
This maybe broke when we upgraded `react-redux` in #985, though
possibly it was broken earlier since the correct context for statically
configuring context is `contextType` and here we had `contextTypes`, but
it's been like that for years, so who knows.

Refs [STCOR-515](https://issues.folio.org/browse/STCOR-515)


zburke:
p.s. [SSO configuration instructions](https://github.com/folio-org/mod-login-saml/blob/master/GUIDE.md#testing)

src/components/SSOLanding.js
```diff
@@ -1,82 +1,57 @@
...
-          {' '}
-          cookie or query parameter
-        </div>
-      );
+  useEffect(() => {
+    const token = getToken();
+    if (token) {
+      requestUserWithPermsDeb(okapi.url, store, okapi.tenant, token);
     }
+  });
```
mkuklis, line 35: @zburke should we add `[token]` here? and perhaps move `const token = getToken();` out of `useEffect`?

Otherwise I think this will execute on every rerender (which perhaps is intentional here?).

mkuklis (approved):
Nice cleanup @zburke! It's great to see a functional rewrite.

I left one little comment.

SergiyVSergiyenko (approved):

Baroquem (approved):

zburke:
Note the change to `test/bigtest/helpers/setup-application.js` here in order to allow testing with cookies by first clearing all existing cookies and then adding those that are passed to `setupApplication()`. 

Thinking through this now, I wonder if this should be split into two separate functions called from `setup` and `teardown` and that only the cookie that is specifically set should be specifically cleared. Thoughts? Do folks think this is likely ok, or likely risky?

zburke:
(FWIW, stumbled on this while investigating [STCOR-514](https://issues.folio.org/browse/STCOR-514).)

PR #1013, opened 2021-02-03
STRIPES-730: remove webpack from stripes-core

mkuklis:
https://issues.folio.org/browse/STRIPES-730

This PR removes webpack config which now sits in [https://github.com/folio-org/stripes-webpack](https://github.com/folio-org/stripes-webpack).

This is now based on stripes-cli v2.0.0.


zburke (approved):
👏 👏 👏  🏆

PR #1014, opened 2021-02-05
STCOR-506: Dispatch setOkapiReady only after all resources are loaded

mkuklis:
https://issues.folio.org/browse/STCOR-506

The issue described in STCOR-506 was related to the fact that the `setOkapiReady` was dispatched inside `getBindings`:

https://github.com/folio-org/stripes-core/compare/STCOR-506?expand=1#diff-c9d695b1d57404dfed01196c8cf479c0e4cd48583f6c78e7db23402582846864L182

This would incorrectly mark okapi as ready causing `Root` to rerender before all other resources were loaded:

https://github.com/folio-org/stripes-core/blob/master/src/components/Root/Root.js#L83

This PR is changing the time when the  `setOkapiReady` is dispatched. It will happen only after all requests to resources are fetched and processed.

### DIFF
--- a/src/discoverServices.js
+++ b/src/discoverServices.js
@@ -8,6 +8,45 @@ function getHeaders(tenant, token) {
   };
 }
 
+function fetchOkapiVersion(store) {
+  const okapi = store.getState().okapi;
+
+  return fetch(`${okapi.url}/_/version`, {
+    headers: getHeaders(okapi.tenant, okapi.token)
+  }).then((response) => { // eslint-disable-line consistent-return
+    if (response.status >= 400) {
+      store.dispatch({ type: 'DISCOVERY_FAILURE', code: response.status });
+      return response;
+    } else {
+      return response.text().then((text) => {
+        store.dispatch({ type: 'DISCOVERY_OKAPI', version: text });
+      });
+    }
+  }).catch((reason) => {
+    store.dispatch({ type: 'DISCOVERY_FAILURE', message: reason });
+  });
+}
+
+function fetchModules(store) {
+  const okapi = store.getState().okapi;
+
+  return fetch(`${okapi.url}/_/proxy/tenants/${okapi.tenant}/modules?full=true`, {
+    headers: getHeaders(okapi.tenant, okapi.token)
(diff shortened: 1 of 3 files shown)


src/discoverServices.js
```diff
@@ -8,47 +8,59 @@ function getHeaders(tenant, token) {
...
+      return response;
+    } else {
+      return response.json().then((json) => {
+        store.dispatch({ type: 'DISCOVERY_SUCCESS', data: json });
+        return Promise.all(json.map(entry => store.dispatch({ type: 'DISCOVERY_INTERFACES', data: entry })));
+      });
+    }
+  }).catch((reason) => {
+    store.dispatch({ type: 'DISCOVERY_FAILURE', message: reason });
+  });
```
axelhunn, line 47: I wonder, if here one of the promises passed to Promise.all() fails, then we get here
- first dispatch 'DISCOVERY_SUCCESS' (before Promise.all)
- and then dispatch 'DISCOVERY_FAILURE' (catch block)

axelhunn (approved):

src/discoverServices.js
axelhunn, line 47: I wonder, if here one of the promises passed to Promise.all() fails, then we get here
- first dispatch 'DISCOVERY_SUCCESS' (before Promise.all)
- and then dispatch 'DISCOVERY_FAILURE' (catch block)

mkuklis: @axelhunn there is a possibility that this happens but I don't think that effects the state given what is happening in the reducer.

mkuklis: But perhaps we should refactor this a bit too (maybe change the names for these events).

aditya-matukumalli (approved):

zburke (approved):
Thank you for cleaning this up! It is much easier to follow what is happening (and not happening) with the way you have organized things.

mkuklis:
I think we still should do a bit more cleanup in this area as you and @axelhunn pointed out but perhaps in a different PR? I'm going to merge this as is for now.

PR #1016, opened 2021-02-09
Added missing rtl-detect dependency

BogdanDenis:
## Description
Added a missing `rtl-detect` dependency that was recently removed, but is still imported in `loginServices`.
This causes test failures in `ui-inventory` module https://jenkins-aws.indexdata.com/job/folio-org/job/ui-inventory/job/UIIN-1407/3/console


zburke (approved):
Whoops, that was clearly an accident! Thanks for catching this.

PR #1018, opened 2021-02-15
STCOR-518 Callouts are not opened after page reload

NikitaSedyx:
purpose: https://issues.folio.org/browse/STCOR-518

issue and possible ways to fix are described in the ticket. Current fix is not ideal (at least it introduces one rerender that can be avoided), but it works and doesn't introduce breaking change.
I think we can change it after release if it's required.


axelhunn (approved):

CHANGELOG.md
```diff
@@ -23,6 +23,7 @@
 * Publish `ERROR` events from `<RouteErrorBoundary>` so handler modules can react to them. Refs STCOR-455.
 * Dispatch `setOkapiReady` when all resources are loaded. Fixes STCOR-506.
 * Configure `swr`. Refs STCOR-516.
+* Callouts are not opened after page reload. Refs STCOR-518.
```
MikeTaylor, line 26: This is a worrying change-log entry! Sure it should say that callouts now  _are_ opened after a page reload!

MikeTaylor (approved):
I don't honestly have the background to understand what's happening here, but I am delighted to see it done!

src/components/Root/Root.js
```diff
@@ -168,15 +178,16 @@ class Root extends Component {
                   onError={config?.suppressIntlErrors ? () => {} : undefined}
                   defaultRichTextElements={this.defaultRichTextElements}
                 >
-                  <CalloutContext.Provider value={this.callout.current}>
+                  <CalloutContext.Provider value={this.state.callout}>
```
aliaksei-chumakou, line 181: isn't it a huge change, that previously `callout.current` were provided, but now it's `callout`?

NikitaSedyx: nope, it's still the same ref, the way of how it's created

previously `this.callout.current` was null by default, now `this.state.callout` is null
`this.callout.current` had a ref link, now it's `this.callout.current` 

so nothing is changed for consumers

but it would be nice to pass `this.callout` instead, in this case all consumers should be updated

aliaksei-chumakou: oh, right. thanks!

aliaksei-chumakou (approved):

mkuklis (approved):
This looks good. Thanks for addressing it @NikitaSedyx!

zburke (approved):
If I understand correctly, the basic problem is that `<Callout>` requires a ref to `<OverlayContainer>`, a component which hasn't been instantiated yet. At present `Callout` is rendered by `Root` with a null ref, and because `Root` never re-renders, that ref remains null. 

This PR has two changes:

1. Lift `OverlayContainer` up a level from `RootWithIntl` to `Root` (which instantiates `RootWithIntl`), so it exists as a sibling to `Callout` rather than a child of it. 
2. Change the way the `ref` provided to `CalloutContext` is instantiated. Previously, the `ref`, initialized to null because the element it will be attached to didn't exist yet, would never be updated because `Root` never re-renders; i.e. it would remain null. Now, because `Callout` sets its ref via `setCalloutRef`, `CalloutContext` will receive an updated ref. 

If this is wrong, can you please explain what is actually happening here?

NikitaSedyx:
@zburke yep, as you described

MikeTaylor:
This problem is  serious impediment, and it's great that we have a solution. Let's go! Please merge ASAP.

PR #1020, opened 2021-02-16
Revert "STCOR-518 Callouts are not opened after page reload"

zburke:
Reverts folio-org/stripes-core#1018

Unfortunately, this causes "Action" menus to render incorrectly. One step forward, two steps back. Rats.

PR #1021, opened 2021-02-16
STCOR-518 Callouts are opened after page reload

NikitaSedyx:
purpose: https://issues.folio.org/browse/STCOR-518

https://github.com/folio-org/stripes-core/pull/1018
https://github.com/folio-org/stripes-core/pull/1020

> Change the way the ref provided to CalloutContext is instantiated. Previously, the ref, initialized to null because the element it will be attached to didn't exist yet, would never be updated because Root never re-renders; i.e. it would remain null. Now, because Callout sets its ref via setCalloutRef, CalloutContext will receive an updated ref.


NikitaSedyx:
magic in `Callout` works without `OverlayContainer` location change

mkuklis (approved):

doytch (commented):
I'm fine with this but can you put a bit of a write-up in the PR description? I don't really like it when we find a bug, fix it, but don't really understand _why_ the fix works. 😆 I know we're just going back to the previous location of `OverlayContainer`...but still. It also makes it less likely for us to make similar goofs in the future if we understand the purpose of `OC`'s location in the tree, etc.

NikitaSedyx:
@doytch I've updated description

it works for callout without OC location change because of trick in callout 
https://github.com/folio-org/stripes-components/blob/master/lib/Callout/Callout.js#L68
https://github.com/folio-org/stripes-components/blob/master/lib/Callout/Callout.js#L23

aliaksei-chumakou (approved):

axelhunn (approved):

doytch:
> @doytch I've updated description
> 
> it works for callout without OC location change because of trick in callout
> https://github.com/folio-org/stripes-components/blob/master/lib/Callout/Callout.js#L68
> https://github.com/folio-org/stripes-components/blob/master/lib/Callout/Callout.js#L23

Right, that makes sense, but my question was about why moving `OverlayContainer` in your first PR - #1018 - broke things. In any case, it doesn't make sense to hold this up over my lack of understanding on that point. But if you could explain why #1018 broke the actions menu either on this PR [or in the Slack thread](https://folio-project.slack.com/archives/C210UCHQ9/p1613485482229300), that'd be helpful and appreciated. 😄

PR #1023, opened 2021-02-23
STCOR-514 dispatch checkSSO on session hydration

zburke:
When restoring a session that was preserved in local-storage, as when
opening a new window when all previous windows had been closed but the
user had never logged out, check SSO/SAML. Previously, these were
handled as a disjuction (if we have a session, rehydrate it; else check
for sso). Now they are simply both handled in series.

There is no reason they need to be alternatives. You could login via
SSO, then close your window, then open a new one, then logout. At this
point, you want to see the "Login with SSO" button again, but previously
you wouldn't have because the SSO/SAML check only happened when starting
new sessions, not when rehydrating existing sessions.

Refs [STCOR-514](https://issues.folio.org/browse/STCOR-514)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@
 * Dispatch `setOkapiReady` when all resources are loaded. Fixes STCOR-506.
 * Configure `swr`. Refs STCOR-516.
 * Callouts are opened after page reload. Refs STCOR-518. 
+* When rehydrating a session from local storage, always dispatch `checkSSO`. Fixes STCOR-514.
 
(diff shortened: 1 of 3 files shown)


Baroquem (approved):

Baroquem (approved):

SergiyVSergiyenko (approved):

mkuklis (approved):
Nice and clean @zburke! Thank you for explaining it too.

doytch:
Thanks for this! Was OOO but I know I've got some people at Duke who'll be psyched for this.

PR #1031, opened 2021-03-12
Added useCustomFields hook

doytch:
The goal here is to:
- use the existing backend implementation of mod-custom-fields
- not hardcode a dependency on a specific implementation of an interface the way that the current stripes-smart-components CF components do.
- provide a hook to access custom fields for ______

Yeah, that blank is doing a lot of heavy lifting. That's because the current design is based around getting custom fields for a specific module, since `custom-fields` is an interface that a module can advertise as part of its ModuleDescriptor. Currently, if we want to show the custom fields for a user, [the existing components look up the custom fields](https://github.com/folio-org/stripes-smart-components/blob/master/lib/CustomFields/pages/ViewCustomFieldsRecord/ViewCustomFieldsRecord.js#L188) for a module that [has a `name` of `"users"`](https://github.com/folio-org/mod-users/blob/master/descriptors/ModuleDescriptor-template.json#L3). This is an unspecified dependency and more critically, breaks the ability to use alternative implementations of modules.

OK, so if we don't use that, what do we use? My thinking here was that the frontend is already aware of the interfaces it relies on. ui-users knows it needs the interface that has an `id` of `users`. So my thinking is that it should be able to say "Give me the custom fields that are related to, or linked to, [the `users` interface.](https://github.com/folio-org/mod-users/blob/master/descriptors/ModuleDescriptor-template.json#L6)"

That last bit - related to, linked to - isn't defined in the architecture or API. But we can infer it by finding a module that provides both a `users` interface, and [a `custom-fields` interface](https://github.com/folio-org/mod-users/blob/master/descriptors/ModuleDescriptor-template.json#L145). If we find that module, then we can use the module's ID to fetch the custom-fields for it and return it to the front end.

**Map to the Code**
`DISCOVERY_PROVIDERS` action: This provides us the data on which to operate. We need a linkage between the module and interfaces it provides, which isn't being saved in the Redux store atm. If we were starting from scratch, I'd rather have this info in the `interfaces` or `modules` entries rather than the new `interfaceProviders` substate, but this makes the change non-breaking.
`useCustomFields` hook: Where we take that that from `stripes.discovery`, poke through it, fetch the custom-fields using react-query if we need to, and return any custom fields.

**Usage**
```
const UserCustomFields = () => {
  const [customFields, error, loading] = useCustomFields('users');
  
  if (loading) return <Spinner />;
  if (error) return <Error>{error}</Error>; 
  
  return (
    <div>
      <h1>These are the Custom Fields that can be defined for a User</h1>
      <ul>{customFields.filter(cf => cf.entityType === 'user').map(cf => <li>{cf.name}</li>)}</ul>
    </div>
  );
}
```

**Blind Spots**
- If a module currently provides `foo` and `custom-fields`, but a future set of implementations splits that across multiple modules, this approach won't work. But given that adding the `custom-fields` interface to a module is trivial (mod-custom-fields handles most of the work itself), I don't think this should be how a backend module should be designed anyway.
- If a future module amalgamates the work of multiple current modules that currently implement custom-fields, then fetching the future list of custom-fields will return a longer list than currently. This should be fine because the frontend should defensively use this list of custom-fields anyway, filtering it by `entityType` or something else.


skomorokh:
It's a shame that we have to go to such lengths but great that you did as it's certainly better than coupling core UI modules to specific implementations! I hope at some point we can improve things so that there is a means to direct the request to the module implementing both the `custom-fields` and `users` interfaces but this neatly abstracts that away so that it'll be a smooth transition once we get there :+1:

mkuklis (approved):
Nicely done @doytch!

MikeTaylor:
Needs a change-log entry.

PR #1033, opened 2021-03-15
STCOR-525 Removed fetch of protected okapi /version endpoint

doytch:
This endpoint [is now permission-protected](https://github.com/folio-org/okapi/blob/200e3929993ecf3f3ed0f2f4ab072ef2c31cf800/okapi-core/src/main/java/org/folio/okapi/managers/InternalModule.java#L369-L374), and the permission generally isn't granted, even to diku_admin users. So fetching it is useless since it always errors out and just dumps an error into the console. 

I've mentioned this before and didn't hear anything about this endpoint's permissions being changed so I'm assuming it'll be staying like this.


zburke (approved):
Filed [STCOR-525](https://issues.folio.org/browse/STCOR-525) to capture this work so we can assign it to a release, and linked it to [OKAPI-917](https://issues.folio.org/browse/OKAPI-917) for posterity.

adamdickmeiss:
Sorry, wan't aware, or it slipped .. We'll see what we can do abt OKAPI-917.

mkuklis (approved):
@doytch thanks for cleaning this up.

zburke:
@doytch, while trying to debug something else I discovered older versions of stripes-core, e.g. v4.1.0, shows the okapi version even though we're hitting an okapi endpoint that should restrict access to `/version`. I think it's just getting the okapi version from `/_/proxy/tenants/${tenantId}/modules` rather than from `/_/version` though I haven't looked carefully. 

I dunno if that's worth restoring, i.e. if people care that much, but pulling out things that used to work always irks me a bit.

doytch:
> @doytch, while trying to debug something else I discovered older versions of stripes-core, e.g. v4.1.0, shows the okapi version even though we're hitting an okapi endpoint that should restrict access to `/version`. I think it's just getting the okapi version from `/_/proxy/tenants/${tenantId}/modules` rather than from `/_/version` though I haven't looked carefully.
> 
> I dunno if that's worth restoring, i.e. if people care that much, but pulling out things that used to work always irks me a bit.

Huh, yeah looks like it. Now that we have the ability to (manually, in the front-end) look up which module provides an interface we could just look up who provides the `"okapi"` interface. I'm not against it.

PR #1035, opened 2021-03-19
STCOR-527 read user's locale from config on login

zburke:
On login, look in `/configurations/entries/...` for user-specific locale
settings, allowing them to overwrite the tenant's settings. Restore the
tenant's settings on logout.

Refs [STCOR-527](https://issues.folio.org/browse/STCOR-527)


doytch (approved):
Though I'm usually a proponent of "do it right the first time" approaches, it seems like it does make sense to just do this piecemeal since apparently even though it annoys everyone, we're never gonna be able to roll out the full user locale stuff all at once. So I'm in favour of this.

I will say, one thing that we've noticed at Duke when speccing out the feature of multi-timezone tenants are the issues with automated messages like patron notice emails. Those are hard-wired into the tenant's timezone so making front-end changes like this puts the emails out of step with what a Folio user sees. This could be a bit confusing. 

I think we can note that if/when we actually have a UI setting for writing these configs, but it bears keeping in mind.

src/loginServices.js
```diff
@@ -161,6 +161,31 @@ export function getLocale(okapiUrl, store, tenant) {
     });
 }
 
+export function getUserLocale(okapiUrl, store, tenant, userId) {
+  if (userId) {
+    return fetch(`${okapiUrl}/configurations/entries?query=(configName==localeSettings and userId==${userId})`,
```
doytch, line 166: Could we actually wrap this in a `hasPerm` so we don't nail the console with an error on login?

zburke: Practically speaking, if a user has permission to do anything then they're gonna end up with config-read permission, so I think the check is unlikely to fail in real-world scenarios. That said, it's  obviously The Right Thing to do here, and for all these on-login fetches.

zburke: Stashed the check in `loadResources()`, which calls `getUserLocale()` et al.

src/loginServices.js
```diff
@@ -131,17 +152,17 @@ export function loadTranslations(store, locale, defaultTranslations = {}) {
...
- * @return Promise
+ * @param {string} url location of locale information
+ * @param {object} store redux store
+ * @param {string} tenant
+ * @returns Promise
  */
-export function getLocale(okapiUrl, store, tenant) {
-  return fetch(`${okapiUrl}/configurations/entries?query=(module==ORG and configName==localeSettings)`,
+function dispatchLocale(url, store, tenant) {
+  return fetch(url,
```
doytch, line 165: Should we move away from `fetch` towards something like `ky` in cases like this so we can get the headers for free? Esp now with i18n headers?

I know it's already on fetch and it'll be annoying to refactor. It's fine to say "naaaah, not now."

doytch (approved):

src/loginServices.js
zburke: Naaaah, not now. This is an easy/small refactor where I don't need to rewrite a bunch of tests. 

Good thinkin' though; filed [STCOR-546](https://issues.folio.org/browse/STCOR-546).

PR #1037, opened 2021-03-29
Compile translations

ryandberger:
- Compiles Stripes Core translations into AST format
- Fixed translation files that were failing compilation because of syntax errors


doytch:
I don't quite understand the process with these translations, could you go over it to explain a bit what we need and what functionality is available?

My understanding is that we'll need to compile the translations in _all_ our modules, basically mirroring this PR everywhere. And then run that NPM script each time we edit the translations so that it's available for the changes in https://github.com/folio-org/stripes-webpack/pull/12

Assuming my understanding is correct (big assumption there!), what's stopping us from bundling the translation compilation into, say, a webpack plugin so that it happens automagically and no module dev needs to think about this?

ryandberger:
> I don't quite understand the process with these translations, could you go over it to explain a bit what we need and what functionality is available?
> 
> My understanding is that we'll need to compile the translations in _all_ our modules, basically mirroring this PR everywhere. And then run that NPM script each time we edit the translations so that it's available for the changes in [folio-org/stripes-webpack#12](https://github.com/folio-org/stripes-webpack/pull/12)
> 
> Assuming my understanding is correct (big assumption there!), what's stopping us from bundling the translation compilation into, say, a webpack plugin so that it happens automagically and no module dev needs to think about this?

So, I had originally thought of doing a centralized compile, but the compilation script does validation against the translations and if there are any syntax issues, it will fail. Rather than having to chase down breaking changes in modules during a platform build, I figured it would be better for compilation to happen at the module level and act as a linting step for translations. I brought this up at an architecture meeting with @zburke, and he was in agreement with this approach, even though it will require some work to be farmed out. 

I will be adding documentation that teams can refer to and also plan to update `ui-app-template` accordingly.

doytch:
> So, I had originally thought of doing a centralized compile, but the compilation script does validation against the translations and if there are any syntax issues, it will fail. Rather than having to chase down breaking changes in modules during a platform build, I figured it would be better for compilation to happen at the module level and act as a linting step for translations. I brought this up at an architecture meeting with @zburke, and he was in agreement with this approach, even though it will require some work to be farmed out.

Makes sense! So Jenkins will compile them before bundling it all up and plopping the package onto the NPM repo...that's handy!

> I will be adding documentation that teams can refer to and also plan to update `ui-app-template` accordingly.

🙏 Thanks for this!

doytch (approved):

PR #1044, opened 2021-04-08
STCOR-534: Move CalloutContext back to RootWithIntl

mkuklis:
This PR moves `<CalloutContext>` back to `<RootWithIntl>`. This was our original setup back in:

https://github.com/folio-org/stripes-core/commit/a2ab520d33308fd415807ab1d3ac871f8e190114

https://issues.folio.org/browse/STCOR-534

The previous setup was causing issues with the `Callout` ref being lost between re-logins since  it's only created once when the app loads initially:

https://github.com/folio-org/stripes-core/blob/master/src/components/Root/Root.js#L188

This change makes sure that:

1. Callout works when locale is changed (STCOR-481)
2. Menus and dialogs display correctly
3. Callout works during re-login and page refresh

### DIFF
--- a/src/RootWithIntl.js
+++ b/src/RootWithIntl.js
@@ -10,6 +10,7 @@ import { CookiesProvider } from 'react-cookie';
 
 import { HotKeys } from '@folio/stripes-components/lib/HotKeys';
 import { connectFor } from '@folio/stripes-connect';
+import { Callout } from '@folio/stripes-components';
 
 import getModuleRoutes from './moduleRoutes';
 import events from './events';
@@ -35,6 +36,7 @@ import {
   AppCtxMenuProvider,
 } from './components';
 import { StripesContext } from './StripesContext';
+import CalloutContext from './CalloutContext';
 
 class RootWithIntl extends React.Component {
   static propTypes = {
@@ -52,6 +54,14 @@ class RootWithIntl extends React.Component {
     token: '',
     history: {},
   };
+
+  state = { callout: null };
+
+  setCalloutRef = (ref) => {
+    this.setState({
+      callout: ref,
+    });
+  }
 
   render() {
     const {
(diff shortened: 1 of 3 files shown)


src/RootWithIntl.js
```diff
@@ -83,7 +84,6 @@ class RootWithIntl extends React.Component {
                         />
                         { (stripes.okapi !== 'object' || stripes.discovery.isFinished) && (
                           <ModuleContainer id="content">
-                            <OverlayContainer />
```
NikitaSedyx, line 86: @mkuklis have you tried dropdowns and modals? in https://github.com/folio-org/stripes-core/pull/1018/files I broke them because `OverlayContainer` was moved from `MainContainer`

mkuklis: Yeah I tested couple of them and they seemed to work. Let me try to load couple more modules to double test it.

zburke (approved):

src/RootWithIntl.js
NikitaSedyx: Screen Shot 2021-04-09 at 09 05 41
Screen Shot 2021-04-09 at 09 06 18
seems dropdowns are ok, but not modals

mkuklis: Thanks @NikitaSedyx! I will take another look at this.

mkuklis:
@NikitaSedyx I ended up moving `CalloutContext` back to `RootWithIntl`. This is where it was when we started having various issues with the `Callout` originally.

Could you please verify if things work with this setup on your end?

I tried to cover all the cases we ran into in the past but I just want to make sure things are working well for you too.

mkuklis:
I merged this in. Please let me know if anything breaks :)

NikitaSedyx:
@mkuklis sorry, had no chance to test it, but looks good to me, I hope everything works :)

PR #1051, opened 2021-04-22
STCOR-531 Add help button

elsenhans:
https://issues.folio.org/browse/STCOR-531

As a staff person
I want to access the FOLIO help site via a ? help icon from the upper right hand side of the universal header
So that I have ready access to help documentation

**Current situation or problem:** There is no way to access the help site from the UI of FOLIO
**Requirement:** To open the FOLIO UI and have access to a button on the top menu bar that launches https://docs.folio.org/docs/ in a new browser tab/window.

**Mockups**
image
image


rchr (approved):
Looks good to me! Thank you!

mkuklis (approved):

PR #1056, opened 2021-05-03
STCOR-529: Introduce module hierarchy provider

mkuklis:
https://issues.folio.org/browse/STCOR-529

The `useModuleHierarchy` hook will return an array of strings. Each string will represent a package name for the given module or plugin used in the hierarchy. 

For example the result from `useModuleHierarchy` executed from the `ui-inventory` will look like this:

`['@folio/inventory']`

The result from `useModuleHierarchy` executed from the `ui-find-instance` plugin embedded inside `ui-inventory` will look like this:

`['@folio/inventory', '@folio/find-instance']`


zburke (commented):
Hmmm, the case of a plugin within an app is a core use-case of STCOR-537, but I don't think this will help with that since it will only return the top-level module. I think at the very least we need that hierarchy, as @skomorokh suggests: 
> it could be just the config for the current module and the nesting can just be package names added to an array under a `moduleNesting` key or something.

The module config is nice, but I think the nesting is actually the most important part.

skomorokh:
https://github.com/folio-org/stripes-core/blob/dc3bd495e00c69737bc145f36c9a2780ed31ba7e/src/moduleRoutes.js#L55

If you put it here you already have the module config available to you in the scope. I find the approach of using the route to look it up very disquieting. Should be pretty easy to put in that file and if you just want it at the top level for apps and then you can build on this work to add the nesting for plugins/handlers later. Settings will also need a top-level provider at some point...

mkuklis:
@skomorokh interesting idea. Let's chat about this a bit more tomorrow so I can understand what you are proposing a bit better. I think ideally it would be nice to just use a single provider for all modules. I see that it could be handy to have access to a given module since we are already looping in `moduleRoutes.js` but then we would end up with one provider per module?
 
I think we will need to rely on react-router (at least the `pathname`) to detect which module is currently active. 

@zburke in my mind the plugin namespacing will happen in STCOR-537 but maybe I misunderstood it?

mkuklis:
@zburke I think you are right here and now when I'm reading it in more details it looks like we will need to enhance the `Pluggable` to update the current module context. @skomorokh let's chat a bit more about this tomorrow.

mkuklis:
@skomorokh the last commit should be a bit closer to what you had in mind (I hope haha). 

@zburke I introduced `moduleNesting` which should include nested plugins (`Pluggable` has been wrapped in the current module provider now).

mkuklis:
@skomorokh thank you for working on this with me and for your feedback and pushing for the array. I renamed this to `ModuleHierarchy`.

The `useModuleHierarchy` hook will return an array which will represent a hierarchy.

When executed from the app module A the result with be:

`[moduleAConfig]`

When executed from the plugin B appended inside module A the result will be:

`[moduleAConfig, PluginBConfig]`

skomorokh (approved):
@mkuklis thanks, it was fun! And I think this'll work out and you can get the current module via `moduleHierarchy[-1]`.

Is the module config already passed to Handler module callbacks?

mkuklis:
@skomorokh I replace the config with just a simple string (the package name) for each module in the hierarchy. The handlers should be now wrapped up too.

zburke (approved):
This is really nice work!

MikeTaylor (approved):
This looks good to me!

The big thing that's missing is documentation. But since the _Developer's Guide_ got moved into a different repository, perhaps the documentation now belongs there, and so is not relevant to the present PR.

One other thing: since what this will actually be used for is namespacing, I'm thinking that nine times out of ten, client code will be calling `moduleHierarchy.join('/')` — or sometimes `moduleHierarchy.join('.')` or `moduleHierarchy.join('-')` or any of a dozen other variants. That seems like exactly the kind of chaos we wanted to avoid. So couldn't we also have a thin layer `useModuleHierarchyAsString` that returns a string with the components joined in a canonical manner?

mkuklis:
Thanks for the feedback @MikeTaylor. I believe what you are proposing will happen in https://issues.folio.org/browse/STCOR-537 described by @skomorokh  with `useNamespace` api outlined by @doytch. Or are you thinking about using `useModuleHierarchy` for other purposes?

MikeTaylor:
> I believe what you are proposing will happen in https://issues.folio.org/browse/STCOR-537 described by @skomorokh with `useNamespace` api outlined by @doytch.

Ah, there is it. I'm obviously not sufficiently familiar with this work to comment intelligently :-)

Great, please proceed!

Baroquem (approved):

PR #1061, opened 2021-05-07
STCOR-537: Introduce useNamespace hook

mkuklis:
https://issues.folio.org/browse/STCOR-537

This PR introduces a `useNamespace` hook described by @skomorokh and outlined by @doytch under [STCOR-537](https://issues.folio.org/browse/STCOR-537).

The signature currently looks like this:

````js
const [namespace, getNamespace] = useNamespace(options);
````

The `namespace` represents a namespace as a string. The `getNamespace` is a function which can be used to retrieve a namespace with a different set of options:

````js
const ns = getNamepace({ key: 'test-key' });
````

Possible options:

- key - a string which will be appended to the given namespace
- ignoreParents - a boolean which can be used to ignore the parents part from a module hierarchy. For example when `useNamespace` is executed from a plugin embedded in app module (e.g. ui-plugin-find-order executing in ui-agreements) and `ignoreParents` is used it will only return `@folio/plugin-find-order` as a namespace.

The PR does not include `useGlobalNamespace` as it felt like we should handle it in a separate JIRA/issue.

### DIFF
--- a/src/hooks/useNamespace.js
+++ b/src/hooks/useNamespace.js
@@ -1,0 +1,31 @@
+import { useModuleHierarchy } from '../components';
+
+// A hook which returns module namespace as a string
+// https://issues.folio.org/browse/STCOR-537
+
+// Example usage:
+
+// from app module (e.g. ui-users)
+// const namespace = useNamespace(); // "@folio/users"
+
+// from plugin embedded in app module (e.g. ui-plugin-find-order executing in ui-agreements)
+// const namespace = useNamespace(); // "@folio/agreements:@folio/plugin-find-order"
+
+// from plugin embedded in app module with `ignoreParents` option (e.g. plugin ui-plugin-find-order executing in ui-agreements)
+// const namespace = useNamespace({ ignoreParents: true }); // "@folio/plugin-find-order"
+
+// from plugin embedded in app module with `key` option present (e.g. ui-plugin-find-order executing in ui-agreements)
+// const namespace = useNamespace({ key: "filters-pane" }); // "@folio/agreements:@folio/plugin-find-order:filters-pane"
+const useNamespace = (options = {}) => {
+  const moduleHierarchy = useModuleHierarchy();
+  const { ignoreParents, key } = options;
+  let namespace = ignoreParents ? moduleHierarchy.pop() : moduleHierarchy.join(':');
+
+  if (key) {
+    namespace += `:${key}`;
+  }
+
+  return namespace;
+};
+
+export default useNamespace;
--- a/src/hooks/useNamespace.js
+++ b/src/hooks/useNamespace.js
@@ -20,10 +20,11 @@ import { delimiters } from '../constants';
 const useNamespace = (options = {}) => {
   const moduleHierarchy = useModuleHierarchy();
   const { ignoreParents, key } = options;
-  let namespace = ignoreParents ? moduleHierarchy.pop() : moduleHierarchy.join(delimiters.NAMESPACE_DELIMITER);
+  const { NAMESPACE_DELIMITER } = delimiters;
+  let namespace = ignoreParents ? moduleHierarchy.pop() : moduleHierarchy.join(NAMESPACE_DELIMITER);
 
   if (key) {
-    namespace += `${delimiters.NAMESPACE_DELIMITER}${key}`;
+    namespace += `${NAMESPACE_DELIMITER}${key}`;
   }
 
   return namespace;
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: src/hooks/useNamespace.js [hunk 1]
```diff
@@ -0,0 +1,31 @@
...
+
+// from plugin embedded in app module with `key` option present (e.g. ui-plugin-find-order executing in ui-agreements)
+// const namespace = useNamespace({ key: "filters-pane" }); // "@folio/agreements:@folio/plugin-find-order:filters-pane"
+const useNamespace = (options = {}) => {
+  const moduleHierarchy = useModuleHierarchy();
+  const { ignoreParents, key } = options;
+  let namespace = ignoreParents ? moduleHierarchy.pop() : moduleHierarchy.join(':');
+
+  if (key) {
+    namespace += `:${key}`;
```
zburke, line 25: Maybe move `:` into `const NAMESPACE_DELIMITER = ':'`? That's either CSci 101 or a total jerk manager everything is fine but I have to put my thumbprint on it move. Your call :) I don't think it needs to be exported, though if `useGlobalNamespace` is implemented separately, maybe it would be helpful for it to be able to leverage that value? OTOH, if we're thinking of the string as opaque-but-user-friendly, then maybe it doesn't matter.

zburke (approved):
😍 pretty tests! This looks great! Thank you!

src/hooks/useNamespace.js
mkuklis: Nice idea @zburke Thank you!

doytch (approved):
I log back in after vacation and find the hierarchy PR done and merged and this looking great. Thanks for this Michal!

skomorokh:
Really pushes things forward, thanks Michal!

I think it might be more flexible if we have the hook return a function so that we can sometimes do:
```
const getNamespace = useNamespace();
```

And  then we can use `getNamespace({ options })` in more deeply nested place as it won't have to follow the Rule of Hooks? I think this might be especially helpful when we're making use of the `key` option as what we want for our key might not be conveniently available up front. You can of course still use the same pattern the current implementation suggests via:
```
const namespace = useNamespace()({ options });
```

Also, in cases where we use multiple namespaces in a file, I think I'd rather see this call inline where the query is being set up as `ns({ options })` than as `goFindWhereIDefinedThis`.

As a bonus, I wonder if there are any minor performance implications to multiple invocations of hooks vs. single invocation with multiple function calls? But hypothetical performance issues are not a reason unto themselves until profiling.

mkuklis:
@skomorokh I'm not sure if I follow why we would want to make it more complex. The namespace would not really change after executing it once in a single place (unless we go deeper in the hierarchy). Are you thinking that we may have a situation where we execute `useNamespace` with different parameters at the same level of hierarchy? Even then I think executing `useNamepace` multiple times and dealing with strings rather than getting a function feels more natural.

Perhaps you can provide an example (a code snippet) where would could see the benefit of a function? 
I'm fine with keeping this discussion open but I think what @doytch proposed is pretty solid to me. It's simple and very easy to reason about.

skomorokh:
I don't see returning a single function as especially complex and hooks that return functions (eg. setState()) are a common pattern. The use cases that prompted my suggestion were as outlined, but the one that's most important and potentially hard to work around is the ability to use something for `key` that's not readily available at the top level where hooks are called. Are you so sure we'll never want that it's worth losing the option to nest these calls over an extra pair of parens?

doytch:
Yeah I can see both sides to this. I _really_ prioritise API cleanliness in general, but do see Jason's points here on having the call be as close to the point-of-use as possible (cleans up one bit of code in exchange for "dirtying up" a difference piece) while also supporting other uses where you don't have the key available. I would like to see examples of those kinds of situations though if we're gonna suggest that that's a key reason for adding this functionality.

Other alternatives we have are:
1. Match the `useState` signature a bit more closely via something like `const [namespace, getNamespace] = useNamespace(opts)`. That way the simple case is handled via `const [ namespace ] = useNamespace()`, but you can also do complex work via `const [ _, getNamespace ] = useNamespace(); ..... console.log(getNamespace({ key: 'foo' });`.
2. Have separate hooks that return just the string or just the function. At the risk of biasing the jury, this feels suboptimal.

skomorokh:
I have no specific examples nor time to spend thinking of one. I just suspect they're possible (as you say, a loop is a good example)  and for me that's enough to add the two characters () because I am not bothered by their presence. And I much prefer the calls to this be inline when they're one-offs but that's a subjective stylistic choice. At any rate, entirely your call, I'm just glad we have namespacing now, thank you!

zburke:
Though I appreciate @skomorokh's concern about limiting the flexibility of this if we only return the string, I would opt for the simplicity of a string to address our known situation over the complexity that comes with building a flexible solution for a possible future situation we can't actually envision but that we agree might theoretically exist. 

That said, if we really don't want to let go of the flexibility, @doytch's suggestion:

> Match the useState signature a bit more closely via something like 
```const [namespace, getNamespace] = useNamespace(opts)```

seems pretty nice to me.

mkuklis:
Ok just made couple more changes to support the new signature for this.

PR #1066, opened 2021-05-20
STCOR-543: Able to close <AppContextDropdown> outside

Vladyslav-Velytskyi:
# STCOR-543 App context menu dropdown not closing

`AppContextDropdown` does not closing when user click outside the dropdown menu. It's needed to pass `open` prop to `DropdownMenu`

Refs [STCOR-543](https://issues.folio.org/browse/STCOR-543)

### Before
before
### After
after


BogdanDenis (approved):

mkuklis (approved):

zburke (approved):
Using a `renderMenu` prop instead of `children` allows the focus-handling from `<Dropdown>` to filter down, is that the idea? That's good to know. 

@JohnC-80, do you think this should be spelled out more explicitly in the [Dropdown README](https://github.com/folio-org/stripes-components/tree/6029e92352ee49a9c208e502b8ee9fddaa765eb3/lib/Dropdown)? Quite possibly, focus handling _is_ sufficiently described there but I'm just too thick to recognize it 😆 .

JohnC-80 (approved):

PR #1067, opened 2021-05-20
STRIPES-683 Set credentials: include

hjiebsco:
Minor change. Please see https://issues.folio.org/browse/STRIPES-683 Set credentials: include on fetch to /saml/login.

### DIFF
--- a/src/loginServices.js
+++ b/src/loginServices.js
@@ -484,6 +484,7 @@ export function requestSSOLogin(okapiUrl, tenant) {
   fetch(`${okapiUrl}/saml/login`, {
     method: 'POST',
     headers: { 'X-Okapi-tenant': tenant, 'Content-Type': 'application/json' },
+    credentials: 'include',
     body: JSON.stringify({ stripesUrl }),
   })
     .then(resp => processSSOLoginResponse(resp));


zburke (approved):

zburke:
@hjiebsco, I just now realized this PR was against the `b7.1` branch rather than `master`. Was that the intent? If yes, ok, but do we want to raise a similar PR against `master` so this can be included in Juniper, rather than just in a patch release for Iris?

zburke:
And do we have approval for a formal patch release? [R1 2021 Hotfix process](https://wiki.folio.org/pages/viewpage.action?pageId=59933801)

hjiebsco:
> @hjiebsco, I just now realized this PR was against the `b7.1` branch rather than `master`. Was that the intent? If yes, ok, but do we want to raise a similar PR against `master` so this can be included in Juniper, rather than just in a patch release for Iris?

I used b7.1 because my testing env has 7.1.1. You are right, this change should go into master.

hjiebsco:
> And do we have approval for a formal patch release? [R1 2021 Hotfix process](https://wiki.folio.org/pages/viewpage.action?pageId=59933801)

My bad. It was not meant to be for a patch release. It is part of the login-saml CSRF feature development.

PR #1071, opened 2021-06-01
useOkapiKy: Pass locale via accept-language header

doytch:
I saw https://github.com/folio-org/stripes-connect/pull/179 which sets the `accept-language` header for stripes-connect queries and realised we should have a `ky` version of this functionality as well. This implements that.


zburke (approved):
Whoops, can't believe I missed this. Thanks for catching it!

axelhunn (approved):

skomorokh (approved):

mkuklis (approved):
👏

PR #1081, opened 2021-06-09
STCOR-551 configure karma for v6 compatibility

zburke:
There are some overlapping things going on here.

* `stripes-webpack` supports the new JSX transform that means we
  shouldn't have to `import React ...` any longer, but BigTest tests
fail without it when using `karma` `v4` (from `stripes-cli` `v1.2`).
* when using `karma` `v6` (from `stripes-cli` `v1.3`), the karma config
  directives `files` and `preprocessors` cause test files not be found
at all.

Bumping the `stripes-cli` dep (so we get `karma` `v6`) and cleaning up
`karma.config.js` (so it is `v6` compatible, and more aligned with
configs in other repos) allows tests to run as they always did. Phew.

Refs [STCOR-551](https://issues.folio.org/browse/STCOR-551)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,8 @@
 * Indicate that logging out of FOLIO will not affect an SSO session. Refs STCOR-532.
 * Introduce `withNamespace` HOC. Refs STCOR-542.
 * Pull user's locale settings from configurations on login, if available. Refs STCOR-527.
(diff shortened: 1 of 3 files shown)


doytch (approved):
Makes sense. Thanks for toiling in the mines. ⛏️

PR #1083, opened 2021-06-10
Folio 3067 replace jenkins

ankitasen-ubmainz:
Proper conditional syntax added for the NPM publish and Mod descriptor publish step. The conditionals work properly and skip if on any other branch than main/master


zburke:
@ankitasen-ubmainz, @funkymalc, This work causes all PRs in this repo appear dirty because of the failed buildNPM Snapshot jobs that are attached to them. Is it possible either finish this or remove the half-implemented piece that is causing those failures? Thanks!

PR #1089, opened 2021-07-08
FOLIO-3067 invoke stripes via yarn

zburke:
Call `yarn stripes ...` instead of just `stripes ...` in order to use
the locally-installed version provided by the dev-deps rather than
relying on a global install which is outside the purview of an
individual build.

Refs [FOLIO-3067](https://issues.folio.org/browse/FOLIO-3067)

### DIFF
--- a/.github/workflows/buildnpm.yml
+++ b/.github/workflows/buildnpm.yml
@@ -1,7 +1,7 @@
-# This workflow will do a clean install of node dependencies, build the source code, 
-# run unit tests, and perform a SonarCloud scan. 
+# This workflow will do a clean install of node dependencies, build the source code,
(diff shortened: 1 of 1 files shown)


.github/workflows/buildnpm.yml
```diff
@@ -156,5 +160,3 @@ jobs:
           file: '{ "file": "${{ env.MODULE_DESCRIPTOR_DIR }}/${{ env.FOLIO_NAME }}.json" }'
```
zburke, line 160: I think the build-failure-log indicates `file: ` should be `files: `.

zburke:
Whoops; I missed #1083, which runs cleanly.

PR #1090, opened 2021-07-09
Fix React unique-key error

MikeTaylor:
Originally, the loop at the end was just

	this.components.map(Component => (
	  <Component key={Component.name} {...props} />
	)

But when the invocation of `<Component>` got wrapped in `<ModuleHierarchyProvider>` in commit bf282ea1, the `key` was left in the wrapped `<Component>`, so it no longer appeared directly inside the list generated by `map`.

This commit fixes this by moving the `key` up a level.


zburke (approved):

mkuklis:
@MikeTaylor nice catch!

PR #1100, opened 2021-08-11
STRIPES-723 Upgrade RxJS to v6

zburke:
https://issues.folio.org/browse/STCOR-536

There is a corresponding PR in stripes-connect which is tied to this work too:

https://github.com/folio-org/stripes-connect/pull/164

Also right after I upgraded this I realize that eholdings is also using epics:

https://github.com/folio-org/ui-eholdings/tree/master/src/redux/epics

They will have to be upgraded too before we can merge this work.


mkuklis (approved):

zburke:
This is no longer necessary since the stripes 7/react 17 work merge to master separately.

PR #1101, opened 2021-08-13
FOLIO-3253 POC tweak login to avoid users-bl

zburke:
POC refactor of login to avoid `users-bl`, which has lots of
dependencies, in favor of `authn/login`, which has very few. This is not
intended to be pretty (obviously, the extra requests need cleaning up
and error handling) but just to demonstrate that it is possible to login
from the UI against a minimal set of back-end and front-end modules.

Refs [FOLIO-3253](https://issues.folio.org/browse/FOLIO-3253)


MikeTaylor:
@zburke This work falls into the category of "never urgent, always important". What does it take — either organizationally or technically — to get this happen? Now that I am once more running a "minimal" FOLIO backend in a VM on my desktop, and find myself allocation a ludicrous 20 Gb of memory to it, it's both ethically and practically intolerable that things should continue as they are.

MikeTaylor:
Pinging @wafschneider so he gets notified of further comments.

MikeTaylor:
@skomorokh and I are keen for this to get merged. @zburke, what hinders us?

zburke:
@MikeTaylor, @skomorokh: In talking with a few other folks including @skoczko and @mkuklis, our thinking around the best approach to FOLIO-3253 has shifted. This PR involves authenticating directly against mod-login (which has very few dependencies) but that endpoint just returns a token, nothing else. Thus, a side-effect is the need to then federate calls to (new) `/_self` endpoints in mod-users, mod-permissions, and mod-optional-thingies in order to hydrate the stripes user object. 

An alternative is to leverage the fact that mod-users-bl already handles mod-inventory (from when service points) as an optional dependency and could handle other mod-optional-thingies in the same way. That eliminates new backend work (no new `/_self` endpoints) and leaves the frontend much the same: 
1. in stripes-core, stash the JSON response from the login request in the redux store ([STCOR-649](https://issues.folio.org/browse/STCOR-649))
2. let other apps respond to the [`LOGIN` event](https://github.com/folio-org/stripes-core/blob/master/src/events.js#L2) and pull [`addReducer`](https://github.com/folio-org/stripes-core/blob/master/src/components/Root/Root.js#L155) from context to update the stripes object ([UISP-32](https://issues.folio.org/browse/UISP-32))
3. remove service-point handling from stripes-core (a breaking change because [`setServicePoints` and `setCurServicePoint` are exported](https://github.com/folio-org/stripes-core/blob/0b9390a3aa23803eb1b0108d6ea26cdb97f83926/index.js#L12), [STCOR-650](https://issues.folio.org/browse/STCOR-650))

I am actively working on (1) and (2).

MikeTaylor:
That all makes sense -- thank you!

mkuklis:
Related to [STCOR-650](https://issues.folio.org/browse/STCOR-650)) this is the place that will have to be refactored:

https://github.com/folio-org/ui-users/blob/56c938ffab7948da3c9a3084155d9a27db4ddb84/src/components/Wrappers/withServicePoints.js#L5-L11

We will have to think about where to put it. It could live in ui-servicepoints handler and instead of coupling it with the ui-users we could trigger an event via `HandlerManager`.

MikeTaylor:
That sounds right to me.

zburke:
This was only ever a POC. With the advent of [Eureka](https://folio-org.atlassian.net/wiki/spaces/PLATFORM/pages/193134643/Folio+Eureka+Platform+Overview) and Keycloak-based authentication, this specific PR is now unnecessary, if not entirely irrelevant.

PR #1107, opened 2021-08-19
UXPROD-3148: Add logic for loading the translations of library-defined controlled vocabularies

attia-alshareef:
## Description
Each FOLIO installation has lists of controlled vocabulary terms (for instance, “Formats” and “Resource types” in Inventory, “Patron Groups” and “Refund Reasons” in Users — see FOLIO-2802 for a more complete list).  In installations that use more than one language, these terms do not display translated values when the locale of the session is changed.

## Use case(s)
When the FOLIO session is started, we need to load the translations of the library-defined controlled vocabularies and merge them with the default stripes apps translations.

## Workflow
For the complete translation workflow, see [Workflow](https://github.com/folio-org/ui-translations/blob/master/README.md#workflow)

## Jira Related Issues Links
https://issues.folio.org/browse/UXPROD-3148 <br />
https://issues.folio.org/browse/FOLIO-3258 <br />
https://issues.folio.org/browse/UXPROD-515 <br />
https://issues.folio.org/browse/FOLIO-2802


src/loginServices.js
```diff
@@ -147,12 +147,30 @@ export function loadTranslations(store, locale, defaultTranslations = {}) {
...
 
+  // fetch library-defined controlled vocabularies translations, then merge it with stripes translations. see [UXPROD-3148] issue
   return fetch(translations[region] ? translations[region] : translations[parentLocale])
     .then((response) => {
       if (response.ok) {
         response.json().then((stripesTranslations) => {
-          store.dispatch(setTranslations(Object.assign(stripesTranslations, defaultTranslations)));
-          store.dispatch(setLocale(locale));
+          if (store.getState().okapi.token) {
+            fetch(`${store.getState().okapi.url}/translations?query=(localeCode==${locale})`,
```
julianladisch, line 156: I disagree with this proposed architecture.

The translations of controlled vocabulary should not be stored in a central translation module.
Instead the translations should live in the same back-end module where the non-translated controlled vocabulary lives.

This allows to have self-contained modules with as few inter-module dependencies as possible. In a micro-services-like architecture modules should be as independent as possible.

This matches the Technical Council's decision that back-end modules should handle i18n/l10n of messages: https://wiki.folio.org/display/DD/Handle+i18n+where+messages+are+generated

Does it have advantages that the translations of controlled vocabulary live in a different module?

attia-alshareef: There are many advantages that push us to save translations in a different module that I will explain in the translation app docs soon, but as a quick answer I will mention some of the risks and challenges that you will face if we are saving translations in the same back-end module:

1- It will require many changes for all back-and-modules that have controlled vocabulary. (Our solution does not require any changes at the level of back-and-modules)

2- Folio supports more than 26 languages ​​so far, which means that we are in the process of complicated multi-lingual schema and dramatic changes on the back-end architecture level are needed.

3- The amount of changes required on the stripes app to display the translated strings ​​instead of the original values ​​will be very large (our solution does not require this size of changes because all translations are fetched from one place and then merged with the default translations which are handled by react -intl )

About the Technical Council's decision that back-end modules should handle i18n/l10n of messages:
I think this is a different case because it is concerned with the error and validation messages generated by developers in the backend and I think they will be static translations (I don't know what is the approach the community follow-up to solve this issue).

PR #1123, opened 2021-10-18
STCOR-574 Conform branding image dimensions to suite different image sizes/ratios

JohnC-80:
It's in the visuals: 
landscape image:
branding-landscape
square image
small width:
image
large width:
image


zburke (approved):
Much nicer. Thanks!

PR #1127, opened 2021-10-28
STCOR-579 increase ky timeout from 10sec to 30sec

NikitaSedyx:
purpose: https://issues.folio.org/browse/STCOR-579
we have some long requests and 10secs is not enough to complete them when stripes-connect can wait for minutes

approach: change default timeout from 10sec to 30sec

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Change history for stripes-core
 
+## 8.1 IN PROGRESS
+
+* Use correct `css-loader` syntax. STCOR-577.
+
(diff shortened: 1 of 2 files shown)


MikeTaylor:
I did say why I thought it was a bad change:

> If an application is unusual, then that application should do the necessary unusual thing to allow for that -- in this case, specifying a non-default timeout.

If you are confident that a typical application of `okapiKy` needs a thirty-second timeout rather than ten, then fine. But if it's just that _your_ application happens to need that, then changing default behaviour is not the way to go.

PR #1132, opened 2021-11-09
bump minor to 8.1

zburke:
New features should land on a new minor version since the last release
so that feature work and bug-fix work can be correctly separated.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@folio/stripes-core",
-  "version": "8.0.0",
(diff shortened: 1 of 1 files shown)


zburke:
> This pull request removes 5 and adds 3 tests. Note that renamed tests count towards both.

This pull request _does no such thing_. It changes one character of the `package.json` version; that is all. I have no idea what this comment is going on about. 

@ankitasen-ubmainz, any ideas where this incorrect info is coming from?

PR #1135, opened 2021-11-23
STCOR-582 Your password changed confirmation page > Continue to Login URL goes to an error page

BogdanDenis:
## Description
Fix "404" error message when user changes password and logs in. This happens because after changing password user is redirected to `/login` route and after successful login they are not redirected. But `/login` route doesn't exist when a user is logged in.

## Issues
[STCOR-582](https://issues.folio.org/browse/STCOR-582)

### DIFF
--- a/test/bigtest/tests/useCustomFields-test.js
+++ b/test/bigtest/tests/useCustomFields-test.js
@@ -12,6 +12,13 @@ const setupWithApp = (App, title) => setupApplication({
   disableAuth: false, // Can't skip login flow bc we need to fetch module info
   modules: [{
     type: 'app',
+    name: '@folio/ui-login',
+    displayName: 'login.title',
+    route: '/login',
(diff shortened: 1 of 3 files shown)


test/bigtest/tests/useCustomFields-test.js
```diff
@@ -57,7 +64,7 @@ const doLogout = async function () {
   await i.clickLogout();
 };
 
-describe('useCustomFields hook', () => {
+describe.only('useCustomFields hook', () => {
```
Vladyslav-Velytskyi, line 67: Are you sure that it should be the only test?

BogdanDenis: Thanks for noticing!

zburke (approved):
Thank you for fixing this!

PR #1147, opened 2021-12-20
Wrap modules in Suspense to handle lazy loading

mkuklis:
This PR wraps UI modules in `Suspense` in order to handle lazy loading introduced in:

https://github.com/folio-org/stripes-webpack/pull/42

lazy-loading


zburke (approved):
Sweeeeeeet! 👏 👏 👏

PR #1148, opened 2021-12-21
the most humble offering possible to the lint gods

zburke:
And there were developers sitting at their laptops nearby, keeping watch
over their code by night. Just then, an angel of lint stood before them,
and the glory of lint shone all around them, and they were terrified.
But the angel said unto them, "Do not be afraid! For behold! I bring you
good news of great joy that will be for all the people: Today in the
module of stripes-core, I discovered one insignificant space with no
semantic value and no visual clutter, but it really bugs me anyway. And
this will be a sign to you: you will find two complaints on every PR
about it, even when they have nothing to do with this file.

And suddenly there appeared around every stripes-core PR a great
multitude of lint complaints, saying:

```
Multiple spaces found before '/\.\/.*\.js$/'
```

When the lint angels had left them and gone into heaven, the developers
said to one another, "One lousy space? Was that guy serious?"

### DIFF
--- a/test/bigtest/network/index.js
+++ b/test/bigtest/network/index.js
@@ -6,7 +6,7 @@ import camelCase from 'lodash/camelCase';
 // of the filename, otherwise webpack will output duplicate filepaths
 //
 // see: https://github.com/webpack/webpack/issues/12087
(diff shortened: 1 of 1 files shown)


mkuklis (approved):
👏

PR #1152, opened 2022-01-07
STCOR-480 STCOR-525 restore okapi version retrieval

zburke:
`/_/version` was available, and then it wasn't, and now it is. Now
shake it all about, do the hokey pokey and turn yourself around, if you
aren't already too turned around.

This started merely as an attempt to deal with a warning about a bad
prop-types setting (`PropTypes.okapi` is not a thing) but as I got in
there and started looking around to figure out the correct prop-type,
and then to add a comment referencing the ticket when the okapi-version
info was removed (#1033), I discovered it had been restored.

Sorted the stripes-shape prop-types for good measure.

Refs [STCOR-480](https://issues.folio.org/browse/STCOR-480), [STCOR-525](https://issues.folio.org/browse/STCOR-525), [OKAPI-917](https://issues.folio.org/browse/OKAPI-917).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 * Use correct `css-loader` syntax. STCOR-577.
 * Your password changed confirmation page > Continue to Login URL goes to an error page. Fixes STCOR-582.
 * Dependency cleanup. Refs STCOR-584.
+* Restore okapi version display. Refs STCOR-480, STCOR-525.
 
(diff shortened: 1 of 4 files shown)


zburke:
Scratch this; it doesn't correctly dispatch the modules and interfaces. I think just reverting #1033, or deriving the okapi version as described by @doytch on that PR is a better option.

PR #1153, opened 2022-01-07
Revert "STCOR-525 Removed fetch of protected okapi /version endpoint"

zburke:
Reverts folio-org/stripes-core#1033


adamdickmeiss (commented):
It looks good, but know to little JS to judge.

mkuklis (approved):

PR #1157, opened 2022-02-02
STRIPES-731: Apps > Settings > Move focus to Settings pane

Javokhir12:
## Purpose
- If a user selects `Settings` from the `Apps` drop-down then move focus to the `Settings` pane.

## Approach
- Add `paneTitleRef` prop to `Pane` component to manage focus automatically.

## Refs
- https://issues.folio.org/browse/STRIPES-731

### DIFF
--- a/src/components/Settings/Settings.js
+++ b/src/components/Settings/Settings.js
@@ -47,6 +47,7 @@ class Settings extends React.Component {
   constructor(props) {
     super(props);
 
+    this.paneTitleRef = createRef();
     const { stripes, modules } = props;
     const settingsModules = modules.settings || [];
 
@@ -75,8 +76,6 @@ class Settings extends React.Component {
       this.paneTitleRef.current.focus();
     }
   }
-
-  paneTitleRef = createRef();
 
   render() {
     const { stripes, location, intl: { formatMessage } } = this.props;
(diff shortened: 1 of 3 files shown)


Javokhir12:
Hi @mkuklis, @zburke , please take a look and  merge.

src/components/Settings/Settings.js
```diff
@@ -70,6 +70,14 @@ class Settings extends React.Component {
       });
   }
 
+  componentDidMount() {
+    if (this.paneTitleRef.current) {
+      this.paneTitleRef.current.focus();
+    }
+  }
+
+  paneTitleRef = createRef();
```
mkuklis, line 79: @Javokhir12 since we already have a constructor here would you mind moving it there? It's just easier to reason about.

mkuklis (approved):

src/components/Settings/Settings.js
Javokhir12: @mkuklis good point, it is done.

JohnC-80 (approved):

zburke (approved):

NikitaSedyx:
looks like code in this MR https://github.com/folio-org/stripes-smart-components/pull/1179/files should handle all cases, with this one we'll have double focus

NikitaSedyx:
oh, sorry, I see it's focus to the first pane in settings

PR #1159, opened 2022-02-03
STCOR-587 Apply h3 tag for System Information

OleksandrHladchenko1:
Ref: https://issues.folio.org/browse/UIU-2037


OleksandrHladchenko1:
Hello, @zburke!  In scope of my task I need to replace DIV for System Information with H3 tag. So, in NavListSection I added tag prop with value h3. It depends on approving this PR(https://github.com/folio-org/stripes-components/pull/1716) Can we apply this change for my task?

zburke:
Replaced by #1161

PR #1160, opened 2022-02-08
STCOR-586: Apps > Settings > Move focus to Settings pane

zburke:
## Purpose
- If a user selects `Settings` from the `Apps` drop-down then move focus to the `Settings` pane.

## Approach
- Add `paneTitleRef` prop to `Pane` component to manage focus automatically.

## Refs
- https://issues.folio.org/browse/STRIPES-731

Replaces #1157

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@ node_modules
 !test/webpack/node_modules/@folio/app2/node_modules/
 !test/webpack/node_modules/
 npm-debug.log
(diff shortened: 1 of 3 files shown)


zburke:
Merged #1157 instead. It appears that creating a PR based on a branch which has also been pushed on a fork allows the SonarCloud check to run and thus for both PRs to move to passing.

PR #1163, opened 2022-02-15
STCOR-588 prune ancient and unnecessary react-intl polyfills

zburke:
Long ago (#876) we added a bunch of polyfills to stripes-core because
react-intl stopped including them and our e2e tests need them. But those
e2e tests are gone, baby, gone, so there is no reason to balloon our
bundle with a bunch of cruft that no contemporary browser requires.
(RelativeTimeFormat has been in Chrome since 2018 and Safari since 2020; 
PluralRules since 2017 and 2019 respectively, and DisplayNames since 2020 
and early 2021 respectively.)

h/t @skomorokh for noticing this disk pollution.

Refs [STCOR-588](https://issues.folio.org/browse/STCOR-588)

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,9 +68,6 @@
   },
   "dependencies": {
     "@apollo/client": "^3.2.1",
-    "@formatjs/intl-displaynames": "^2.2.3",
-    "@formatjs/intl-pluralrules": "^2.2.1",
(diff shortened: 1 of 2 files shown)


mkuklis (approved):
Thanks for cleaning this up!

PR #1167, opened 2022-02-21
compile translations before running tests

zburke:
Use the same workflow order as on regular PRs, compiling translations
before running tests for a quieter console.


zburke:
I'm not sure why I pushed this up as a separate branch; it's completely incorporated in #1165.

PR #1169, opened 2022-02-28
STCOR-561 Configurable mechanism to warn of stale frontend

skomorokh:
This is controlled via  a new stripes config property:

staleBundleWarning - When present, this optional object will enable a periodic check to let users know when to reload the page because the bundle has changed. It takes the following properties:

* header - instead of the default to fetch and compare the body of a file (perhaps containing a build or commit id), use HTTP HEAD and the specified header
* interval - how many minutes to wait between checks, defaults to 15
* path - path to fetch, this is the only required parameter

Refs [STCOR-561](https://issues.folio.org/browse/STCOR-561)


mkuklis (approved):
Nicely done 👏

src/components/StaleBundleWarning.js
```diff
@@ -0,0 +1,51 @@
...
+      if (typeof config?.path !== 'string') return null;
+      try {
+        if (typeof config?.header === 'string') {
+          const res = await ky(config.path, { method: 'head' });
+          return res.headers.get(config.header);
+        } else {
+          return await ky.get(config.path).text();
+        }
+      } catch (e) {
+        console.warn(`'Error checking for new bundle ${e}`); // eslint-disable-line no-console
```
zburke, line 26: Pedantic: `'Error` should be `Error`

src/components/StaleBundleWarning.js
```diff
@@ -0,0 +1,51 @@
...
+    staleTime: refetchInterval,
+    refetchInterval,
+    enabled: !stale,
+  });
+  useEffect(() => {
+    if (!previous) {
+      setPrevious(query.data);
+      return;
+    }
+    if (previous !== query.data) setStale(true);
```
zburke, line 39: Be like Perl: always put braces around conditionals. In fact, I can't believe we don't have a lint rule for this!

src/components/StaleBundleWarning.js
```diff
@@ -0,0 +1,51 @@
...
+    queryKey: ['StaleBundleWarning'],
+    queryFn: async () => {
+      if (typeof config?.path !== 'string') return null;
+      try {
+        if (typeof config?.header === 'string') {
+          const res = await ky(config.path, { method: 'head' });
+          return res.headers.get(config.header);
+        } else {
+          return await ky.get(config.path).text();
+        }
```
zburke, line 24: It would be great if you provided an example config object in a header comment to this component, explaining what it does and how to use `stripes.config.js` to use it. You provided a nice [explanation and screenshot on #slack](https://folio-project.slack.com/archives/CAN13SWBF/p1646060923092559), but that information will be harder to find when this is less front-of-mind than it is right now.

zburke (changes requested):
This is a nice feature and the code looks great but we could really use a bit of documentation about why we need this, and test cases demonstrating the various config options.

src/components/StaleBundleWarning.js
mkuklis: I would also say add an extra line before that `if` too :)

skomorokh: We do have a lint rule inherited from the airbnb ruleset we based our config on. It disallows multi-line if without braces but allows oneliners. I kinda like that, after all, do we put braces on our ternaries?

I did add some more whitespace though, it was a bit squished. Thanks!

src/components/StaleBundleWarning.js
skomorokh: I was intending to put the docs at the top of the PR in the dev guide where we have the other config options but they're in a really confusing place there and discussed on Slack, we really need a proper document for Stripes deployment rather than development which is where this really belongs. And an example would be nice indeed.

So for now it can live inline on the file as you suggest.

src/components/StaleBundleWarning.js
zburke: I feel like our lint rules let you down here, so that's definitely not on you. Filed [ESCONF-15](https://issues.folio.org/browse/ESCONF-15) to make braces mandatory so we can stop having debates like this. If our lint rules allow it, nobody should ever be allowed to comment on it :)

zburke: I closed ESCONF-15; there is not widespread agreement around brace-formatting and besides, the _whole point_ of lint-rules is to avoid these kinds of debates on PRs. IOW, if lint didn't complain on this PR, then @zburke shouldn't be complaining on this PR. Srsly, some people 🙄 .

zburke (approved):
I wrote some tests :)

PR #1170, opened 2022-02-28
STCOR-592 shout out to STCOR-592 for the next passerby

zburke:
[STCOR-592](https://issues.folio.org/browse/STCOR-592)


skomorokh:
Good idea!

PR #1173, opened 2022-03-16
STCOR-576: Create/Reset password page > Display password requirements on page

mariiaschv:
## Description
Display password requirements on Create/Reset password page.

Created a custom hook `usePasswordRules` and had to create a new Ky instance there instead of using `useOkapiKy` because on 
Create/Reset password page the value of `token` in `useStripes` is not defined, so the request fails with the `401 Invalid token in handleAuthorize` error.

## Screenshots
chrome_VXyvpmoHnj

## Issues
[STCOR-576](https://issues.folio.org/browse/STCOR-576)


src/components/CreateResetPassword/components/PasswordRequirementsField/PasswordRequirementsField.js
```diff
@@ -0,0 +1,101 @@
...
+import styles from './PasswordRequirementsField.css';
+
+const passwordRequirementsNames = [
+  'password_length',
+  'numeric_symbol',
+  'special_character',
+  'alphabetical_letters',
+];
+
+class PasswordRequirementsField extends React.Component {
```
BogdanDenis, line 15: Can we not use `PasswordValidationField` from `stripes-core` here? It seems that it does exactly what is needed in the story

mariiaschv: No, unfortunately it cannot be used here for a number of reasons.

Svitlana-Korol (approved):

src/components/CreateResetPassword/components/PasswordRequirementsList/PasswordRequirementsList.js
```diff
@@ -0,0 +1,101 @@
...
+        rule.description = `Must ${splittedRuleDescription[1]}`;
+
+        return rule;
+      });
+  }
+
+  render() {
+    const { passwordValue } = this.props;
+
+    const compareNewPasswordToRules = () => {
```
NikitaSedyx, line 66: why is it created in render function? additionally why `unfulfilledRules` is an instance property?

for me this function is redundant

src/components/CreateResetPassword/components/PasswordRequirementsList/PasswordRequirementsList.js
```diff
@@ -0,0 +1,101 @@
...
+
+const passwordRequirementsNames = [
+  'password_length',
+  'numeric_symbol',
+  'special_character',
+  'alphabetical_letters',
+];
+
+class PasswordRequirementsList extends React.Component {
+  static manifest = Object.freeze({
```
NikitaSedyx, line 16: seems react-query can be used

src/components/CreateResetPassword/components/PasswordRequirementsList/PasswordRequirementsList.js
```diff
@@ -0,0 +1,101 @@
...
+import styles from './PasswordRequirementsList.css';
+
+const passwordRequirementsNames = [
+  'password_length',
+  'numeric_symbol',
+  'special_character',
+  'alphabetical_letters',
+];
+
+class PasswordRequirementsList extends React.Component {
```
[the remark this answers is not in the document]
mariiaschv: I've converted the component to be functional, used react-query and gotten rid of `compareNewPasswordToRules` function, thanks!

src/components/CreateResetPassword/components/PasswordRequirementsList/PasswordRequirementsList.js
```diff
@@ -0,0 +1,84 @@
...
+  const { rules } = usePasswordRules();
+
+  useEffect(() => {
+    if (requiredRules.length <= 0 && rules?.length > 0) {
+      const requiredRulesSet = rules
+        .filter(rule => passwordRequirementsNames.includes(rule.name))
+        .map(rule => {
+          const splittedRuleDescription = rule.description.split(' must ');
+
+          rule.description = `Must ${splittedRuleDescription[1]}`;
```
BogdanDenis, line 33: Please add a new translation with this message

BogdanDenis (changes requested):

src/components/CreateResetPassword/components/PasswordRequirementsList/PasswordRequirementsList.js
mariiaschv: Fixed, thank you!

BogdanDenis (approved):

src/components/CreateResetPassword/components/PasswordRequirementsList/PasswordRequirementsList.js
```diff
@@ -0,0 +1,94 @@
...
+  useEffect(() => {
+    if (requiredRules.length <= 0 && rules?.length > 0) {
+      const requiredRulesSet = rules
+        .filter(rule => passwordRequirementsNames.includes(rule.name))
+        .map(rule => {
+          const splittedRuleDescription = rule.description.split(' must ');
+
+          rule.description = intl.formatMessage(
+            { id: 'stripes-core.createResetPassword.ruleTemplate' },
+            { description: splittedRuleDescription[1] },
```
zburke, line 38: It would be really helpful to have some comments in here explaining what's going on. Without an example, splitting on `must` and passing off a substring as the description is totally opaque. I think rule descriptions tend to look like 

> The password must blah blah blah

and this just stripes off "The password must". So, pretty simple, but not that obvious.

src/components/CreateResetPassword/components/PasswordRequirementsList/queries/usePasswordRules.js
```diff
@@ -0,0 +1,35 @@
...
+      ]
+    },
+    retry: 0,
+    timeout: 30000,
+  });
+
+  const { data } = useQuery(
+    ['requirements-list'],
+    async () => {
+      return kyInstance.get('tenant/rules').json();
```
zburke, line 26: Don't we need a `limit` clause on the query like `tenant/rules?limit=100` in case there are more than 10 rules? There are at least 11 in folio-snapshot and bugfest at present.

zburke (changes requested):
This is nice and clean, but I think we need a `limit` clause on the rules query to override the default limit (10). 100 is likely fine here, I mean, I hope no tenant has more than 100 rules! 😬

zburke (approved):
LGTM!

PR #1174, opened 2022-03-21
STCOR-596 Update FOLIO Home icon

JohnC-80:
See https://issues.folio.org/browse/UX-473 (linked to STCOR-596)
Before:
image

After:
image


ryandberger:
Looks like there's an app icon test that needs to be updated to pass.

zburke (commented):
No objections to the logo change, of course, but can you explain why the XML declaration was dropped? If anything, I would expect an "update the svg" change to add or strengthen the metadata, e.g. by adding a DOCTYPE/DTD declaration, rather than to drop existing metadata.

JohnC-80:
Yep.. thanks for the comments -  let me first remember to 'wash' my svg before use... I can see some other garbage from the editor ended up getting exported.

zburke (approved):

PR #1175, opened 2022-03-24
STCOR-597 remove react-hot-loader cruft

zburke:
Whooops, it wasn't even a dev-dep but we still had it in our source.
It's no longer necessary in any case since react-refresh is totally
self-contained within stripes-webpack.

Refs [STCOR-597](https://issues.folio.org/browse/STCOR-597), [STRIPES-725](https://issues.folio.org/browse/STRIPES-725)


mkuklis (approved):
🙇

PR #1181, opened 2022-04-12
STCOR-599: Display different message when user attempts to access the set password link when still logged in

mariiaschv:
## Purpose
Display different message when user attempts to access the set password link when still logged in.

## Screenshots
chrome_Y8znjD5Niq

## Issues
[STCOR-599](https://issues.folio.org/browse/STCOR-599)


zburke (changes requested):
I don't think this is the right fix. It changes the message for _all_ bad requests. e.g. somebody who accesses a URL like https://folio-snapshot.dev.folio.org/asdfasdfasdf, would see "Please log out..." instead of "The requested URL was not found." 

Maybe this work can be done directly in `./src/components/CreateResetPassword/`, or maybe `./src/RootWithIntl.js` where that component is invoked, in order to correctly limit its scope?

mariiaschv:
> I don't think this is the right fix. It changes the message for _all_ bad requests. e.g. somebody who accesses a URL like https://folio-snapshot.dev.folio.org/asdfasdfasdf, would see "Please log out..." instead of "The requested URL was not found."
> 
> Maybe this work can be done directly in `./src/components/CreateResetPassword/`, or maybe `./src/RootWithIntl.js` where that component is invoked, in order to correctly limit its scope?

@zburke thank you! I wasn't attentive when implemented this. Fixed it and also attached gif to the description of the PR.

zburke (approved):
That was quick! :shipit:

BogdanDenis (approved):

PR #1185, opened 2022-04-14
STCOR-600 Set Mirage timing to 0

JohnC-80:
This will result in an immediate response from the server - in most cases, it speeds up the tests dramatically...

Refs [STCOR-600](https://issues.folio.org/browse/STCOR-600)


zburke (commented):
I'm all for the change in `setup-application`. 

Can you explain the "why" behind the change in how moment locales are loaded in `loginServices`?

zburke:
@JohnC-80: ping: I'm all for the 0-mirage-timing change in setup-application. Can you explain the "why" behind the change in how moment locales are loaded in loginServices? It seems unrelated to this ticket, and potentially looks like an outright bug.

zburke:
[mirage docs say the set timing delay to 0 "during testing"](https://miragejs.com/api/classes/server/#timing) and 400 ms "during development" but of course don't tell you how you can flip between testing and development. I'm not sure how, but I think we already have this in place. 

Do you agree, @JohnC-80 , and if so can you close this and [STCOR-600](https://issues.folio.org/browse/STCOR-600) as well?

PR #1188, opened 2022-04-26
STCOR-618 configure jest/rtl

zburke:
Refs [STCOR-618](https://issues.folio.org/browse/STCOR-618)


mkuklis (approved):
Thank you for setting this up!

PR #1189, opened 2022-04-26
STCOR-615 export queryLimit returning an int

zburke:
Export a function, `queryLimit`, to provide a default value for API
query limit clauses UI objects can import.

Refs [STCOR-615](https://issues.folio.org/browse/STCOR-615)


zburke:
(Must be merged after #1188)

src/queryLimit.test.js
```diff
@@ -0,0 +1,7 @@
+import queryLimit from './queryLimit';
+
+describe('queryLimit', () => {
```
ryandberger, line 3: Not for here, but noting that we should ignore these globals such as `describe`

zburke: @ryandberger , I think #1188 will take care of that.

mkuklis (approved):

PR #1190, opened 2022-04-28
STCOR-619 Add message user cannot access app

AndreiShumski2:
https://issues.folio.org/browse/STCOR-619
Screenshot 2022-04-28 at 15 02 18


zburke (approved):
This is better than just `null` for sure, but the message is barely noticeable. Could we wrap it in a `<Headline tag="h2">` or something similar?

PR #1193, opened 2022-05-09
STCOR-620 separate route- and component-based tests

zburke:
Route-based tests that leverage BigTest to mount and component-based
tests that mount directly don't play nice together. The problem appears
to be with how the DOM node under which a component is mounted is
cleaned up, or not, and a fight for control of this node ensues. For
reasons unclear, even if both native code and BigTest are given the same
`id` value, they don't use the same element; instead they each create
their own element with the same `id`.

The approach taken here is to run all the component-based tests first
(these are present in `src/components/...`) followed by all the
route-based tests (these are present in `test/bigtest/tests/...`).
Hopefully, running strictly in series prevents them from stepping on
each other's toes.

Refs [STCOR-620](https://issues.folio.org/browse/STCOR-620)


zburke:
Ugh, so many failing tests in CI that ran smoothly locally :( Sorry, friends. I'll have to take another swing at this.

JohnC-80 (approved):
Looks really good! Thanks for all the work on this @zburke !

PR #1194, opened 2022-05-10
STCOR-619 no permission screen

AndreiShumski2:
## Purpose:
As a user with limited permissions or an administrator troubleshooting,
I want an indicator that I cannot access something due to permissions
so that I can better diagnose an issue as missing permissions, rather than a larger issue in FOLIO.

### Screenshot
Screenshot 2022-05-10 at 14 38 52


zburke:
NB: This replaces #1190 which was reverted in #1192.

AndreiShumski2:
@zburke could you please take a look? Thanks 🙏

NikitaSedyx (changes requested):
changes are required because of https://github.com/folio-org/stripes-core/pull/1202

AndreiShumski2:
> 

It definitly doesn't work for Settings. But if user goes to Settings, it'll display Settings page without any setting in the list (in case user has no settings permissions). Not sure that's critical. The main concern was the blank screen.

zburke:
This doesn't work for me:
```
Uncaught TypeError: Cannot read properties of undefined (reading 'displayName')
    at renderToggle (AppContextDropdown.js:32:26)
    at Popdown (Popdown.js:221:19)
    ...
```
which corresponds to 
```
 {selectedApp.displayName}
```
`selectedApp` comes in via props; logging it out during render shows me 
```
AppContextDropdown.js:22 AppContextDropdown {id: 'clickable-orders-module' ...
AppContextDropdown.js:22 AppContextDropdown {id: 'clickable-orders-module', ...
AppContextDropdown.js:22 AppContextDropdown undefined
```

src/components/MainNav/CurrentApp/AppContextDropdown.js
```diff
@@ -29,7 +29,7 @@ class AppContextDropdown extends React.Component {
         open={open}
         label={
           <>
-            {selectedApp.displayName}
+            {selectedApp?.displayName}
```
zburke, line 32: Thanks! I wonder if this was a new glitch introduced by React.lazy, since previously all apps would be loaded, but that is no longer the case. (CC: @mkuklis)

zburke (approved):
LGTM; thanks for sticking with it!

PR #1202, opened 2022-05-27
STCOR-626: Cache getModule calls

mkuklis:
https://issues.folio.org/browse/STCOR-626

It looks like the `getModule` calls are currently executed during a regular render (which happens on every route change):

https://github.com/folio-org/stripes-core/blob/97bcff4029aeec39ca9ea5b69bf6192d6458f2f2/src/moduleRoutes.js#L75

https://github.com/folio-org/stripes-core/blob/97bcff4029aeec39ca9ea5b69bf6192d6458f2f2/src/Pluggable.js#L25

This was working fine till now but I think only by accident because the calls are cached in the lower level (stripes-webpack).

Unfortunately, this breaks with the lazy-loading approach and it should be addressed in a similar fashion to what happens in `Settings` (single initialization in the constructor):

https://github.com/folio-org/stripes-core/blob/f33c8b0dc3d21a5ccd2a537956d6f8942d229120/src/components/Settings/Settings.js#L64

 @NikitaSedyx and @zburke I believe this should address the two remaining issues:

1. Each module is fully remounted even when the nested routing (related only to a given module) changes
2. Plugins not working correctly

@NikitaSedyx I would appreciate it if you could give it a try with the modules you are working on and see if these two issues go away.

I tested it with ui-inventory and things seem to work as expected.

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,6 +46,7 @@
     "@babel/eslint-parser": "^7.15.0",
     "@bigtest/convergence": "^0.10.0",
     "@bigtest/interactor": "^0.7.2",
+    "@bigtest/mirage": "^0.0.1",
     "@bigtest/mocha": "^0.5.1",
     "@bigtest/react": "^0.1.2",
     "@folio/eslint-config-stripes": "^6.0.0",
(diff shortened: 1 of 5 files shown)


mkuklis:
The tests fail but they seem to be related to a missing `@bigtest/mirage`.

NikitaSedyx (approved):
I've posted in slack the same idea :) Tjet team tested our modules + plugins locally, seems it's working, thank you @mkuklis

PR #1205, opened 2022-06-02
STCOR-629 provide @bigtest/mirage dev-dep

zburke:
[Tests use it](https://github.com/folio-org/stripes-core/blob/be772d924fca97e2189e954a1c9075c629b6f645/test/bigtest/network/start.js#L24), so we gotta provide it. This was probably erroneously
moved into stripes-webpack, where [it isn't used and was finally removed](https://github.com/folio-org/stripes-webpack/pull/61),
which surfaced the fact that it was missing here.

Refs [STCOR-629](https://issues.folio.org/browse/STCOR-629)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 * Support testing with Jest/RTL. Refs STCOR-618.
 * Export `queryLimit` to provide a default API query limit. Refs STCOR-615.
 * Update NodeJS to v16 in GitHub Actions. Refs STCOR-623.
+* Add `@bigtest/miragejs` dev-dep since tests rely on it. Refs STCOR-629.
 
 ## [8.1.0](https://github.com/folio-org/stripes-core/tree/v8.1.0) (2022-02-11)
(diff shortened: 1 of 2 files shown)


EthanFreestone (approved):
I see the dev-dep on master, thank you @zburke and apologies for the pain point.

zburke:
@EthanFreestone, in fact this was restored it in #1202 which means all that's happening here is a CHANGELOG update. (#1202 was the first PR we merged in @folio/stripes-core after removing `@bigtest/mirage` in `@folio/stripes-webpack`, and I even wrote the commit comment, which makes me wonder how I forgot about doing that 🤦.) 

Regardless, note that it was restored here as _dev-dep_ and not a direct-dep as it was in `@folio/stripes-webpack`. This means it will not be installed by apps that pull in `@folio/stripes-core`, so if your app is leveraging test configs from `./test/...` (a) it should stop because importing by private-path is a bad practice and (b) it will need to provide `@bigtest/mirage` as its own dev-dep.

zburke:
No real value in updating the CHANGELOG here as this is a dev-dep only that was changed in a separate PR, i.e. since that change didn't affect the PUBLIC API, we don't need public documentation.

PR #1206, opened 2022-06-06
STCOR-631 Added useCallout hook

BogdanDenis:
## Description
Added `useCallout` hook for easier mocking in tests

## Issues
[STCOR-631](https://issues.folio.org/browse/STCOR-631)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 * Support testing with Jest/RTL. Refs STCOR-618.
 * Export `queryLimit` to provide a default API query limit. Refs STCOR-615.
 * Update NodeJS to v16 in GitHub Actions. Refs STCOR-623.
+* Provide `useCallout` hook. Refs STCOR-631.
 
 ## [8.1.0](https://github.com/folio-org/stripes-core/tree/v8.1.0) (2022-02-11)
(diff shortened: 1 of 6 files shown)


zburke (approved):
Good thinking! Related questions that don't need to be sewn up here: 

* Do you think there should be a `./src/hooks` directory where we keep all these context-related hooks, or do you think keeping them next to their context components is better? (I'm not sure about this one.)
* Do you think all context providers should be paired with hooks and exported at the top level? (I feel like this is probably a good idea.)

BogdanDenis:
> Good thinking! Related questions that don't need to be sewn up here:
> 
> * Do you think there should be a `./src/hooks` directory where we keep all these context-related hooks, or do you think keeping them next to their context components is better? (I'm not sure about this one.)
> * Do you think all context providers should be paired with hooks and exported at the top level? (I feel like this is probably a good idea.)

1. That's a good point, it's great that you brought you up, because in many cases we export two or three things from one context file - the context itself, a hook and a HOC. So I'll change my code to follow this pattern
2. Yes, if a context is meant to be used outside of `stripes-core` I think it's a good way to do export them

zburke (approved):

PR #1213, opened 2022-06-09
STCOR-635 wrap pluggable in suspense

zburke:
Wrap Pluggable's return value in Suspense in order to avoid whole-tree
re-renders, since the whole tree from the components up to the nearest
`<Suspense>` is removed and replaced with its fallback. For us, this is
particularly a problem when a plugin is part of a dropdown menu that
uses calculations to set its position. Because the entire tree will be
unmounted, the calculations all return zero, resulting in incorrect
placement with zero offset from the let margin ([STRWEB-53](https://issues.folio.org/browse/STRWEB-53)).

h/t @bogdandenis and @mkuklis for figuring this all out.

Refs [STCOR-635](https://issues.folio.org/browse/STCOR-635)

### DIFF
--- a/src/Pluggable.js
+++ b/src/Pluggable.js
@@ -2,7 +2,7 @@ import React, { useMemo } from 'react';
 import PropTypes from 'prop-types';
 import { modules } from 'stripes-config';
 import { withStripes } from './StripesContext';
-import { ModuleHierarchyProvider } from './components';
+import { LoadingView, ModuleHierarchyProvider } from './components';
 
 const Pluggable = (props) => {
   const plugins = modules.plugin || [];
@@ -37,7 +37,9 @@ const Pluggable = (props) => {
   if (cachedPlugins.length) {
     return cachedPlugins.map(({ plugin, Child }) => (
       <ModuleHierarchyProvider module={plugin}>
-        <Child {...props} actAs="plugin" />
+        <Suspense fallback={<LoadingView />}>
+          <Child {...props} actAs="plugin" />
+        </Suspense>
       </ModuleHierarchyProvider>
     ));
   }
@@ -47,7 +49,11 @@ const Pluggable = (props) => {
     // eslint-disable-next-line no-console
     console.error(`<Pluggable type="${props.type}"> has ${props.children.length} children, can only return one`);
   }
-  return props.children;
+  return (
+    <Suspense fallback={<LoadingView />}>
+      {props.children}
+    </Suspense>
+  );
 };
 
 Pluggable.propTypes = {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 * Update NodeJS to v16 in GitHub Actions. Refs STCOR-623.
 * Provide `useCallout` hook. Refs STCOR-631.
 * Add message to indicate user cannot access app/record. Refs STCOR-619.
+* Wrapp `<Pluggable>` in `<Suspense>` to avoid whole-tree re-render. Refs STCOR-635.
 
 ## [8.1.0](https://github.com/folio-org/stripes-core/tree/v8.1.0) (2022-02-11)
 [Full Changelog](https://github.com/folio-org/stripes-core/compare/v8.0.0...v8.1.0)


mkuklis (approved):

zburke:
This is a nice idea, but it doesn't work. In a ui-inventory build with resolutions to this branch and to [stripes-webpack with lazy turned back off for plugins](https://github.com/folio-org/stripes-webpack/pull/67), `stripes` isn't making its way through the hierarchy from `StripesContext` in [stripesConnect](https://github.com/folio-org/stripes-core/blob/976a95c2161c6f737a80e7d799301e1c82eb032c/src/stripesConnect.js#L14): 
```
Uncaught TypeError: Cannot read properties of undefined (reading 'connect')
    at new ConnectedComponent (stripesConnect.js:14:47)
    at constructClassInstance (react-dom.development.js:12716:1)
    ...
```
Interestingly, I also had to add dev-deps for stripes-form and stripes-final-form. 

@mkuklis , @BogdanDenis , any ideas what could be going on?

BogdanDenis:
@zburke hmm, I get a completely different error
```
Cannot access '__WEBPACK_DEFAULT_EXPORT__' before initialization
```
I'll investigate this more today

BogdanDenis (changes requested):

BogdanDenis (approved):

src/Pluggable.js
```diff
@@ -1,8 +1,9 @@
-import React, { useMemo } from 'react';
+import React, { Suspense, useMemo } from 'react';
 import PropTypes from 'prop-types';
 import { modules } from 'stripes-config';
 import { withStripes } from './StripesContext';
 import { ModuleHierarchyProvider } from './components';
+import { LoadingView } from '@folio/stripes-components';
```
BogdanDenis, line 6:
```suggestion
import { Loading } from '@folio/stripes-components';
```

src/Pluggable.js
```diff
@@ -37,7 +38,9 @@ const Pluggable = (props) => {
   if (cachedPlugins.length) {
     return cachedPlugins.map(({ plugin, Child }) => (
       <ModuleHierarchyProvider module={plugin}>
-        <Child {...props} actAs="plugin" />
+        <Suspense fallback={<LoadingView />}>
```
BogdanDenis, line 41:
```suggestion
        <Suspense fallback={<Loading />}>
```

src/Pluggable.js
```diff
@@ -47,7 +50,11 @@ const Pluggable = (props) => {
     // eslint-disable-next-line no-console
     console.error(`<Pluggable type="${props.type}"> has ${props.children.length} children, can only return one`);
   }
-  return props.children;
+  return (
+    <Suspense fallback={<LoadingView />}>
```
BogdanDenis, line 54:
```suggestion
    <Suspense fallback={<Loading />}>
```

BogdanDenis (changes requested):
Should use `<Loading>` instead - `<LoadingView>` will render a `<Paneset>` and this can potentially break pane resizing if there's a parent `<Paneset>`. Just spent a few hours debugging an issue that was caused by nested panesets, even though in theory they are valid.

I know it was I who suggested using `<LoadingView>` - so that's me fixing my mistake here

zburke:
@BogdanDenis, we turned off lazy-loading of modules for Morning Glory in stripes-webpack in https://github.com/folio-org/stripes-webpack/pull/69, which should turn `<Suspense>` into a do-nothing component because there are no separate chunks to wait for. IOW, we _do_ need to `<Loading>` instead of `<LoadingView>` for the nested-paneset reason you identify, but it doesn't matter now, right? I guess I'm just making sure your stripes-webpack is up to date :)

BogdanDenis:
> @BogdanDenis, we turned off lazy-loading of modules for Morning Glory in stripes-webpack in [folio-org/stripes-webpack#69](https://github.com/folio-org/stripes-webpack/pull/69), which should turn `<Suspense>` into a do-nothing component because there are no separate chunks to wait for. IOW, we _do_ need to `<Loading>` instead of `<LoadingView>` for the nested-paneset reason you identify, but it doesn't matter now, right? I guess I'm just making sure your stripes-webpack is up to date :)

Yes, since lazy loading is turn off for now we don't need it. Just wanted to comment this so we won't miss this when we decide to turn lazy loading back on and merge this PR

Also, I spoke with @JohnC-80 about nested panesets, and he said that they should work, and do work for plugins and settings, so I was wrong about nested panesets causing problems
But still it would make more sense IMO to use `<Loading>` here

PR #1220, opened 2022-06-22
STCOR-638 record detail panes are empty when printed

usavkov-epam:
## Purpose
https://issues.folio.org/browse/STCOR-638

This PR relates to [#1835](https://github.com/folio-org/stripes-components/pull/1835) and makes it possible to support pane content printing. (See PR [#1835](https://github.com/folio-org/stripes-components/pull/1835) to more details)


usavkov-epam:
Duplicate

PR #1224, opened 2022-07-05
STCOR-640: Add `button.duplicate` translation key

ncovercash:
# [Jira](https://issues.folio.org/projects/STCOR/issues/STCOR-640)

This is consistent with the UX pattern, as described [here](https://ux.folio.org/docs/guidelines/ux-patterns/duplicate-a-record/).  Adds a centralized duplicate button key to prevent modules having to provide separate translations for this common use. 

_Je parle un peu français_, so I added translations for fr and fr_FR.

### DIFF
--- a/translations/stripes-core/en.json
+++ b/translations/stripes-core/en.json
@@ -22,6 +22,7 @@
   "button.new": "+ New",
   "button.new_tooltip": "Add {entry}",
   "button.edit": "Edit",
+  "button.duplicate": "Duplicate",
   "button.delete": "Delete",
   "button.deleteEntry": "Delete {entry}",
   "button.saveAndClose": "Save and close",
--- a/translations/stripes-core/fr.json
+++ b/translations/stripes-core/fr.json
@@ -5,6 +5,7 @@
     "button.new": "+ New",
     "button.new_tooltip": "Add {entry}",
     "button.edit": "Edit",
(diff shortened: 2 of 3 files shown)


zburke:
@ncovercash, what's the goal here, just to get "Duplicate" into one central place so that ui-whatever can (should) use `<FormattedMessage id="stripes-core.button.duplicate" />` instead of having `id="ui-whatever.duplicate"` keys spread across multiple repos? That's fine if so. I just want to make sure I understand :) Even when there is a verbose Jira description, it's still helpful to include a one-liner in the GitHub PR to briefly explain the "why" in addition to the subject-line's "what". 

NB: values for anything outside `en.js` are handled through a separate process so any changes committed here will be overwritten. There is nothing wrong with adding them, but the data flow is one-way from `en.json` out to Lokalise.co, and then one-way back in from Lokalise to the various locale.json files so it's wasted effort.

zburke (approved):

ncovercash:
@zburke that's the goal, sorry for the gap in communication.  Didn't know about that for lokalise (that process is a bit of a mystery to me, admittedly) - will keep that in mind for the future

zburke:
@ncovercash If you can update this branch I'm happy to merge the PR.

ncovercash:
@zburke done

PR #1227, opened 2022-07-08
STCOR-640: Add `button.duplicate` translation key

zburke:
#1224 doppelgänger


zburke:
merged the original, #1224, instead.

PR #1233, opened 2022-08-01
STCOR-603 show supplemental data on the /settings/about page

zburke:
Optionally decorate the `/settings/about` (Settings: Software versions)
page with information from Okapi Environment variables,
mod-configuration values, or stripes.config, in descending order. Three
values may be displayed: a version, an install date, and a message.

The Okapi Environment keys are `ABOUT_INSTALL_DATE`,
`ABOUT_INSTALL_VERSION`, and `ABOUT_INSTALL_MESSAGE`. The mod-configuration
values all correspond to `module`: `@folio/stripes-core`, `configName`:
`aboutInstall` and the `code`s `date`, `version`, and `message`. The
`stripes.config.js` values are from the `stripes.config` object with the
keys `aboutInstallDate`, `aboutInstallVersion`, and
`aboutInstallMessage`.

Refs [STCOR-603](https://issues.folio.org/browse/STCOR-603), NH-603 (No taxes, no services! Live Free or Die!)

### DIFF
--- a/src/queries/useConfigurations.js
+++ b/src/queries/useConfigurations.js
@@ -29,7 +29,7 @@ const useConfigurations = ({ module, configName, code }) => {
 
   const { isFetching, data, error } = useQuery(
     [namespace, configName, code],
-    async () => {
+    () => {
       if (stripes.hasPerm('configuration.entries.collection.get')) {
         return ky.get(configurationsApi(module, configName, code))
           .json();
--- a/src/queries/useConfigurations.js
+++ b/src/queries/useConfigurations.js
@@ -1,0 +1,48 @@
+import { useQuery } from 'react-query';
+
+import queryLimit from '../queryLimit';
+import useOkapiKy from '../useOkapiKy';
+import { useStripes } from '../StripesContext';
+
+
+export const configurationsApi = (module, configName, code) => {
+  const params = [];
+  if (module) {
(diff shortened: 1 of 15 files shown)


src/queries/useConfigurations.js
```diff
@@ -0,0 +1,48 @@
...
+};
+
+const useConfigurations = ({ module, configName, code }) => {
+  const stripes = useStripes();
+  const ky = useOkapiKy();
+  const namespace = '@folio/stripes-core:QUERY_KEY_CONFIGURATIONS';
+
+  const { isFetching, data, error } = useQuery(
+    [namespace, configName, code],
+    async () => {
```
mkuklis, line 32: @zburke I don't think we need `async` here since there is no `await` right?

mkuklis: Or perhaps the `await` is missing before `ky.get`?

mkuklis (approved):
A clean, well tested and documented code ❤️ .

I left just one little comment.

src/queries/useConfigurations.js
zburke: I think you're correct that `async` here is superfluous. It's `useQuery` that would be calling `await` here, but given `ky.get()` is already async there's no value in restating it. I'll remove it. 

Thanks for mentioning this and helping to keep the code as clean as possible!

PR #1234, opened 2022-08-02
STCOR-645 add cs_CZ to supported locales

zburke:
Add `cs-CZ` (Czech, Czechia) to the list of supported locales.

Refs [STCOR-645](https://issues.folio.org/browse/STCOR-645)

### DIFF
--- a/src/loginServices.js
+++ b/src/loginServices.js
@@ -31,7 +31,7 @@ export const supportedLocales = [
   'ar',     // arabic
   'zh-CN',  // chinese, china
   'zh-TW',  // chinese, taiwan
-  'cs_CZ',  // czech, czechia
+  'cs-CZ',  // czech, czechia
   'da-DK',  // danish, denmark
   'en-GB',  // british english
   'en-SE',  // english, sweden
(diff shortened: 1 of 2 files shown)


src/loginServices.js
```diff
@@ -31,6 +31,7 @@ export const supportedLocales = [
   'ar',     // arabic
   'zh-CN',  // chinese, china
   'zh-TW',  // chinese, taiwan
+  'cs_CZ',  // czech, czechia
```
dltj, line 34: @zburke - Is it okay that Czech uses an underscore as a separator (as it should be) or at this point should it be a dash (as it has historically been)?

zburke: Whoops, that's a bug. Sharp eyes, thank you! [The dash is correct (RFC-5646)](https://datatracker.ietf.org/doc/html/rfc5646#section-2.1). I don't know the our underscore convention came from, if it's a peculiarity of Lokalise or something we did by accident.

dltj: Yes, unfortunately that decision is lost to the mists of time.  I have a vague memory that there was a reason why we did it this way, but I've lost all hope of trying to reconstruct that reasoning.  

Thanks, Zak.

PR #1235, opened 2022-08-03
STCOR-646: Settings > About> stripes-final-form version is not displayed

BogdanDenis:
## Description
Display `stripes-final-form` version in Settings > About page

## Screenshots
image

## Issues
[STCOR-646](https://issues.folio.org/browse/STCOR-646)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 * Use documenation's root URL in NavBar `?` link. Refs STCOR-621.
 * Allow customization of login page's CSS. Refs STCOR-643.
 * Allow customization of navbar CSS. Refs STCOR-644.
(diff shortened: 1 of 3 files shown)


zburke:
This would create a circular dependency between stripes-core and stripes-final-form. [stripes-core is a peer-dep of stripes-final-form](https://github.com/folio-org/stripes-final-form/blob/3f92aaeb798458f1993ba5ac98cf0d022135ff5c/package.json#L44) and is leveraged in [StripesFinalFormWrapper](https://github.com/folio-org/stripes-final-form/blob/3f92aaeb798458f1993ba5ac98cf0d022135ff5c/lib/StripesFinalFormWrapper.js#L8).

PR #1243, opened 2022-09-11
STCOR-649 persist login response to storage

zburke:
Persist the response object from the login request to storage. This allows parsing of that object to be handled by external modules that are legit allowed to have knowledge of the optional attributes that may be present as result of optional modules being included by `users-bl`. Technically, `stripes-core` shouldn't contain logic for parsing response data from interfaces it doesn't have knowledge of, thus the removal here of parsing related to Service Points.

Eventually, all code that deals with Service Points should be removed, but as some of those functions are exported, that is technically a breaking change. Additional code in `<ProfileDropdown>` that deals with labeling the user-profile dropdown menu will also need to be refactored.

Things of note here: 
* `updateUser` exported from `./src/loginServices.js`, and a corresponding method on the `stripes` object on in `./src/okapiReducer.js`, spreads the arguments it receives onto the user object, allowing the user to be modified outside of stripes-core. That sounds sketchy, but it's essentially a generic version of the `setServicePoints` and `setCurServicePoint` functions that allow the user's service-point attributes to be modified outside of stripes-core. Without a method like this, other modules are restricted to either (a) re-parsing the login-response object themselves or (b) only having access to the `stripes.user` object as initialized in `./src/components/Root/`. 
* All the changes in `./src/loginServices.js` that aren't directly related to service points aren't real changes. New `return` statements etc just provide consistent return types, or correctly return promises rather than executing synchronously.

Refs [STCOR-649](https://issues.folio.org/browse/STCOR-649)


src/loginServices.js
```diff
@@ -360,38 +363,41 @@ function createOkapiSession(okapiUrl, store, tenant, token, data) {
...
  */
 function validateUser(okapiUrl, store, tenant, session) {
   return fetch(`${okapiUrl}/bl-users/_self`, { headers: getHeaders(tenant, session.token) }).then((resp) => {
-    if (resp.status >= 400) {
+    if (resp.ok) {
+      const { token, user, perms } = session;
+      return resp.json().then((data) => {
+        store.dispatch(setLoginData(data));
+        store.dispatch(setSessionData({ token, user, perms }));
+        return loadResources(okapiUrl, store, tenant, user.id);
```
mkuklis, line 380: @zburke not sure why but it feels wrong to me to handle loading resources inside `createOkapiSession`. I would expect `createOkapiSession` to just create the session. I wonder if we could instead load them inside `processOkapiSession`.

src/loginServices.js
```diff
@@ -446,36 +452,55 @@ function processSSOLoginResponse(resp) {
...
-    handleLoginError(dispatch, resp);
-  } else {
-    resp.json()
-      .then(json => createOkapiSession(okapiUrl, store, tenant, token, json))
-      .then(() => {
+  if (resp.ok) {
+    return resp.json()
+      .then(json => {
+        return createOkapiSession(okapiUrl, store, tenant, token, json)
+          .then(() => json);
```
mkuklis, line 495: Nested `then` is hard to read but I get why you had to handle it this way. It looks like you want to get the outer `json` from the original `resp`.

I wonder if the processing of that original response could be handled by the function which calls the `processOkapiSession` so the `processOkapiSession` doesn't really have to worry about processing it (handling errors, etc). The `processOkapiSession` could just get a json as an argument instead of `resp`. With that approach this could be simplified with:

````js
return createOkapiSession(okapiUrl, store, tenant, token, json)
  .then(store.dispatch(setOkapiReady())
  .then(() => json)
````

and actually, if we move the `loadResources` here (from `createOkapiSession`) this could become:

````js
return createOkapiSession(okapiUrl, store, tenant, token, json)
  .then(okapiSess => loadResources(okapiUrl, store, tenant, okapiSess.user.id))
  .then(store.dispatch(setOkapiReady())
  .then(() => json)
````

mkuklis (approved):
@zburke beautiful work 🎉 . Thank you for cleaning up all these promise responses and for adding the comments. 

I left just couple small comments after reviewing it a bit closer. Please feel free to ignore them if you feel they are too nit-picky.

src/loginServices.js
zburke: This isn't nitpicky at all; it's a great idea! I'd lean toward a separate PR for it just to avoid the "too many things at once" syndrome, though I'm guilty of that here already. A lot of the work that looks like new work, introducing promise chains and such, isn't really new work though -- at least, it doesn't move anything; it just formalizes/awaits the operations as they already were.

PR #1244, opened 2022-09-12
STCOR-652: Extend Prop-Types of Route

EthanFreestone:
We are in a situation where passing information down from the top level Route is necessary, and can be accomplised y passing props to a child function, however this trips PropTypes.

Currently working around by using Route from react-router-dom, but would be nice to expand PropTypes here so we don't need a separate import for this use case

### DIFF
--- a/src/components/NestedRouter.js
+++ b/src/components/NestedRouter.js
@@ -39,7 +39,10 @@ export function Route({ component: Component, children, ...props }) {
 }
 
 Route.propTypes = {
-  children: PropTypes.node,
+  children: PropTypes.oneOfType([
+    PropTypes.func,
+    PropTypes.object,
+  ]),
   component: PropTypes.oneOfType([
     PropTypes.func,
     PropTypes.object,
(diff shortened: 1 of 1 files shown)


zburke:
> Currently working around by using Route from react-router-dom, but would be nice to expand PropTypes here so we don't need a separate import for this use case

I don't quite understand how the "separate import" work-around works; can you explain in a little more detail? Assuming this lack of understanding is just my problem, can you write up a quick Jira and link it in the PR title? That'll help us (and POs) connect this change to a particular flower version.

src/components/NestedRouter.js
```diff
@@ -39,7 +39,10 @@ export function Route({ component: Component, children, ...props }) {
 }
 
 Route.propTypes = {
-  children: PropTypes.node,
+  children: PropTypes.oneOfType([
+    PropTypes.func,
+    PropTypes.object,
```
mkuklis, line 44: @EthanFreestone is there any reason why you used `object` here instead of `node`?

I wonder if this can be replaced with something like:

 ````js

children: PropTypes.oneOfType([
    PropTypes.arrayOf(PropTypes.node),
    PropTypes.node,
    PropTypes.func
  ])
````

EthanFreestone: Ah, I actually thought I had used node, will change immediately :)

zburke (approved):

PR #1246, opened 2022-09-23
re-export history dependency

JohnC-80:
Module runtime code will typically receive this handed down from our `react-router` context, but modules can also import it as a `devDependency` for their local test suites, which means that there can be a difference in major versions/functionality between what runs in the test suite vs what runs in the platform runtime. Re-exporting this will give ui modules something that they can use in their test suites to stay in sync and remove their local `devDependency`.


JohnC-80:
Result of further conversation in stripes-architecture :

Since jest tests do a great deal of mocking to different degrees, potentially mocking all of stripes core, why re-export this and make  stripes-core they only external dependency for your test suite to get that re-export? better to do the same peerDep/devDep workflow we've had with other dependencies.

PR #1251, opened 2022-10-20
STCOR-395 v1 don't depend on 3rd party for login form.

JohnC-80:
This PR strips React Final Form out of the login form. This is one approach on the path of better login security...

Approach: Use only the state of the DOM to store the login info. No form values from react state. No Final Form... just the steps needed to provide credentials to log-in services...


skomorokh:
Woo! Solid start!

PR #1253, opened 2022-10-26
STCOR-651  - (STCOR-395v2) - Render login form within iframe

JohnC-80:
associated JIRA's : https://issues.folio.org/browse/STCOR-651 

This PR depends on the `STCOR-395-login-iframe` branch of `stripes-webpack` to work, as it imports/loads the iframe content...

The iframe contents use vanilla js/HTML.
The iframe sends the okapi token to local storage directly... This is reverse order of the logic from before (store, then local storage).  After this, it signals the application, sending the login response data that couldn't be stored in localStorage, and the application updates the store accordingly.
The iframe uses the JS `postMessage` api for any communication with the main app. This handles various functionality including translations and letting the app know that the login was successful.

Still left:
- [ ] Styling
- [ ] Rendering the forgot password links in the iframe
- [ ] Localization step of error message presentation
- [ ] Tests
- [ ] Security sanity review


JohnC-80:
Closing this for sake of better approaches to CSP.

PR #1257, opened 2022-11-14
STCOR-659 suppress react-intl warnings

zburke:
Leverage `stripes.config.suppressIntlErrors` to suppress warnings in addition to errors, now that there is FINALLY support for the [`onWarn` prop](https://github.com/formatjs/formatjs/issues/3454) as of `v5.25.1`.

See also PR #912 where we added `onError` handling.

Refs [STCOR-659](https://issues.folio.org/browse/STCOR-659)

### DIFF
--- a/src/components/Root/Root.js
+++ b/src/components/Root/Root.js
@@ -165,7 +165,7 @@ class Root extends Component {
                   messages={translations}
                   textComponent={Fragment}
                   onError={config?.suppressIntlErrors ? () => {} : undefined}
-                  onWarn={config?.suppressIntlErrors ? () => {} : undefined}
+                  onWarn={config?.suppressIntlWarnings ? () => {} : undefined}
                   defaultRichTextElements={this.defaultRichTextElements}
                 >
                   <RootWithIntl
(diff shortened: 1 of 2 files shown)


MikeTaylor (approved):
Good to see!

src/components/Root/Root.js
```diff
@@ -165,6 +165,7 @@ class Root extends Component {
                   messages={translations}
                   textComponent={Fragment}
                   onError={config?.suppressIntlErrors ? () => {} : undefined}
+                  onWarn={config?.suppressIntlErrors ? () => {} : undefined}
```
skomorokh, line 168: Hm, if you made it `suppressIntlWarnings` here wouldn't that suffice to give us  the option of suppressing errors, warnings, or both?

skomorokh (approved):
Yay, I'm glad to see they've deigned to let us filter out this noise in our consoles. Thanks for surfacing that in the stripes config. I'm happy with it as-is but do think you wouldn't have to go far out of your way to give people the option to choose whether to suppress errors, warnings or both (see other comment)

PR #1260, opened 2022-11-18
STCOR-661 Fix redirect in Production error page

BogdanDenis:
## Description
Fix Production error redirect to app's main page
Issue was that `module.home` isn't a real property - need to use `module.route`

## Screenshots

https://user-images.githubusercontent.com/19309423/202692134-be3c87d2-df35-4726-b104-30a9a0092977.mp4

## Issues
[STCOR-661](https://issues.folio.org/browse/STCOR-661)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 # 8.4.0 (IN PROGRESS)
 
 * Allow suppression of `react-intl` warnings, in addition to errors. Refs STCOR-659.
+* Catastrophic Messaging | Return to MARC authority. Fixes STCOR-661.
 
 ## [8.3.0](https://github.com/folio-org/stripes-core/tree/v8.2.0) (2022-06-14)
 [Full Changelog](https://github.com/folio-org/stripes-core/compare/v8.2.0...v8.3.0)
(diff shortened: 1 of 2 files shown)


zburke (changes requested):
I'm not convinced this is a stripes-core bug where we use the wrong property. Rather, it feels like a ui-marc-authorities bug where ui-marc-authorities is missing a required property. Additionally, this will cause incorrect behavior (though not truly "broken" behavior) for apps like [ui-users](https://github.com/folio-org/ui-users/blob/6b71b551f4016df8888dc9f5c2e43cea583c4b1f/package.json#L20-L21) and [ui-inventory](https://github.com/folio-org/ui-inventory/blob/99f61d75c25bd650ff2f0263f1c5a5dc65e931e6/package.json#L17-L18) that use different `home` and `route` values. 

Alternatively, we could do sth like `escapeRoute={stripes.home ?? stripes.route}` in order to be more tolerant of apps with different `home`/`route` values, or that omit a `home` value.

BogdanDenis:
Thanks @zburke , I missed that `module.home` was actually used in some apps. I've updated the PR with `module.home ?? module.route`

Dmytro-Melnyshyn (approved):

zburke (approved):
LGTM

PR #1267, opened 2022-12-20
STCOR-657 PasswordValidationField swallows error messages from API queries

uladislausamets:
## Purpose
This PR relates to [STCOR-657](https://issues.folio.org/browse/STCOR-657)
## Approach
The query was removed as it was invalid and not used by BE


UladzislauKutarkin (approved):

zburke (approved):
LGTM. It's worth noting that even though the query has been fixed here, the `getRule` method could still lacks a catch clause so if that query failed for some other reason, e.g. due to problems with permissions or if it returns a 500, there will still be an uncaught error.

NB: When fixing a bug, the PR title should state the solution, i.e. the thing that is changing, rather than restating the problem, i.e. the thing that's going away. A good rule of thumb for the first line of a commit is to have it complete the sentence, "After this commit, the code will ..." e.g. "include valid query for password validation rules".

PR #1268, opened 2022-12-29
STCOR-674 reset volatile stripes values on logout

zburke:
It is possible to modify certain stripes values via ui-developer, e.g. Settings > Developer > Configuration makes it possible to set the logging categories, show permissions in the user-menu, ignore all if-permission checks, etc.

These stripes values should be reset to their default values on logout.

Refs [STCOR-674](https://issues.folio.org/browse/STCOR-674)


src/Stripes.test.js
```diff
@@ -0,0 +1,123 @@
...
+      const logger = { log: jest.fn() };
+      const s = new Stripes({ logger, config: { funky: 'chicken' } });
+      s.reset();
+      expect(s.config).toMatchObject({});
+      expect(s.logger.categories).toEqual('core,action,xhr');
+      expect(s.logger.prefix).toEqual('stripes');
+      expect(s.logger.timestamp).toBe(false);
+    });
+  });
+});
```
MikeTaylor, line 123: These seem like very pedantic tests for such small change, but ¯\_(ツ)_/¯

src/Stripes.test.js
```diff
@@ -0,0 +1,123 @@
...
+    });
+  });
+
+  describe('hasInterface', () => {
+    describe('returns undefined', () => {
+      it('before discovery', () => {
+        const logger = { log: jest.fn() };
+        const s = new Stripes({ logger });
+        expect(s.hasInterface('monkey')).toBeUndefined()
+        expect(logger.log).toHaveBeenCalled();
```
MikeTaylor, line 26: Maybe add the expected logging string here and elsewhere?

src/Stripes.test.js
```diff
@@ -0,0 +1,123 @@
...
+      it('if the interface is absent', () => {
+        const logger = { log: jest.fn() };
+        const s = new Stripes({ logger, discovery: { interfaces: { 'funky': '3.0' }} });
+        expect(s.hasInterface('monkey')).toBeUndefined();
+        expect(logger.log).toHaveBeenCalledWith('interface', 'interface \'monkey\' is missing');
+      });
+    });
+
+    describe('returns true', () => {
+      it('when the interface is present and no version is provided', () => {
```
MikeTaylor, line 45: The interface-version in this test is _provided_ (in the config) but not _required_ (in the `hasInterface` call), so maybe rename the test?

src/Stripes.test.js
```diff
@@ -0,0 +1,123 @@
...
+        const s = new Stripes({ logger, discovery: { interfaces: { 'monkey': '3.0' } } });
+        expect(s.hasInterface('monkey', '3.0')).toBe('3.0');
+        expect(logger.log).toHaveBeenCalled();
+      });
+    });
+
+    it('returns false when the requested version is absent', () => {
+      const logger = { log: jest.fn() };
+      const s = new Stripes({ logger, discovery: { interfaces: { 'monkey': '3.0' }} });
+      expect(s.hasInterface('monkey', '4.0')).toBe(0);
```
MikeTaylor, line 63: Not really an issue for the present PR, but ...

It's weird that `hasInterface` returns 0 here instead of false.

src/Stripes.test.js
```diff
@@ -0,0 +1,123 @@
...
+      expect(s.hasPerm('monkey')).toBeUndefined();
+      expect(logger.log).toHaveBeenCalledWith('perm', 'not checking perm \'monkey\': no user permissions yet');
+    });
+  });
+
+  describe('reset', () => {
+    it('restores config and logger to default values', () => {
+      const logger = { log: jest.fn() };
+      const s = new Stripes({ logger, config: { funky: 'chicken' } });
+      s.reset();
```
MikeTaylor, line 116: I'm not sure I understand what the config is being reset _to_ in this case. Please advise.

MikeTaylor (changes requested):
Most of this looks good to me, but please see STCOR-674 for my thoughts on why we maybe shouldn't do this.

Even if we discard this "bugfix", it looks like there are some useful non-related tests in this PR, which we could preserve.

src/Stripes.test.js
zburke: My intent was that `reset()` would restore default values, i.e. those pulled from `stripes-config` (for `config`) and from `./src/configureLogger.js` (for `logger`). 

As you pointed out above, however, my impl doesn't actually do that correctly. Imagining the impl were correct, however, how would you change line 113 to be more descriptive?

src/Stripes.test.js
zburke: Existing code was untested, and they were easy tests to write, so 🤷

src/Stripes.test.js
MikeTaylor: I think we should just document in the Stripes Developer Guide what the API actually is, and not start a fight with unknown existing code. (I'm 98% certain its return value is only ever checked for truthiness, but ... who knows? And in the Firefly world of course it's literally impossible to know.)

src/Stripes.test.js
MikeTaylor: I would be more inclined to set a value into the config, check that it's set, call `reset` and then check that the value is not set any longer.

MikeTaylor (approved):
Looks great now!

src/Stripes.js
```diff
@@ -128,6 +131,15 @@ class Stripes {
   clone(extraProps) {
     return new Stripes(Object.assign({}, this, extraProps));
   }
+
+  /**
+   * reset
+   * reset config and logger properties to their default values
+   */
+  reset() {
+    this.config = config;
```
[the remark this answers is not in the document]
mkuklis: @zburke I think @MikeTaylor is right here. I was not aware that the original `config` can be modified. If that's the case do you think we should clone it when the original stripes object is created here?

https://github.com/folio-org/stripes-core/blob/a9ca1c2ac8a7df97be4ad436d460709b100605a1/src/components/Root/Root.js#L128

Otherwise, we may still have a problem similar to:

https://jsbin.com/zoxiperuva/edit?js,console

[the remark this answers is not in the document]
zburke: Unfortunately, @MikeTaylor is absolutely correct. My first crack at doing a deep clone when initializing stripes didn't work however. I'll spend a few more minutes on this, but the likely short-term outcome is that I'll unassign the ticket and leave it open.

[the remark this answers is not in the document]
MikeTaylor: What, after all the work we've put into it already? :-)

I urge you to consider the point made on the Jira. If after thinking that through you remain convinced that this is the right thing to do, then I think we should put in the extra effort to get this over the line.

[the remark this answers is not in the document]
mkuklis: I agree with @MikeTaylor! @zburke this all looks great. I think we just have to make sure we don't mutate the original config and it will be ready to go.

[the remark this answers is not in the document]
MikeTaylor: … assuming we want to do it at all, of course! See https://issues.folio.org/browse/STCOR-674#:~:text=added%20a%20comment%20%2D-,Yesterday,-I%27m%20not%20sure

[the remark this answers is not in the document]
mkuklis: @MikeTaylor I missed your comment and also didn't realize that the mutation only happens via ui-developer. Hmm... I will leave it to you two to decide what to do here. I kind if like what Zak did here and the additional tests too.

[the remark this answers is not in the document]
MikeTaylor: Even if we don't keep the change in functionality, we should definitely rescue the new tests.

zburke:
Calling `reset()` would only work if the values from `stripes-config` were properly frozen and cloned (yay, immutability!) in the first place instead of edited in-place. The idea behind this work still has merit, but this PR is junk.

PR #1271, opened 2023-01-18
STCOR-675 use stricter tenant-locale query

zburke:
The additional clause
```
(cql.allRecords=1 NOT userId="" NOT code="")
```
guarantees the response will contain only a single row. Without that clause, the clause was limited to
```
module==ORG and configName==localeSettings
```
which allowed it to be potentially be polluted by other rows that included `userId` or `code` values, meaning the result was non-deterministic.

Refs [STCOR-675](https://issues.folio.org/browse/STCOR-675)

### DIFF
--- a/src/loginServices.js
+++ b/src/loginServices.js
@@ -182,8 +182,14 @@ function dispatchLocale(url, store, tenant) {
  * @returns {Promise}
  */
 export function getLocale(okapiUrl, store, tenant) {
+  const query = [
+    'module==ORG',
+    'configName == localeSettings',
+    '(cql.allRecords=1 NOT userId="" NOT code="")'
+  ].join(' AND ');
+
   return dispatchLocale(
-    `${okapiUrl}/configurations/entries?query=(module==ORG and configName==localeSettings)`,
+    `${okapiUrl}/configurations/entries?query=(${query})`,
     store,
     tenant
   );
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 * Catastrophic Messaging | Return to MARC authority. Fixes STCOR-661.
 * Reset App Context Dropdown state when switching apps/unmounting. Fixes STCOR-664.
 * PasswordValidationField swallows error messages from API queries. Fixes STCOR-657.
+* Guarantee a single-row response to the tenant's locale-config query. Fixes STCOR-675.
 
 ## [8.3.0](https://github.com/folio-org/stripes-core/tree/v8.2.0) (2022-06-14)
 [Full Changelog](https://github.com/folio-org/stripes-core/compare/v8.2.0...v8.3.0)


src/loginServices.js
```diff
@@ -182,8 +182,14 @@ function dispatchLocale(url, store, tenant) {
  * @returns {Promise}
  */
 export function getLocale(okapiUrl, store, tenant) {
+  const query = [
+    'module==ORG',
+    'configName == localeSettings',
+    '(cql.allRecords=1 NOT userId="" NOT code="")'
```
MikeTaylor, line 188: Please tell me `userId=""` is not how RMB does a "this field is not present" seach?

MikeTaylor, line 188: Also, why does `code` come into this?

MikeTaylor (commented):
Holding fire till I understand what's happening here.

src/loginServices.js
zburke: Alas, `cql.allRecords=1 NOT userId=""` is _exactly_ how CQL specifies [all records where `userId` is not defined](https://github.com/folio-org/raml-module-builder/tree/57ad315376b1bb9e9b6f7d25e89c31ebcfdd4635#cql-matching-undefined-or-empty-values). Given that ui-tenant-settings [only provides `module` and `configName` when storing settings](https://github.com/folio-org/ui-tenant-settings/blob/c1c5514f444092c82867db6f4392069444746219/src/settings/Locale.js#L40-L41), asking for `userId` and `code` to be absent feels like the correct way to retrieve a unique record.

zburke: A config entry has up to four qualifiers (`module`, `configName`, `code`, `userId`) and a `value`. ui-tenant-settings stores its value by specifying only the `module` and `configName` qualifiers. Thus, this query seeks to retrieve the record where those qualifiers are present and the others are absent. If we don't ensure that `code` is absent, the query is open to pollution from records like 
```
{
  module: "ORG",
  configName: "localSettings",
  code: "hahahah yer in trouble now",
  value: "{locale: \"Klingon\"}"
}
```

MikeTaylor: You are right about code-pollution, of course. I feel like an idiot for not having seen that immediately.

zburke: Pffffft. If you're an idiot than we all are; this insufficiently-strict-query bug has been lurking since [the initial implementation in 2017](https://github.com/folio-org/stripes-core/pull/17).

MikeTaylor: Well, that (`NOT userId=""`) is horrifying.

BTW., in CQL, `foo and (cql.allRecords=1 not bar)` is exactly equivalent to (foo not bar)`. I _would_ suggest using the more concise version of the syntax here, but since this area already reeks of special-cases, I wouldn't be totally confident that the involvement of `cql.allRecords=1` isn't also special here. So if you've tested the present version and it does what you want, I say leave it.

zburke: I got my sample from the RMB docs so I'm gonna stick with it, since I think most folks on FOLIO will use that as a reference. But I appreciate you teaching me about the other options!

PR #1274, opened 2023-01-27
STCOR-650 remove setServicePoints, setCurServicePoint

zburke:
*BREAKING*:
Remove functions related to service points. Given this repo does not have a direct dependency on the `inventory` interface, it must not contain code that relies on values provided by that interface. Handling of service point data on login moved to ui-service-points in [UISP-32](https://issues.folio.org/browse/UISP-32), hence the corresponding functions that did that work here can now be removed.

Since these functions were publicly exported, this is a breaking change.

Refs [STCOR-650](https://issues.folio.org/browse/STCOR-650)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,8 @@
 * PasswordValidationField swallows error messages from API queries. Fixes STCOR-657.
 * Guarantee a single-row response to the tenant's locale-config query. Fixes STCOR-675.
 * *BREAKING*: Remove SWR, a POC that never took off. Refs STCOR-516, STCOR-611.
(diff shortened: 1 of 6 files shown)


zburke:
@MikeTaylor: FYI, in conjunction with https://github.com/folio-org/ui-servicepoints/pull/128, this removes the last vestiges of service-points in stripes-core, once again restoring it to a totally library-agnostic system.

MikeTaylor:
That is beautiful to see!

(And a rare example of when a breaking change really _is_ necessary! 👍)

JohnC-80 (approved):

PR #1277, opened 2023-02-02
STCOR-686 - Switching service points not being captured in circulation log record

Terala-Priyanka:
## Purpose
Switching service points not being captured in circulation log record

## Refs
https://issues.folio.org/browse/STCOR-686

#### Current Issue:
Performing a checkout, after switching current service point of the logged in user, is not picking up the new/switched current service point.
#### Root Cause:
Although ui-servicepoints module is propagating the new current service point to the store, it is not being picked up by checkout app, as the apps/modules are not being updated with updated stripes object.
#### Fix:
This is fixed by adding stripes object into the dependency array of useMemo react hook, used for preparing the cacheModules.


zburke (approved):

mkuklis:
@Terala-Priyanka I think we have to be careful here. stripes object can be modified not just by service points and potentially causing multiple component reconnections. Perhaps instead of listing entire `stripes` object in the list of the `useMemo` dependencies we could narrow it down to service points?

Terala-Priyanka:
> @Terala-Priyanka I think we have to be careful here. stripes object can be modified not just by service points and potentially causing multiple component reconnections. Perhaps instead of listing entire `stripes` object in the list of the `useMemo` dependencies we could narrow it down to service points?

@mkuklis 
Yes, as I see from https://github.com/folio-org/stripes-core/blob/25bbf3416a7394de2ee9dc927d465f4f3347a37d/src/components/Root/Root.js#L121 , there is much more data that stripes object houses.

I think stripes should be in the list of useMemo dependencies here at Stripes-core level. 
It should be the responsibility of every module to limit the re-renders.
What do you think?

Terala-Priyanka:
@zburke and @mkuklis 
Apparently, there is not much impact of page load, although the page load can be affected by several other factors.

https://user-images.githubusercontent.com/104053200/217276417-0e734796-7add-4fb7-ad7b-35654ebd1985.mp4

mkuklis (approved):

zburke:
@Terala-Priyanka, thanks for taking the time to do some basic evaluation of the effect this could have. I'm content to merge this now.

Terala-Priyanka:
Thank you for your review @zburke and @mkuklis

PR #1279, opened 2023-02-07
STCOR-687: Configure global onError handler for React Query mutations

mkuklis:
When the mutation executed via `useMutation` from React Query fails and it doesn't provide a local `onError` handler
the errors are not handled at all leaving the end user without any feedback.

This PR configures a global onError which will be triggered only when there is no `onError` provided directly to `useMutation`.

https://issues.folio.org/browse/STCOR-687


zburke (approved):
👏 👏 👏 🥳 😻 😍 
It's great that `onError` accepts the mutation itself as a param so you can inspect `mutation.options.onError`. Very clever! This is clear, simple, and powerful. Nice work!

Assuming we merge this (I see no reason not to) we should announce it on #stripes-update. It's a good (but loud!) change in behavior that everybody should learn about so they can have a chance to adjust their code. 

We could also comment this a bit so instead of having separate documentation to maintain we could tell people, "Just go read `createReactQueryClient`. It'll tell you about it's helpful defaults, and about it's designed-to-be-overridden defaults."

zburke:
@mkuklis, is there a generic way to capture the response body from the failed mutation? It would be nice to show that if possible, though I guess the goal is really just to get a developer's attention that they have an unhandled condition, and this certainly does that.

mkuklis:
@zburke good idea. I added a little error parser. "It ain't pretty but it's honest work"

honest_work

PR #1281, opened 2023-02-14
STCOR-668 include rimraf, miragejs as dev-deps

zburke:
Move non-production deps out of `dependencies` and into `devDependencies`. Remove `rimraf` altogether since it's no longer in use. 

Refs [STCOR-668](https://issues.folio.org/browse/STCOR-668), [STCOR-681](https://issues.folio.org/browse/STCOR-681)


zburke:
@skomorokh, I feel sorta nostalgically bad removing this line from the `initial commit` from October 2016, but I don't think it's serving a purpose any longer. Agreed?

skomorokh:
Agreed! More than enough cruft around to not get rid of as much of it as we can. Besides, git lets us travel back in time whenever we want some nostalgia.

JohnC-80 (approved):

PR #1284, opened 2023-02-21
STCOR-696: If you choose a numbering system and language in Settings > Tenant > Language and localization, saving the new values does not change the locale

UladzislauKutarkin:
Ticket: [STCOR-696](https://issues.folio.org/browse/STCOR-696)
Goal: Fix bug when we choose language with numeric system fail with error in console and language not changing
Result: 

https://user-images.githubusercontent.com/72550466/220290773-9d9fdf98-e0db-4380-9635-460f94955fc3.mov

### DIFF
--- a/src/loginServices.js
+++ b/src/loginServices.js
@@ -107,6 +107,7 @@ export function loadTranslations(store, locale, defaultTranslations = {}) {
   // Since moment.js don't support translations like it or it-IT-u-nu-latn
   // we need to build string like it_IT for fetch call
   const loadedLocale = locale.replace('-', '_').split('-')[0];
+  const momentLocale = locale.split('-', 2).join('-');
 
   // react-intl provides things like pt-BR.
   // lokalise provides things like pt_BR.
@@ -125,8 +126,8 @@ export function loadTranslations(store, locale, defaultTranslations = {}) {
     // For moment, we want to import and load the most-specific
     // locale possible without the numbering system suffix,
     // e.g. it-IT if available, falling back to it if that fails.
-    import(`moment/locale/${locale}`).then(() => {
-      moment.locale(locale);
+    import(`moment/locale/${momentLocale}`).then(() => {
+      moment.locale(momentLocale);
     }).catch(() => {
       import(`moment/locale/${parentLocale}`).then(() => {
         moment.locale(parentLocale);
--- a/src/loginServices.js
+++ b/src/loginServices.js
@@ -104,6 +104,9 @@ function canReadConfig(store) {
  */
 export function loadTranslations(store, locale, defaultTranslations = {}) {
   const parentLocale = locale.split('-')[0];
+  // Since moment.js don't support translations like it or it-IT-u-nu-latn
+  // we need to build string like it_IT for fetch call
+  const loadedLocale = locale.replace('-', '_').split('-')[0];
 
   // react-intl provides things like pt-BR.
   // lokalise provides things like pt_BR.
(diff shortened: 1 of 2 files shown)


src/loginServices.js
```diff
@@ -104,6 +104,9 @@ function canReadConfig(store) {
  */
 export function loadTranslations(store, locale, defaultTranslations = {}) {
   const parentLocale = locale.split('-')[0];
+  // Since moment.js don't support translations like it or it-IT-u-nu-latn
+  // we need to build string like it_IT for fetch call
+  const loadedLocale = locale.replace('-', '_').split('-')[0];
```
zburke, line 109: I don't think this line does what you want it to, because first it'll replace dashes with underscores, then split on dashes (there will now be none), then return the first element, which means it'll be identical to the `locale` argument. Thus function needs to care about four different values:
1. locale including region and number scheme if present, separated by dashes, e.g. `fr` or `fr-FR` or `fr-FR-u-nu-latn`, because we want to call `dispatch(setLocale(locale))` with a full string in order for react-intl to get the numbering system so it can handle number formatting correctly and for the document's `lang` attribute. 
2. locale including region if present, separated by dashes, e.g. `fr` or `fr-FR` because that's the kind of value `getLangDir` and `moment` expect. 
3. locale without the region, e.g. `fr`, because if `moment` doesn't recognize sth like `en-SE` we want it to fall back to loading `en`. 
4. locale including region if present, separated by underscore, e.g. `fr` or `fr_FR` because that's the kind of value our sad sad lokalise system uses as a key into the `translations` array imported from `stripes-config`.

UladzislauKutarkin: So @zburke what I need to change ? I don't have any ideas for now

uladislausamets (approved):

ryandberger (approved):

src/loginServices.js
```diff
@@ -113,21 +116,32 @@ export function loadTranslations(store, locale, defaultTranslations = {}) {
...
+    // locale possible without the numbering system suffix,
+    // e.g. it-IT if available, falling back to it if that fails.
+    import(`moment/locale/${locale}`).then(() => {
+      moment.locale(locale);
+    }).catch(() => {
+      import(`moment/locale/${parentLocale}`).then(() => {
+        moment.locale(parentLocale);
+      }).catch(e => {
+        // eslint-disable-next-line no-console
+        console.error(`Error loading locale ${parentLocale} for Moment.js`, e);
```
zburke, line 135: Given a locale like `en-SE-u-nu-latn`, this will run
```
import('en-SE-u-nu-latn')
.catch(() => import ('en'))
```
But what we want to be doing is
```
import('en-SE')
.catch(() => import ('en'))
```

zburke (changes requested):
Sorry; this isn't quite right for the moment paths we want to attempt to load.

src/loginServices.js
UladzislauKutarkin: @zburke good point. Done!

zburke (approved):
LGTM! Thanks for your perseverance in figuring out the problem, and you patience while we worked toward a complete solution. 🤝

zburke:
NB: please squash when you merge to keep the commit history nice and clean, and consider updating the commit subject/PR subject to something that succinctly describes the state of the code now (i.e. what the fix is), rather than what the problem was before this commit (i.e. what the bug was). e.g. "STCOR-696 correctly configure moment, even given an extended locale" or something like that. 

Rule of thumb for bugs: In Jira, the subject should describe the problem. In Git, the subject should describe the solution.

UladzislauKutarkin:
> LGTM! Thanks for your perseverance in figuring out the problem, and you patience while we worked toward a complete solution. 🤝

NP! We making FOLIO better!

PR #1303, opened 2023-04-25
STRIPES-861: Setup module federation

mkuklis:
https://issues.folio.org/browse/STRIPES-861


JohnC-80:
Closing this awesome piece of work - it played a great role in getting us to the point where we are!

PR #1307, opened 2023-05-03
STCOR-689 STCOR-690 Support switch consortium active affiliation

usavkov-epam:
## Purpose
https://issues.folio.org/browse/STCOR-689
https://issues.folio.org/browse/STCOR-690

As a central tenant of a consortium, it should be able to switch the current context to manage institutional tenants. Header `X-Okapi-Tenant` acts as a pointer to the current context when sending requests to the server. Requests in the FOLIO are sent with a predefined header `X-Okapi-Tenant`, which is stored in redux store and passed to the Stripes object. Therefore, context switching requires the creation of additional functionality that will process this action.

---

If a user login into the consortium central tenant, he expects to see its name in the top-right corner of the screen (in the profile dropdown trigger, above a current SP name).

## Approach
image

- Add consortia services, which are responsible for providing consortium data, switching active affiliation, and persisting it in the okapi session;
- Handle in the okapi reducer actions for set/clear consortium data and set current tenant ID;
- Export service, that will handle switching affiliation in the separate module (after click "Switch active affiliation" in profile dropdown);
- Request consortium data after user login and loading resources - we need a list of available interfaces to determine if the consortium data needs to be loaded;
- Render active tenant (affiliation) name in the profile dropdown trigger;
- Implement unit test;

<details open><summary>Preview</summary>
<p>

firefox_YiaXFrBDj8

</p>
</details>


src/components/MainNav/ProfileDropdown/ProfileDropdown.js
```diff
@@ -122,11 +129,20 @@ class ProfileDropdown extends Component {
   }
 
   toggleDropdown() {
+    // Get items after rechecking for item visibility
+    this.userLinks = this.getDropdownMenuLinks();
```
usavkov-epam, line 133: Previously, the function to check the visibility of dropdown items was called only once in the component constructor, however, the conditions for displaying could change, so it makes sense to do an additional check when opening the dropdown.

BogdanDenis (approved):

zburke (changes requested):
Sorry; consortium-related functionality, like service points, does not belong in stripes-core at all. Please move this functionality to its own repository.

NikitaSedyx:
@zburke but consortium-related functionality is part of header which is part of stripes-core, what is the right place to move it?
especially when this consortia related staff is kind of part of auth process and stripes-core handles this process as well

mkuklis (approved):
@usavkov-epam I left you one little comment about potentially refactoring one function to async/await.

src/consortiaService.js
```diff
@@ -0,0 +1,164 @@
...
+ * 1. Fetch data related to current consortium;
+ * 2. Update consortium in okapi session;
+ * 3. Update current tenant.
+ *
+ * @param {redux-store} store
+ * @param {object} data
+ *
+ * @returns {void}
+ */
+export const fetchCurrentConsortiumData = (store, data) => {
```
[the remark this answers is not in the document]
usavkov-epam: Yes, for sure, I've updated that function, thanks!

src/consortiaService.js
```diff
@@ -0,0 +1,153 @@
...
+
+    const [
+      { tenants },
+      { userTenants: userAffiliations },
+    ] = await Promise.all([
+      fetchConsortiumTenants(
+        url,
+        centralTenant,
+        { consortiumId },
+      ).then(resp => resp.json()),
```
mkuklis, line 127: @usavkov-epam it would be nice if `fetchConsortiumTenants` already did that for us so you can just get the results here without using `then`.

src/consortiaService.js
```diff
@@ -0,0 +1,153 @@
...
+      fetchConsortiumUserAffiliations(
+        url,
+        centralTenant,
+        token,
+        { consortiumId, userId: data.user.id },
+      ).then(resp => resp.json()),
```
mkuklis, line 133: @usavkov-epam same here.

src/consortiaService.js
```diff
@@ -0,0 +1,153 @@
...
+ * @param {redux-store} store
+ * @param {object} data
+ *
+ * @returns {void}
+ */
+export const fetchCurrentConsortiumData = async (store, data) => {
+  try {
+    const centralTenant = okapi.tenant;
+    const { token, url } = store.getState().okapi;
+    const { consortia, totalRecords } = await fetchConsortia(url, centralTenant).then(resp => resp.json());
```
mkuklis, line 113: @usavkov-epam `fetchConsortia` could take care of parsing the json so you wouldn't have to worry about it here.

src/consortiaService.js
usavkov-epam: Updated

src/consortiaService.js
usavkov-epam: Updated

mkuklis: @usavkov-epam Thank you! This is so much easier to read now.

NikitaSedyx:
closing this one as https://github.com/folio-org/stripes-core/pull/1308 opened instead

PR #1308, opened 2023-05-08
STCOR-689 STCOR-690 Support switch consortium active affiliation

NikitaSedyx:
## Purpose
https://issues.folio.org/browse/STCOR-689
https://issues.folio.org/browse/STCOR-690

As a central tenant of a consortium, it should be able to switch the current context to manage institutional tenants. Header `X-Okapi-Tenant` acts as a pointer to the current context when sending requests to the server.

---

If a user login into the consortium central tenant, he expects to see its name in the top-right corner of the screen (in the profile dropdown trigger, above a current SP name).

## Approach
- Validate user based on current tenant value stored in a session (if absent - use tenant from stripes configs);
- Expose `updateTenant` service, which will update the current tenant, fetch resources (perms, SPs and etc.) and update session;
- Allow `<HandlerManager>` to invoke handlers for other module types (`app`, `settings`);
- Render active tenant (affiliation) name in the profile dropdown trigger;
- Implement unit test;

<details open><summary>Preview</summary>
<p>

firefox_Wtyh0AQ9Zx

</p>
</details>

### DIFF
--- a/src/components/HandlerManager/HandlerManager.js
+++ b/src/components/HandlerManager/HandlerManager.js
@@ -20,15 +20,8 @@ class HandlerManager extends React.Component {
   constructor(props) {
     super(props);
     const { event, stripes, modules, data } = props;
-    const uniqueModules = Object.values(modules)
-      .flat()
-      .filter((module, index, self) => {
-        return index === self.findIndex((m) => (
-          m.module === module.module
-        ));
-      });
 
-    this.components = getEventHandlers(event, stripes, uniqueModules, data);
+    this.components = getEventHandlers(event, stripes, modules.handler, data);
   }
 
   render() {
--- a/src/components/HandlerManager/HandlerManager.js
+++ b/src/components/HandlerManager/HandlerManager.js
@@ -20,7 +20,7 @@ class HandlerManager extends React.Component {
   constructor(props) {
     super(props);
     const { event, stripes, modules, data } = props;
-    this.components = getEventHandlers(event, stripes, modules.handler, data);
+    this.components = getEventHandlers(event, stripes, Object.values(modules).flat(), data);
   }
 
   render() {
--- a/src/components/HandlerManager/HandlerManager.js
+++ b/src/components/HandlerManager/HandlerManager.js
@@ -20,7 +20,15 @@ class HandlerManager extends React.Component {
   constructor(props) {
     super(props);
     const { event, stripes, modules, data } = props;
-    this.components = getEventHandlers(event, stripes, Object.values(modules).flat(), data);
+    const uniqueModules = Object.values(modules)
+      .flat()
+      .filter((module, index, self) => {
+        return index === self.findIndex((m) => (
+          m.module === module.module
+        ));
+      });
+
+    this.components = getEventHandlers(event, stripes, uniqueModules, data);
   }
 
   render() {
--- a/src/components/MainNav/ProfileDropdown/ProfileDropdown.js
+++ b/src/components/MainNav/ProfileDropdown/ProfileDropdown.js
@@ -79,7 +79,13 @@ class ProfileDropdown extends Component {
 
   getModulesWithLinks = () => {
     const { modules } = this.props;
-    return ([].concat(...Object.values(modules)))
+    return Object.values(modules)
+      .flat()
+      .filter((module, index, self) => {
+        return index === self.findIndex((m) => (
+          m.module === module.module
+        ));
+      })
       .filter(({ links }) => links && Array.isArray(links.userDropdown));
   }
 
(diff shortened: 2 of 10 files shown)


usavkov-epam (approved):

NikitaSedyx:
hi @mkuklis  @zburke please review one more implementation for consortia stuff (follows recommendation to keep it agnostic to compare with https://github.com/folio-org/stripes-core/pull/1307)

src/components/HandlerManager/HandlerManager.js
```diff
@@ -20,7 +20,15 @@ class HandlerManager extends React.Component {
   constructor(props) {
     super(props);
     const { event, stripes, modules, data } = props;
-    this.components = getEventHandlers(event, stripes, modules.handler, data);
+    const uniqueModules = Object.values(modules)
```
mkuklis, line 23: @NikitaSedyx I think the original idea for this was to only support the `handler` modules. Could you explain why was this changed here?

NikitaSedyx: because non-handler modules can't handle events like `LOGIN`, according to https://github.com/folio-org/stripes/blob/master/doc/dev-guide.md#handlers-and-events 
```
Stripes modules of any type can provide a handler; the handler module type is provided for the case where the handler is the only thing the module provides, i.e. there is no main-screen app and no settings.
```
and it's strange that we provide a handler in app or settings module, but it can't handle events

NikitaSedyx: as an alternative, we can add `handler` type to our module, but it will contradict with description of handler module, additionally it will require to refactor ProfileDropdown logic to handle duplicates (actually the same logic that is applied here)

IMO following current documentation,  HandlerManager should support handler of all module types, not just handlers

zburke: I agree with @mkuklis here. I don't think this needs to change; you just need to add the `handler` to the `stripes.actsAs` array in the `package.json` of the ui-$whatever repo. Have you tried that?

NikitaSedyx: yes, in this case we have duplicates in ProfileDropdown actions because it lists action for each value in `actsAs` - as I mentioned above it's not a problem as we just need to add module uniqueness 

but here I have another point, for action menu we can add config to any module, define handler to return component and it will work without `stripes.actsAs: handler`, at the same time to handle events via `HandlerManager` we need to define module as a handler, it's usage inconsistency

NikitaSedyx: so we have to use cases to use handlers one doesn't require module as a handler, another one requires

NikitaSedyx: @zburke @mkuklis reverted changes for HandlerManager, probably handler docs section should be updated, at least 2 devs were confused by it :)

mkuklis: Thank you @NikitaSedyx. I agree with you that the docs should include the information about how to turn the existing ui-module into a `handler` and that the `HandlerManager` is intended to be used only by `hander` modules. Do you think that would be sufficient?

NikitaSedyx: I believe it should be enough, I was a bit confused by 2 points
1. any module can provide handler (but in fact ignored by `HandlerManager`)
2. handler type is used for case when module provides only handler

looks like what you mentioned covers them

mkuklis: Thank you @NikitaSedyx

mkuklis (approved):

src/components/MainNav/ProfileDropdown/ProfileDropdown.js
```diff
@@ -228,27 +242,40 @@ class ProfileDropdown extends Component {
...
     );
   }
 
+  renderProfileTriggerLabel = ({ open }) => {
+    const { okapi } = this.props.stripes;
+    const userData = this.getUserData();
+    const servicePointName = userData?.curServicePoint?.name;
+    const tenantName = userData?.tenants?.find(({ id }) => id === okapi.tenant)?.name;
+
+    const withLabel = Boolean(servicePointName || tenantName);
```
zburke, line 264: A better convention than `with...` is `is...` or `has...`. We use `withFoo` as a naming convention for HOCs that provide a `Foo` object to the component they wrap.

zburke (approved):

src/components/MainNav/ProfileDropdown/ProfileDropdown.js
NikitaSedyx: agree, missed this migrating from https://github.com/folio-org/stripes-core/pull/1307 PR

PR #1311, opened 2023-05-17
STCOR-721 Add auth errors handler on Login screen

NikitaSedyx:
purpose: we'd like to show more fields on login screen when auth is failed (https://issues.folio.org/browse/STCOR-721)

approach: add new event and HandlerManager on login screen so modules can extend Login form


NikitaSedyx:
not sure about tests, should we add jest or bigtest?

NikitaSedyx:
hi @JohnC-80 @zburke @mkuklis 
could you please take a look?

mkuklis (approved):

PR #1315, opened 2023-06-01
STCOR-723: Improve login error message when back-end not ready

julianladisch:
When using a Vagrantfile the nginx server immediately serves the frontend. Okapi starts deploying the back-end modules, however, depending on available resources (cpus etc.) it takes 5-30 minutes until all back-end modules have been deployed and a login can succeed.

During this deploy time Okapi returns a 404 error for any login attempt using POST /bl-users/login

stripes-core displays this error message: "Sorry, the information entered does not match our records".

Change to "System error. Please wait a few minutes and try again."

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 * Apps icons replaced with blue rectangle on apps menu. Refs STCOR-707.
 * Display consortium active affiliation in the profile dropdown trigger. Refs STCOR-689.
 * Support switch consortium active affiliation. Refs STCOR-690.
+* Use tenant query param for password reset. Refs STCOR-724.
 * Improve login error message when back-end not ready. Refs STCOR-723.
 
 ## [9.0.0](https://github.com/folio-org/stripes-core/tree/v9.0.0) (2023-01-30)
--- a/src/components/CreateResetPassword/CreateResetPasswordControl.js
+++ b/src/components/CreateResetPassword/CreateResetPasswordControl.js
@@ -7,15 +7,19 @@ import processBadResponse from '../../processBadResponse';
 import { stripesShape } from '../../Stripes';
 import { setAuthError } from '../../okapiActions';
 import { defaultErrors } from '../../constants';
+import OrganizationLogo from '../OrganizationLogo';
 
 import CreateResetPassword from './CreateResetPassword';
 import PasswordHasNotChanged from './components/PasswordHasNotChanged';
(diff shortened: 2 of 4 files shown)


MikeTaylor (approved):
Seems OK.

Should there be a CHANGELOG entry?

zburke (commented):
@julianladisch , @MikeTaylor , I'm not sure how I feel about this. This feels like we're assuming a pretty generic condition (we received a 404) corresponds with a very specific condition (we received a 404 because not all modules have finished starting up, but we think they _are_ starting up, and if you just give them more time and try again later then everything will probably be fine). 

Are there circumstances where Okapi could return a 404 for a reason _other_ than "some modules aren't ready yet"? If not, then I guess this is OK, but it still feels weird. But if yes, then this is definitely not OK. I'm not good at reading raml files, but it sure looks to me like https://github.com/folio-org/mod-users-bl/blob/e8f676fd7d56e46810a0d3c1aacffeec0358c5f3/ramls/mod-users-bl.raml can throw 404s for reasons other than "some modules aren't ready yet".

MikeTaylor:
@zburke I shared your concern when I came to review this, but I'm happy with it for three reasons:
1. The message is actually pretty generic: "System error. Please wait a few minutes and try again", not something like "Okapi has not yet finished loading modules. FOLIO should be ready in a few minutes".
2. The old message "Sorry, the information entered does not match our records" was actively misleading so this is an improvement even in scenarios where the 404 is not due to the system being only partially booted.
3. I don't see that the RAML at https://github.com/folio-org/mod-users-bl/blob/e8f676fd7d56e46810a0d3c1aacffeec0358c5f3/ramls/mod-users-bl.raml#L189-L220 allows for _any_ 404 response, so the only reason for a 404 is surely that Okapi can't map the path and so never gets as far as invoking mod-users-bl. (The 404s in the RAML are for other endpoints.)

zburke:
Thanks for explaining your thinking here, @MikeTaylor, and for explaining on Jira why Okapi and mod-users-bl can throw 404s. I still don't love this; it feels like replacing one not-great error message with another not-great error message, but it's clear that my initial reaction was an overreaction.

zburke (approved):

MikeTaylor:
It _is_ replacing one not-great error message with another not-great error message. But that's an improvement.

zburke:
@julianladisch thanks for merging this. Please remember to use "Squash and merge" for future PRs in UI repositories. This helps keeps the commit history clean.

PR #1316, opened 2023-06-02
STCOR-724 Use tenant query param for password reset

NikitaSedyx:
purpose:
for consortia we have one UI bundle for central tenant as a result there is no way to identify correct tenant to reset password
https://issues.folio.org/browse/STCOR-724

approach:
BE defines tenant in query param that can be used in headers, in case it's not defined (old links) bundle tenant is used instead


JohnC-80 (approved):

zburke (changes requested):
Please update the changelog and write a test for this new function. The easiest approach may be to externalize and export the function with a new signature like `getTenant(stripes, location)` so you can test it in isolation.

NikitaSedyx:
@zburke make sense, I left it without test as it's not so easy to test this part in scope of component as takes much more time to compare with initial task, will do

zburke (approved):

PR #1330, opened 2023-07-19
STCOR-729 update test deps for react v18

zburke:
Missed a few test-related deps in #1324. Additionally, convert from local deps/config to importing from jest-config-stripes. 

Refs [STCOR-729](https://issues.folio.org/browse/STCOR-729)


src/queries/useConfigurations.test.js
```diff
@@ -73,10 +73,14 @@ describe('Given useConfigurations', () => {
...
       );
 
-      const { result, waitFor } = renderHook(() => useConfigurations({}), { wrapper });
+      const { result } = renderHook(() => useConfigurations({}), { wrapper });
 
+      // waitFor retries until it stops catching an error
+      // so we convert isLoading from being truthy to being errory
       await waitFor(() => {
-        return !result.current.isLoading;
+        if (result.current.isLoading) {
```
vashjs, line 81: Maybe it makes sense  move this part to separate function? I see it used in 4 places

```
if (result.current.isLoading) {
  throw new Error('Kaboom');
}
```

vashjs (approved):

JohnC-80 (approved):

src/queries/useConfigurations.test.js
zburke: 🤷 I see your point, but there's not a lot going on here. I can imagine this also being written as
```
result.current.isLoading && throw new Error('Kaboom');
``` 
in which case we'd be replacing one line with ... one other line. It's a good thing to keep in mind though.

PR #1331, opened 2023-07-20
STCOR-733 Added new isUserInCentralTenant helper

BogdanDenis:
## Description
With many modules implementing Consortia features it's a good idea to move common code into shared modules
Added two functions:
 - `checkIfUserInCentralTenant` checks if user is in a central tenant
- `checkIfUserInMemberTenant` checks if user is in a member tenant

We need two because in a case when we're not in a Consortia environment both will return `false`.
So code like:
```js
if (!checkIfUserInCentralTenant(stripes)) {
  // do something only for member tenants
}
```
will do `something` on non-Consortia environment because `checkIfUserInCentralTenant` returns `false` when the interface is not present

## Issues
[STCOR-733](https://issues.folio.org/browse/STCOR-733)

### DIFF
--- a/index.js
+++ b/index.js
@@ -39,5 +39,6 @@ export {
 export { supportedLocales } from './src/loginServices';
 export { supportedNumberingSystems } from './src/loginServices';
 export { userLocaleConfig } from './src/loginServices';
+export * from './src/consortiaServices';
 export { default as queryLimit } from './src/queryLimit';
 export { default as init } from './src/init';
--- a/src/consortiaServices.js
+++ b/src/consortiaServices.js
@@ -1,0 +1,17 @@
+/* eslint-disable import/prefer-default-export */
+
+export function checkIfUserInCentralTenant(stripes) {
+  if (!stripes.hasInterface('consortia')) {
+    return false;
+  }
+
+  return stripes.okapi.tenant === stripes.user.user?.consortium?.centralTenantId;
+}
+
+export function checkIfUserInMemberTenant(stripes) {
+  if (!stripes.hasInterface('consortia')) {
+    return false;
+  }
+
+  return stripes.okapi.tenant !== stripes.user.user?.consortium?.centralTenantId;
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 * Use tenant query param for password reset. Refs STCOR-724.
 * Improve login error message when back-end not ready. Refs STCOR-723.
 * Replace `align-items: start` with `flex-start`; it is more widely supported. Refs STCOR-722.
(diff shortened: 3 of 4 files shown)


src/consortiaServices.js
```diff
@@ -0,0 +1,9 @@
+/* eslint-disable import/prefer-default-export */
+
+export function isUserInCentralTenant(stripes) {
```
Dmytro-Melnyshyn, line 3: I agree with this name too, but can we rename it to `checkIfUserInCentralTenant` to make it easier to name after import?

`const isUserInCentralTenant = checkIfUserInCentralTenant(stripes);`

index.js
```diff
@@ -39,5 +39,6 @@ export {
 export { supportedLocales } from './src/loginServices';
 export { supportedNumberingSystems } from './src/loginServices';
 export { userLocaleConfig } from './src/loginServices';
+export { isUserInCentralTenant } from './src/consortiaServices';
```
Dmytro-Melnyshyn, line 42: Please update the CHANGELOG file.

Dmytro-Melnyshyn (approved):

src/consortiaServices.js
zburke: 🤷 `const isCentralTenantUser = isUserInCentralTenant(stripes);`

mariia-aloshyna (approved):

src/consortiaServices.js
BogdanDenis: @zburke it's going to give a lint error:
image

zburke: I don't understand. I suggested slightly different names for the variable and the function specifically to avoid the scope overlap. But I don't have strong feelings here, and I'm sorry if weighing in has muddied the water. 

I generally like the convention of using verbs at the start of functions that return boolean values, e.g. `hasPermission` or `isAuthorized` or `isInCentralTenant`, but that's just a nice convention, not a rule that must be followed. You have to remember, I learned to program with vi during the previous millennium, before intellisense or code-completion 👴 . Back then it was super helpful to have these kinds of conventions because they saved you from having to look up the source to find out, "Does this return false on failure or throw an exception?" `isValid(stuff)` was a good hint that it would return a boolean.

zburke (approved):
LGTM

I've dragged my feet on this because I don't like the architecture of building consortial functionality directly into stripes-core, but that's an argument with the overall architecture, not this specific implementation.

BogdanDenis:
> LGTM
> 
> I've dragged my feet on this because I don't like the architecture of building consortial functionality directly into stripes-core, but that's an argument with the overall architecture, not this specific implementation.

Thank you for your feedback, @zburke. Do you think it's better to create a separate module `stripes-consortia-components` or something like that?

PR #1334, opened 2023-07-31
STCOR-736: Add `checkIfSharedRecord` and `checkIfLocalRecord` to consortia utils to store common helpers.

Dmytro-Melnyshyn:
## Purpose
Add `checkIfSharedRecord` and `checkIfLocalRecord` to consortia utils to store common helpers.

## Issues
[STCOR-736](https://issues.folio.org/browse/STCOR-736)

### DIFF
--- a/src/consortiaServices.js
+++ b/src/consortiaServices.js
@@ -15,3 +15,11 @@ export function checkIfUserInMemberTenant(stripes) {
 
   return stripes.okapi.tenant !== stripes.user.user?.consortium?.centralTenantId;
 }
+
+export function checkIfSharedRecord(stripes, source) {
+  return source.includes('CONSORTIUM-') || checkIfUserInCentralTenant(stripes);
+}
+
+export function checkIfLocalRecord(stripes, source) {
+  return ['MARC', 'FOLIO'].includes(source) && checkIfUserInMemberTenant(stripes);
+}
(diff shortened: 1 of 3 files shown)


src/consortiaServices.js
```diff
@@ -15,3 +15,11 @@ export function checkIfUserInMemberTenant(stripes) {
 
   return stripes.okapi.tenant !== stripes.user.user?.consortium?.centralTenantId;
 }
+
+export function checkIfSharedRecord(stripes, source) {
+  return source.includes('CONSORTIUM-') || checkIfUserInCentralTenant(stripes);
```
BogdanDenis, line 20: I think `checkIfSharedRecord` and `checkIfLocalRecord` should be in `ui-inventory` because `source` is specific only to Instance records. Please move it to some utility file in that module

BogdanDenis (changes requested):

src/consortiaServices.js
Dmytro-Melnyshyn: The `source` also will be applied in `ui-quick-marc` for authority records. It can be found in `authority-storage/authorities/id`.

BogdanDenis: For `ui-quick-marc` please see [my comment](https://github.com/folio-org/ui-quick-marc/pull/570#discussion_r1279061948)
IMO just two modules using one property is not enough to put this code in `stripes-core` and we can get await with passing `shared` as a prop or even duplicating two small functions
Also, having these utils in `stripes-core` implies that they can potentially be used for any type of record, but that's not the case.

PR #1337, opened 2023-08-25
Use Stripes CLI command

ryandberger:
since calling from stripes-core is deprecated.

### DIFF
--- a/doc/adding-permissions.md
+++ b/doc/adding-permissions.md
@@ -36,9 +36,9 @@ $
 
 ## Generate a module descriptor
 
-This can be done using a script provided with stripes-core (and which is used in back-end deployment):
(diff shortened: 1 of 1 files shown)


MikeTaylor (approved):
Thanks for catching this.

zburke:
@ryandberger , thanks for catching this! Within a repo, the best way to run this is 
```
yarn -s stripes mod descriptor --full > MD.json
```
to make sure you're invoking the local `stripes-cli` install, rather than a (potentially out of date/out of sync) global install. I can open a PR with this change, but stepping back even further, do you think these instructions still belong here in stripes-core, or should they be updated and moved (like other pages were) out to [`folio-org/stripes`](https://github.com/folio-org/stripes/tree/master/doc)?

PR #1340, opened 2023-09-07
STCOR-671 handle access-control via cookies

zburke:
## WIP warning

Handle access-control via HTTP-only cookies instead of storing `X-Okapi-Token` in local storage. By keeping the token in an HTTP-only cookie, this prevents any JS code from accessing and tampering with the value.

Notable changes:
* Sessions now time out instead of being valid indefinitely.
* Authentication requests are processed at `/bl-users/login-with-expiry` instead of `/bl-users/login`.
* "Activity" is tracked by an event listener that silently refreshes the cookie in the background, before it expires, as long as it detects activity (currently, mouse-clicks and keyboard events).

Refs [STCOR-671](https://issues.folio.org/browse/STCOR-671), [FOLIO-3627](https://issues.folio.org/browse/FOLIO-3627)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@
 * Bump `@formatjs/cli` to `^6.1.3`.
 * Forgot password and Forgot username : add placeholder to input box. Refs STCOR-728.
 * Include `yarn.lock`. Refs STCOR-679.
+* *BREAKING* Handle access-control via cookies. Refs STCOR-484.
 
 ## [9.0.0](https://github.com/folio-org/stripes-core/tree/v9.0.0) (2023-01-30)
 [Full Changelog](https://github.com/folio-org/stripes-core/compare/v8.3.0...v9.0.0)
--- a/package.json
+++ b/package.json
@@ -74,6 +74,7 @@
     "graphql": "^16.0.0",
     "history": "^4.6.3",
     "hoist-non-react-statics": "^3.3.0",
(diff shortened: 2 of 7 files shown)


zburke:
I am closing this in favor of an implementation based on service workers that intercepts fetch requests transparently handles RTR as needed. This has several benefits over the plan described above, 
> "Activity" is tracked by an event listener that silently refreshes the cookie in the background, before it expires, as long as it detects activity (currently, mouse-clicks and keyboard events).

Principally:
* tokens are refreshed as-needed and only when-needed, rather then being refreshed before they expire
* sessions timeout after a period of inactivity reaches the _RT_'s expiration instead of timing out with the AT's expiration

Additional details in [Slack#refresh-token-rotation](https://folio-project.slack.com/archives/C05JK4YH3A5/p1695841278260819)

zburke:
I am closing this in favor of an implementation based on service workers that intercepts fetch requests transparently handles RTR as needed. This has several benefits over the plan described above, 
> "Activity" is tracked by an event listener that silently refreshes the cookie in the background, before it expires, as long as it detects activity (currently, mouse-clicks and keyboard events).

Principally:
* tokens are refreshed as-needed and only when-needed, rather then being refreshed before they expire
* sessions timeout after a period of inactivity reaches the _RT_'s expiration instead of timing out with the AT's expiration

Additional details in [Slack#refresh-token-rotation](https://folio-project.slack.com/archives/C05JK4YH3A5/p1695841278260819)

PR #1346, opened 2023-10-04
STCOR-671 handle access-control via cookies

zburke:
Handle access-control via HTTP-only cookies instead of storing the JWT in local storage and providing it in the `X-Okapi-Token` header of fetch requests, and proxy all requests through a service worker that performs Refresh Token Rotation as needed to make sure the access-token remains fresh. 

Notable changes:

* Fetch requests are proxied by a Service Worker that intercepts them, validates that the Access Token is still valid, and performs token rotation (if it is not) before completing the original fetch.
* Sessions automatically end (i.e. the user is automatically logged out) when the Refresh Token expires. 
* Access control is managed by including an HTTP-only cookie with all requests. This means the Access Token formerly available in the response-header as `X-Okapi-Token` is never accessible to JS code. 

Replaces #1340. It was gross and I really don't want to talk about it. Let us never mention it again. 

* Requires folio-org/stripes-connect/pull/223
* Requires folio-org/stripes-smart-components/pull/1397
* Requires folio-org/stripes-webpack/pull/125

Refs [STCOR-671](https://issues.folio.org/browse/STCOR-671), [FOLIO-3627](https://issues.folio.org/browse/FOLIO-3627)


NikitaSedyx:
@zburke should we have so much logs here? - if there are removed in production bundle, it's fine to have them

zburke:
> @zburke should we have so much logs here? - if there are removed in production bundle, it's fine to have them

No, logs should be purged. I forgot to remove them when I moved this from "draft" to "ready". I wanted to use the stripes logger in `service-worker.js`, passed in from App.js through to registerServiceWorker and into the service-worker itself, but it didn't work. I could call `logger.log()` in the register function and it would work, but inside the service-worker none of the logger's functions were findable on the object. It was so weird.

src/service-worker.js
```diff
@@ -0,0 +1,421 @@
...
+ * and send a TOKEN_EXPIRATION event to clients that includes the new AT/RT
+ * expiration timestamps.
+ * @param {Event} event
+ * @returns Promise
+ * @throws if RTR fails
+ */
+export const rtr = async (event) => {
+  // console.log('-- (rtr-sw) ** RTR ...');
+
+  // if several fetches trigger rtr in a short window, all but the first will
```
steveellis, line 129: I'm not completely understanding what problem this would solve. Why would "several fetches trigger rtr in a short window"? Is it that lots of requests can arrive in the filter at once since the browser of course processes 8 or 10 at once in separate threads normally. If this is firing every

The reason I ask is because of the single use feature of the RT. You can only POST a single RT once. If you post it twice (say if two requests send at roughly the same time or in close proximity), mod-at will consider this a "leaked token" and log the user out of all of their devices.

I think this means that the RT should only be sent when it needs to be sent or just lazily in the background, which is at some interval that is less than its TTL. Occasionally it will result in a 401 because it expired.

src/service-worker.js
```diff
@@ -0,0 +1,421 @@
...
+      return Promise.resolve(new Response({}));
+    });
+};
+
+/**
+ * passThroughWithAT
+ * Given we believe the AT to be valid, pass the fetch through.
+ * If it fails, maybe our beliefs were wrong, maybe everything is wrong,
+ * maybe there is no God, or there are many gods, or god is a she, or
+ * she is a he, or Lou Reed is god. Or maybe we were just wrong about the
```
steveellis, line 277: May he rest in peace

src/service-worker.js
```diff
@@ -0,0 +1,421 @@
...
+self.addEventListener('message', async (event) => {
+  if (event.data.source === '@folio/stripes-core') {
+    console.info('-- (rtr-sw) reading', event.data);
+    if (event.data.type === 'OKAPI_CONFIG') {
+      okapiUrl = event.data.value.url;
+      okapiTenant = event.data.value.tenant;
+    }
+
+    // for reasons unclear to me, this does not work. the value comes through
+    // as a simple object rather than an instand of Logger. calling logger.log()
```
steveellis, line 402: Typo `instand`

steveellis: Going into proofreader mode here. Capital F at start of comment. Comma after Logger. Anyway I guess the comment is just WIP anyway. :)

src/service-worker.js
```diff
@@ -0,0 +1,421 @@
...
+ */
+self.addEventListener('message', async (event) => {
+  if (event.data.source === '@folio/stripes-core') {
+    console.info('-- (rtr-sw) reading', event.data);
+    if (event.data.type === 'OKAPI_CONFIG') {
+      okapiUrl = event.data.value.url;
+      okapiTenant = event.data.value.tenant;
+    }
+
+    // for reasons unclear to me, this does not work. the value comes through
```
steveellis, line 401: What doesn't work?

steveellis: Ok I get that what doesn't work is the whole point of the comment. But how is what is not working related to the goal of the PR?

src/service-worker.js
steveellis: I think what we want is for the service worker to not be too ambitious. He's lazy after all. But his laziness is a virtue. If he does his work well, in his lazy way, he doesn't need to be ambitious. Ambitious would mean that he's trying to solve every problem that comes his way just as soon as these problems arrive. But he's not like that. His reliability is how he makes up for his lack of ambition. He takes care of everything eventually. And because of that he doesn't need to be ambitious.

But why does he care about every request that comes his way then?

Only because just in case, something unforeseen has caused a problem that is outside of his control, he can send the user to the login screen, which is after all, much better than the alternative which would never to be sent anywhere at all.

src/service-worker.js
zburke: I was trying to leverage the stripes logger within the service worker so we could have access to the same filter tools but it doesn't work: the logger object gets munged on its way into the service worker. But don't worry about this. It's tangential to the core work here, and will probably just be removed before it's committed.

src/service-worker.js
zburke: Simultaneous fetches are common, e.g. the ui-users landing page fires several requests to fill in the values on its search form: patron groups, departments, tags, custom fields. The browser handles these in parallel streams, but as you pointed out, the RT is single use, so we need to guard against the situation where an AT has expired and then several requests arrive all at once. We need to let the first one through and force the others to wait. 

As to laziness, this code goes through great pains to be lazy, i.e. to implement just-in-time RTR, passing along requests as-is when it believes the AT is valid. But that's as lazy as we can be: the nature of a fetch event listener is that it hears about every fetch, so we need to handle all the surrounding edge cases: the expiration data in storage said the AT (or RT) was good but it turned out to be bad, here are five simultaneous fetches after a period of inactivity during which the AT expired, etc.

A timer-based approach is simpler, and doesn't need to involve service workers at all, but it limits sessions to the AT's expiration instead of the RT's. If we want sessions that last as long as the RT does, we have to implement JIT renewal as above, and it gets complicated pretty quickly.

BogdanDenis (approved):
Hey @zburke, looks good, don't have any issues with implementation.
One question though - do all UI modules **need** to remove `X-Okapi-Token` header from all requests, or will it just be ignored?

Dmytro-Melnyshyn (approved):

zburke:
@BogdanDenis, that's tricky to answer. Basically, if you pass the X-Okapi-Token in a request header AND in an AT via a cookie, then I think they have to match. All the UI code I saw in tickets linked to [FOLIO-3627](https://issues.folio.org/browse/FOLIO-3627) was implemented sorta like 
```
if (stripes.okapi.token) {
  headers['x-okapi-token': stripes.okapi.token;
}
``` 
which should be fine, because `stripes.okapi.token` will be undefined after this work merges.

PR #1347, opened 2023-10-06
STCOR-594 read ky timeout from stripes-config value

vashjs:
Based on the [STCOR-594](https://issues.folio.org/browse/STCOR-594), timout inside useOkapiKy should be configurable


zburke (approved):
LGTM. Thanks for finally picking this up! It's nice to chip away at some of these little changes that can have a big impact.

PR #1361, opened 2023-10-31
STCOR-574 rotate tokens well before they expire

zburke:
Rotate tokens slightly before they expire. This solves a problem in ui-data-export where every-five-second polling has caused some requests to land in a gap of about three seconds between when the AT was actually minted and when we stored it on the client side. i.e. it makes it much less likely that a token will expire in flight.

Refs [STCOR-574](https://issues.folio.org/browse/STCOR-754)

### DIFF
--- a/src/service-worker.js
+++ b/src/service-worker.js
@@ -63,6 +63,31 @@ const IS_ROTATING_RETRIES = 100;
 const IS_ROTATING_INTERVAL = 100;
 
 /**
+ * TTL_WINDOW
+ * How much of a token's TTL can elapse before it is considered expired?
+ * This helps us avoid a race-like condition where a token expires in the
+ * gap between when we check whether we think it's expired and when we use
+ * it to authorize a new request. Say the last RTR response took a long time
+ * to arrive, so it was generated at 12:34:56 but we didn't process it until
+ * 12:34:59. That could cause problems if (just totally hypothetically) we
+ * had an application (again, TOTALLY hypothetically) that was polling every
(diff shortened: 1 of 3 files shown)


src/service-worker.js
```diff
@@ -62,6 +62,31 @@ const IS_ROTATING_RETRIES = 100;
...
+ * cause RTR to happen a little early (i.e. a little more frequently) but that
+ * should be OK since it increases our confidence that when an AT accompanies
+ * the RTR request it is still valid.
+ *
+ * Value is a float, 0 to 1, inclusive. Closer to 0 means more frequent
+ * rotation; 1 means a token is valid up the very last moment of its TTL.
+ * 0.95 is just a SWAG at a "likely to be useful" value. Given a 600 second
+ * TTL (the current default for ATs) it corresponds to 570 seconds.
+ */
+export const TTL_WINDOW = 0.95;
```
steveellis, line 88: I don't see any harm in trying this value out. But you could probably be fine with a more conservative value like `.8` or even `.5` just to be safe. In other words trying to get closer to `.1` could be considered maybe a premature optimization.

Another way to think about it: the expiration times aren't necessarily goals or targets for when you should refresh. They're just expiration times for security and since refreshing is cheap you can do it any old time.

steveellis (approved):

src/service-worker.js
zburke: Effectively, before this change, the value was 1.0 🚀 😆

PR #1362, opened 2023-11-01
STCOR-753: Accessibility Issues - core components - Include label text in element's accessible name

UladzislauKutarkin:
Accessibility Issues - core components - Include label text in element's accessible name
Screenshot 2023-11-01 at 11 22 55
Also icon Apps renamed to `All applications` to improve accessibility.

### DIFF
--- a/src/components/MainNav/ProfileDropdown/ProfileDropdown.js
+++ b/src/components/MainNav/ProfileDropdown/ProfileDropdown.js
@@ -256,23 +256,13 @@ class ProfileDropdown extends Component {
   }
 
   renderProfileTriggerLabel = ({ open }) => {
-    const { okapi } = this.props.stripes;
-    const userData = this.getUserData();
-    const servicePointName = userData?.curServicePoint?.name;
-    const tenantName = userData?.tenants?.find(({ id }) => id === okapi.tenant)?.name;
-
-    const hasLabel = Boolean(servicePointName || tenantName);
-
     return (
-      hasLabel ? (
         <>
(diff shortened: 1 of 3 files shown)


zburke (commented):
Before and after examples would help me here. Please provide screenshots or examples of old text/new text.

It looks like this PR removes a substantial amount of dynamic information (the tenant name, the service point name) from the label and replaces it with the static string (the translation corresponding to `mainnav.myProfileAriaLabel`). Is that really the intent? Has this ticket been OK'ed by the teams who originally added that information to the button? 

Fix lint.

UladzislauKutarkin:
> Since this is a change to visible labels which affect button widths, this change really needs to be routed through appropriate design channels in order to resolve it in the best way that upholds accessibility and prioritizes the case for the current implementation. While better accessibility is something we would always like to see, we need to think about these changes a little bit more, and hopefully we can end up with some more prescriptive direction for developer action. The 'Apps' button is displayed in cases where there's not enough horizontal room for all of the app icons/buttons. Changing that text to 'All Applications' is a bit contrary to this functionality - and the button itself doesn't actually reveal 'all applications' - only those that overflow the space - so we need to decide on what is best there. Changing the label of the profile dropdown is a major end-user-experience changing item - definitely has to be run by a designer - and probably a community of users before we can accept this change.

Hi @JohnC-80 and @zburke . Let me discuss it with Magda (our PO). And I agree that we need to discuss it with design team. It sounds logical to me.

PR #1363, opened 2023-11-01
STCOR-756 correctly evaluate token lifespan

zburke:
The most important work here is fixing the bug from #1361 that incorrectly evaluated whether a token was still valid. The original implementation shrank the total lifespan of the token rather than shrinking only the period of its lifespan in the future.

Additional improvments here include:
* evaluate `navigator.serviceWorker` more cautiously; some browsers (e.g. Firefox in Incognito) may deny access to service workers. See [STCOR-757](https://issues.folio.org/browse/STCOR-757) for additional details.
* use `{ source, type, value }` shaped messages consistently when exchanging messages with the service worker.
* shrink the tokens' validity windows when receiving the `TOKEN_EXPIRATION` message instead of calculating the shorter window each time the token is evaluated. This is more efficient.
* await the promise returned by `postTokenExpiration` from `navigator.serviceWorker.ready` in order to prevent requests from being sent when the service worker isn't ready for them.
* use the `new Response()` constructor correctly, i.e. stringify the JSON value; otherwise clients will receive the string `[object Object]` instead of the empty JSON object, `{}`.

Note: the login test is turned off here due to the fact that the login flow invokes `navigator.serviceWorker.ready`, which returns a Promise that only returns when the service worker enters a ready state, but the karma build does not configure the service worker, hence this test times out every time. This is not great, but resolving it is a non-trivial task.

Refs [STCOR-756](https://issues.folio.org/browse/STCOR-756)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 * Ensure `<AppIcon>` is not cut off when app name is long. Refs STCOR-752.
 * Use cookies and RTR instead of directly handling the JWT. Refs STCOR-671, FOLIO-3627.
 * Shrink the token lifespan so we are less likely to use an expired one. Refs STCOR-754.
+* Correctly evaluate token lifespan; use consistent protocol for service worker messages. Refs STCOR-756.
 
 ## [10.0.0](https://github.com/folio-org/stripes-core/tree/v10.0.0) (2023-10-11)
 [Full Changelog](https://github.com/folio-org/stripes-core/compare/v9.0.0...v10.0.0)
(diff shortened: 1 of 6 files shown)


steveellis (approved):

steveellis:
@zburke  noticing the failed test:
image

Which says that the login page shouldn't be shown on successful login. Also noticed a few minutes ago on [the snapshot env](https://folio-snapshot.dev.folio.org/), the login page seemed to be shown after giving it the correct credentials. This happened multiple times in chrome on mac, and then a few minutes later the problem went away when I tried to login again.

PR #1369, opened 2023-11-09
STCOR-759 return non-okapi request fetches as-is

zburke:
Instead of catching errors from non-Okapi requests and converting them to rejected promises, just `return fetch()` straight up, whatever that response contains.

h/t @miketaylor for pointing me in this direction, and reading minified code to try to suss the problem. That's dedication!

Refs [STCOR-759](https://issues.folio.org/browse/STCOR-759), [UILDP-129](https://issues.folio.org/browse/UILDP-129)

### DIFF
--- a/src/service-worker.js
+++ b/src/service-worker.js
@@ -417,11 +417,7 @@ export const passThrough = (event, te, oUrl) => {
 
   // default: pass requests through to the network
   // console.log('-- (rtr-sw) passThrough NON-OKAPI', req.url)
-  return fetch(event.request, { credentials: 'include' })
-    .catch(e => {
-      console.error(e); // eslint-disable-line no-console
-      return Promise.reject(new Error(e));
-    });
+  return fetch(event.request);
 };
 
 /**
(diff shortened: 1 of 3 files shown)


JohnC-80 (approved):

src/service-worker.js
```diff
@@ -417,11 +417,7 @@ export const passThrough = (event, te, oUrl) => {
 
   // default: pass requests through to the network
   // console.log('-- (rtr-sw) passThrough NON-OKAPI', req.url)
-  return fetch(event.request, { credentials: 'include' })
-    .catch(e => {
-      console.error(e); // eslint-disable-line no-console
-      return Promise.reject(new Error(e));
-    });
+  return fetch(event.request);
```
MikeTaylor, line 420: Just checking: was it deliberate that as well as taking out this catch clause you also removed `credentials: 'include'`?

zburke: It was, based on [examples from the documentation](https://developer.mozilla.org/en-US/docs/Web/API/FetchEvent). Looking at that example again, I wonder if an _even better_ implementation here would be to do some of the request-sniffing that currently happens in `passThrough` in the event handler instead, returning immediately for requests that happen outside the RTR workflow, before calling `event.respondWith(...)`.

Thank you for mentioning this. I will investigate further.

PR #1381, opened 2023-12-12
Add back button support to multi-tenant workflow

ryandberger:
- When navigating from tenant selection screen to login screen, this allows user to navigate back with the browser back button.
- Removed duplicate `config` propType definition.
- Removed unused `LoginComponent`.


zburke (approved):
LGTM

PR #1382, opened 2023-12-14
STCOR-753: Add aria-label for `ProfileDropdown.js`

UladzislauKutarkin:
[STCOR-753](https://issues.folio.org/browse/STCOR-753)
Here I reproduced steps that was provided by @JohnC-80 in comments section.
We add `aria-label` for `ProfileDropdown`. Also, we removed `aria-label` from `App dropdown`.


src/components/MainNav/ProfileDropdown/ProfileDropdown.js
```diff
@@ -266,7 +266,7 @@ class ProfileDropdown extends Component {
     return (
       hasLabel ? (
         <>
-          <span className={css.button__label}>
+          <span className={css.button__label} aria-label={`${tenantName} ${servicePointName} profile`}>
```
Dmitriy-Litvinenko, line 269: Hello @UladzislauKutarkin , @JohnC-80 Could you please clarify should we move this string to translate message for localization ` profile` and pass `tenantName` and `servicePointName` as value. Example https://github.com/folio-org/stripes-smart-components/blob/master/lib/ChangeDueDateDialog/LoanList.js#L99

zburke: Agreed, @Dmitriy-Litvinenko: this message must be passed through react-intl just like the example you link to. This will allow "profile" to be correctly localized.

zburke (changes requested):
As @Dmitriy-Litvinenko noted, the new `aria-label` must be handle via react-intl so it can be localized.

src/components/MainNav/ProfileDropdown/ProfileDropdown.js
UladzislauKutarkin: Done! Good point

UladzislauKutarkin: @Dmitriy-Litvinenko please take a look again!

Dmitriy-Litvinenko (approved):

Terala-Priyanka (approved):

src/components/MainNav/ProfileDropdown/ProfileDropdown.js
```diff
@@ -266,7 +267,14 @@ class ProfileDropdown extends Component {
...
       hasLabel ? (
         <>
-          <span className={css.button__label}>
+          <span
+            className={css.button__label}
+            aria-label={intl.formatMessage({
+              id: 'stripes-core.mainnav.profile.ariaLabel',
+              tenantName,
+              servicePointName,
+            })}
```
zburke, line 276: Not quite; this needs to be 
```
aria-label={intl.formatMessage(
  { id: 'stripes-core.mainnav.profile.ariaLabel' },
  { tenantName, servicePointName },
)}
```

zburke (changes requested):
Close, but not quite right. `formatMessage()` requires two arguments.

UladzislauKutarkin:
> Sorry, still not quite right, and sorry I didn't catch this sooner: closely reviewing @JohnC-80's [request on STCOR-753](https://issues.folio.org/browse/STCOR-753?focusedCommentId=196180&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-196180), this PR must "expand the accessible label" of the profile dropdown; instead, it adds a new label.
> 
> Instead of adding an `aria-label` in `renderProfileTriggerLabel` (line 272), you need to update `renderProfileTrigger` and provide a correct `ariaLabel` to the `<NavButton>` there (line 248).

@zburke please take a look. I made requested changes

zburke (changes requested):
@UladzislauKutarkin, is there a reason you created a new translation key `mainnav.profile.ariaLabel` instead of repurposing the existing one, `mainnav.myProfileAriaLabel`? 

Either reuse the existing one (preferable) or keep the new one and remove the old one.

UladzislauKutarkin:
> mainnav.myProfileAriaLabel

UladzislauKutarkin:
> @UladzislauKutarkin, is there a reason you created a new translation key `mainnav.profile.ariaLabel` instead of repurposing the existing one, `mainnav.myProfileAriaLabel`?
> 
> Either reuse the existing one (preferable) or keep the new one and remove the old one.

Ye, you are right, I made new one key for prev implementation. Since we made new approach, I returned old translation key and deleted new one

zburke (approved):
LGTM! 👏 👏 👏

PR #1383, opened 2023-12-18
STCOR-770: Export getEventHandler to be able to create events in other modules.

Dmytro-Melnyshyn:
## Purpose
Export `getEventHandler` to be able to create events in other modules.

## Related PRs
https://github.com/folio-org/ui-consortia-settings/pull/101
https://github.com/folio-org/ui-inventory/pull/2373

## Issues
[STCOR-770](https://issues.folio.org/browse/STCOR-770)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 * Export `unregisterServiceWorker` to eliminate zombie service workers. Refs FOLIO-3627.
 * Fix duplicated "FOLIO" in document title in some cases. Refs STCOR-767.
 * Refactor away from `color()` function. Refs STCOR-768.
+* Export `getEventHandler` to be able to create events in other modules. Refs STCOR-770.
 
 ## [10.0.0](https://github.com/folio-org/stripes-core/tree/v10.0.0) (2023-10-11)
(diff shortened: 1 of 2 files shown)


skomorokh:
Isn't this already exposed via registering your module as `"actsAs": [..., "handler"]` in `package.json`?

mariia-aloshyna (approved):

zburke (approved):

skomorokh:
Aha, sorry, I'd understood the `handler` module type to be a pub/sub system but it seems that there are only a few events from stripes-core and some special package.json `link` thing and no way to publish a general event.

So effectively your code here is the new Stripes API for publishing events. I think we should then actually export a method to do that rather than leave it up to each module to loop through subscribers. From there we might later improve the implementation to support working across tabs/windows, etc.

Taking a closer look, it doesn't necessarily seem like you need to create a new event mechanism to solve this though? The current affiliation is persisted to the back end and comes in via react-query.  So modules should be fetching the current value, or if caching is in use, the cache should be invalidated by the control to switch affiliation.  Thus it should be possible to reset the search query when that value changes by storing the affiliation alongside the query and resetting if it no longer matches?

PR #1384, opened 2023-12-19
STCOR-771 load DayJS locale alongside moment locale - loginServices

JohnC-80:
### [STCOR-771](https://issues.folio.org/browse/STCOR-771)
Efforts to taper and ultimately remove our dependency on the sunset `moment` library begin with dynamically loading and setting the correct DayJS locale file. Applications that use `moment` can begin their migration to DayJS.

### Approach 
DayJS brings along a JSON file that houses a list of all its included locales. We can check this file before attempting to load a new locale or broadcast an error to the console.
This PR implements a separate function `loadDayJSLocale` that accepts the locale string.
The rest of the 'magic' is accompanied by webpack 'magic comments' - turning webpack's dynamic import in to something like a `require.context()` that tell webpack how to name chunknames of locale files for loading and to exclude the types (`d.ts`) file in the directory because it could potentially fail a build.


zburke (changes requested):
* Please update CHANGELOG.md
* Adding a peer-dep is a breaking change. I'm OK with that, but we need to socialize this change on Slack#stripes-updates before we merge it. 
* I understand from the current PR description that this change lays the groundwork for future Day.js-based work directly in applications, but it isn't clear to me what the "future work" is. Some additional context or an example would be helpful, e.g. "it means applications can use `dayjs()`"

JohnC-80:
As per conversation in stripes-architecture, the migration path for this will only mean importing `dayjs` from `stripes-components`, as we'll be re-exporting a [pre-extended version of the library](https://github.com/folio-org/stripes-components/pull/2192/files#diff-be49f0a3c890c3f02a4651bbd78b0115e8631dafae3c83e21b9ad1145a848bdcR16-R30).
This means that the platform will not need to add `dayjs` to its dependency tree, nor will ui-modules.

If they import stripes-components, they'll be able to access `dayjs` as well.

zburke (approved):
It would be great to supplement the function-header comments for `loadTranslations` to explain what `loadDayJSLocale()` is doing in there and what the long-term migration plan is, i.e. that eventually the `moment` stuff is going away and `DayJS` will be the only thing available.

PR #1385, opened 2023-12-20
STCOR-773 show legacy or application-based discovery info

zburke:
Discovery needs to remain backwards compatible with Okapi APIs; ditto the `/settings/about` page that displays module and interface version information.

There is not really as much work here as it appears. All the new components were split out of `About` in order to allow sub-sections to be reused with both application-based and module-based discovery information. Likewise, `loginServices.js` and `discoveryServices.js` were modestly refactored to handled both APIs.

And there are Jest/RTL tests to replace the BTOG test that could not be easily updated to handle the new APIs since its `stripes-config` stub is part of `@folio/stripes-cli` instead of being declared locally.

Refs [STCOR-773](https://issues.folio.org/browse/STCOR-773)


ryandberger (approved):
Looks good overall! Appreciate the organization of the refactoring and the jest tests!

zburke:
Screenshot 2024-01-09 at 10 56 13 AM (2)

PR #1388, opened 2024-01-10
Graceful handling when provides array is null in API response

ryandberger:
Previously only showed first application in the list before quietly failing on a null exception. Now shows all applications returned in API response.

image

### DIFF
--- a/src/discoverServices.js
+++ b/src/discoverServices.js
@@ -246,9 +246,9 @@ export function discoveryReducer(state = {}, action) {
             modules: action.data.moduleDescriptors.map((d) => {
               return {
                 name: d.id,
-                interfaces: d.provides.map((i) => {
+                interfaces: d.provides?.map((i) => {
                   return { name: i.id + ' ' + i.version };
-                }),
+                }) || [],
               };
             }),
           },


craigmcnally (approved):
Looks about as straightforward as it gets to me

zburke (changes requested):
On closer inspection, the thing you're fixing here in [STCOR-786](https://issues.folio.org/browse/STCOR-786) is indeed the same underlying cause as [STCOR-788](https://issues.folio.org/browse/STCOR-788) so I linked them and closed the latter as a dupe. There was a small additional requirement in STCOR-788 (log to the JS console when the value is missing) that we need to add here. 

A few bookkeeping things: 
* commit subjects/PR titles should start with the Jira slug. This links Jira and GitHub, and it makes the log _much_ more helpful, both on GitHub (where commit messages are truncated just to their subjects) and at the CLI with `--abbrev-commit`. It's courteous to link to the Jira in the body of the PR as well, e.g. `Refs [STCOR-786](https://issues.folio.org/browse/STCOR-786)`. Please make sure to keep this format, with the Jira slug at the start of the subject, when you squash and merge.
* please add a line to CHANGELOG that describes the fix (often you can use the commit subject) and references the Jira ticket.

PR #1392, opened 2024-01-17
STCOR-793 avoid potentially unsafe optional chaining

zburke:
Optional chaining may, of course, return `undefined`, causing `foo?.permissions.map(...)` to throw a TypeError since you cannot call `map` on `undefined`.

Refs [STCOR-793](https://issues.folio.org/browse/STCOR-793)


src/loginServices.js
```diff
@@ -397,9 +397,11 @@ export function spreadUserWithPerms(userWithPerms) {
     ...userWithPerms?.user?.personal,
   };
 
-  // remap data's array of permission-names to set with
-  // permission-names for keys and `true` for values
-  const perms = Object.assign({}, ...userWithPerms?.permissions?.permissions.map(p => ({ [p.permissionName]: true })));
+  // remap permissions from [foo, bar, bat] to { foo: true, bar: true, ... }
+  const perms = userWithPerms?.permissions?.permissions.reduce((acc, i) => {
```
BogdanDenis, line 401: I think `perms` can still be `undefined` in some cases
In this example which is structurally identical to real code `result` will be undefined, which is a surprise to me because I would have expected an exception...
```
const a = {};

const result = a?.b?.c.reduce(acc => acc, {});
```

Anyways, looks like we need a default value for `perms`

BogdanDenis (changes requested):

src/loginServices.js
zburke: It's OK if `perms` ends up undefined, but we need to avoid throwing an error when parsing that nested structure when it is sparsely populated. I just noticed the last element before `.reduce` is not protected with `?`, and that still fails in my terminal:
```
let userWithPerms = { permissions: { foo: 'bar' }}
const perms = userWithPerms?.permissions?.permissions.reduce((acc, i) => { acc[i.permissionName] = true; return acc; }, {});
Uncaught TypeError: Cannot read properties of undefined (reading 'reduce')
```
It's surprising to me that SonarCloud is OK with it 😕

BogdanDenis (approved):

JohnC-80 (approved):

PR #1400, opened 2024-01-22
STCOR-794: Disable Continue button when no tenant selected

ryandberger:
- Fixes [STCOR-794](https://issues.folio.org/browse/STCOR-794)
- Use an element Ref to determine if submit button should be disabled since using `okapi.tenant` retains the previous value when navigating back to the tenant selection screen.


zburke (approved):
LGTM

PR #1408, opened 2024-01-24
Still more sso test

JohnC-80:


JohnC-80:
Added jest test/removed BTOG test in a separate PR

PR #1409, opened 2024-01-25
STCOR-801 Move async localforage.clear to afterEach

JohnC-80:
The call to `localForage.clear` returns a promise, so it's best to handle it within an async call so that tests can wait for it. This seems to explain some difference we're seeing in timing between local test runs and CI such as the test runs from https://github.com/folio-org/stripes-core/pull/1392


zburke (approved):
Let's try it. Since `afterEach` can be properly `async`, maybe this is our ticket to real reliability. Nice sleuthing! 🕵️

PR #1410, opened 2024-01-26
STCOR-796 replace x-okapi-token credentials with RTR and cookies

zburke:
Move auth tokens into HTTP-only cookies and implement refresh token rotation (STCOR-671) by overriding `global.fetch` and `global.XMLHttpRequest`, disabling login when cookies are disabled (STCOR-762). This functionality is implemented behind an opt-in feature-flag (STCOR-763). 

Okapi and Keycloak do not handle the same situations in the same ways. Changes from the original implementation in PR #1376:

* When a token is missing:
  * Okapi sends a 400 text/plain response
  * Keycloak sends a 401 application/json response
* Keycloak authentication includes the extra step of exchanging the OTP
  for the AT/RT and that request needs the `credentials` and `mode`
  options
* Some `loginServices` functions now retrieve the host the access from
  the `stripes-config` import instead of a function argument

Refs [STCOR-796](https://issues.folio.org/browse/STCOR-796), [STCOR-671](https://issues.folio.org/browse/STCOR-671)


ryandberger (approved):
Lot to take in here, but approving knowing that most of this is already in master branch and the subtle changes have been tested against keycloak.

aidynoJ (approved):

PR #1411, opened 2024-01-30
Add Jest tests for SSO Landing - remove BTOG tests for the component.

JohnC-80:
https://issues.folio.org/browse/STCOR-798

We tried a handful of things, but none lead to the stability that we need -
- supplying adequate data from mirage
- moving an async function to `afterEach`

The tested component comes very close to the top of the DOM hierarchy, so there isn't much for elements to wait for, aside from the element itself.
The underlying logic affects local storage, which seems the biggest hangup - sometimes the localstorage would clear, sometimes it would not - which would mean the tested component wouldn't even render - which would mean test failure :(

This PR removes the flaky tests and adds an isolated unit test written with Jest in its place.


zburke (approved):
LGTM, but do we need the inner describe blocks or can they be removed and have those descriptions directly on the it-blocks? It's a little weird to me to have "processes login" under "displays error" given that the (presumptive) error here is "login failed".

zburke:
<petty>When you squash and merge PRs, please make sure the commit subject gets updated to include the Jira at the beginning. Release scripts that update Jira tickets to correctly set their release-version rely on this, and the commit log is a lot easier to parse when it looks like
* STCOR-123 implement foo
* STCOR-135 implement bar

PR #1415, opened 2024-02-09
STCOR-769 Utilize the 'tenant' procured through the SSO login process

usavkov-epam:
## Purpose
To support Single Sign-On (SSO) authorization in consortium mode, it's necessary to explicitly pass the tenant in the request header to load data.

## Approach
The tenant, which is obtained from the returned SSO token for legacy mode or from the cookie or search parameter for RTR mode, should be passed as an argument to the `requestUserWithPermsDeb` function.

## Refs
FE:
Jira - https://folio-org.atlassian.net/browse/STCOR-769

BE:
Jira (Add tenant to search params) - https://folio-org.atlassian.net/browse/MODLOGSAML-185


src/components/SSOLanding.js
```diff
@@ -28,24 +29,35 @@ const SSOLanding = () => {
...
+        No <code>tenant</code> in cookie or query parameter
       </div>
     );
   }
 
   return (
     <div data-test-sso-success>
       <p>
-        Logged in with token <tt>{token}</tt> from {getParams()?.ssoToken ? 'param' : 'cookie'}.
+        Logged in with tenant <tt>{tenant}</tt>.
```
NikitaSedyx, line 60: I would propose to show spinner here instead as for `useSecureTokens=true` we'll have no token available and message says nothing in general

src/components/Root/FFetch.js
```diff
@@ -248,7 +251,7 @@ export class FFetch {
       }
 
       // AT is valid or unnecessary; execute the fetch
-      if (this.isPermissibleRequest(resource, this.tokenExpiration, okapi.url)) {
+      if (rtrIgnore || this.isPermissibleRequest(resource, this.tokenExpiration, okapi.url)) {
```
NikitaSedyx, line 254: it's the easiest fix to make SSO work as currently request to fetch permissions and after set up okapi session is blocked https://github.com/folio-org/stripes-core/pull/1415/files#diff-bb2e7a2fa6dcb390a09c12b8e1ee5c0f3354c479d0e537252e7f6328a029d0a6R266

https://github.com/folio-org/stripes-core/pull/1415/files#diff-c9d695b1d57404dfed01196c8cf479c0e4cd48583f6c78e7db23402582846864R751

NikitaSedyx: we can create tokenExp in okapi session before permissions request, but after it will create send refresh request anyway as createOkapiSession parses response - as for me it's better to not initiate partial okapi session before perms request

src/loginServices.js
```diff
@@ -748,9 +748,14 @@ export function requestLogin(okapiUrl, store, tenant, data) {
...
  */
 function fetchUserWithPerms(okapiUrl, tenant, token) {
+  const rtrIgnore = !token;
+
   return fetch(
     `${okapiUrl}/bl-users/_self?expandPermissions=true&fullPermissions=true`,
-    { headers: getHeaders(tenant, token) },
+    {
+      headers: getHeaders(tenant, token),
+      rtrIgnore,
```
JohnC-80, line 757: This would build in an rtrIgnore flag that could be added to any fetch call... this doesn't seem like the right approach.

Could `/bl-users/_self` just be added to the `permissible` list?

NikitaSedyx: I don't think we can as we need to refresh tokens when users restore page, but for SSO is exception because we get session right after login

NikitaSedyx: > that could be added to any fetch call

that's true, but as I mentioned in another comment, we need to create temporary okapiSession

but in general `rtrIgnore` just simplifies skipping RTR if really required as in any place it's possible to change token expiration values, why can it be done? I have no idea )

NikitaSedyx:
@zburke sure, tests were not fixed because it was not clear that this approach is acceptable, I'll take care of them

what do you think about other comments to update view of this component?

src/components/SSOLanding.js
zburke: @NikitaSedyx's suggestion here is solid and would be a nice improvement over this unlocalized message (😬 ), but if you want to minimize changes to the existing implementation, I think your change is OK as-is.

src/components/SSOLanding.js
```diff
@@ -28,24 +29,35 @@ const SSOLanding = () => {
...
     }
-  }, [token, store]);
+  }, [tenant, token, store]);
 
-  if (!token) {
+  if (!tenant) {
     return (
       <div data-test-sso-error>
-        No <code>ssoToken</code> cookie or query parameter
+        No <code>tenant</code> in cookie or query parameter
```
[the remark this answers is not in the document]
NikitaSedyx: I did smth close to it, redirect to login page, but with auth error so user is notified and from login user can do again - I believe error message should be improved a bit

src/components/SSOLanding/useSSOSession.js
```diff
@@ -0,0 +1,73 @@
...
+  const [isFailed, setIsFailed] = useState(false);
+  const store = useStore();
+  const dispatch = useDispatch();
+
+  const location = useLocation();
+  const [cookies] = useCookies(['ssoToken']);
+
+  const params = getParams(location);
+
+  const token = getToken(cookies, params);
```
steveellis, line 44: @NikitaSedyx and @zburke I think this means that this code still assumes that a token can be obtained from the cookie or the querystring, which would be true if we're running in "legacy mode" only not in RTR-land. So I think this means that https://folio-org.atlassian.net/browse/STCOR-764 isn't going to be resolved by this PR unless I'm missing something.

src/components/SSOLanding/useSSOSession.js
```diff
@@ -0,0 +1,73 @@
...
+  const tenant = getTenant(params, token);
+
+  useEffect(() => {
+    requestUserWithPerms(okapi.url, store, tenant, token)
```
steveellis, line 48: @zburke @NikitaSedyx -- Or is it that this function is happy with `undefined` for `token` and will use the new httpOnly cookies in the request?

src/components/SSOLanding/useSSOSession.js
zburke: @steveellis, you're correct that this function will return undefined under some circumstances, but that's fine. We could include checks, like the on lines 27-29, all over the place, but since it's an optional argument in the functions we pass it to, i.e. those functions already contain logic like `if (!token) {...}`, there is no value including that logic again here.

src/components/SSOLanding/useSSOSession.js
zburke: Correct, @steveellis

zburke (approved):

src/components/SSOLanding/useSSOSession.js
NikitaSedyx: it's correct that we support legacy and RTR approaches, but they now have common validation - _self response, if it fails so we redirect user to logic as there is no active session

I fully removed validation of tokens, query params etc, so we can support both authentication mechanisms (I had no goal to remove legacy as stripes supports it) with the same source of truth to know that session is valid or not

if valid token is provided in query params or httpOnly=false cookie - we set up UI session
if valid AT is provided in cookies - we set up UI session
if invalid token is provided in query params or httpOnly=false cookie - redirect to login with error
if invalid AT is provided in cookies (or not provided at all) - redirect to login with error

_self request says us that we can set up session or not

steveellis: Ok great guys, makes sense. Then I think this is fine. When we merge this it can be more easily tested by @renatsafiulin in https://folio-org.atlassian.net/browse/RANCHER-1250.

steveellis: I closed https://folio-org.atlassian.net/browse/STCOR-764 based on this.

src/constants/ssoErrorCodes/ssoErrorCodes.js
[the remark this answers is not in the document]
zburke: Yeah, fair, "following existing pattern" is legit. Your call on the best option here:
* the cleaner thing that is different than its peers
* the ugly thing that is consistent with its peers

Sorry to be cranky at you when you're just following existing examples. It's fine with me if you want to just file a clean up ticket to fix all this stuff in one swoop.

PR #1420, opened 2024-02-13
Add opt-in "Really logout?" stripes.config.js option

ryandberger:
- Fulfills [STCOR-803](https://folio-org.atlassian.net/browse/STCOR-803).
- If `config.confirmLogout` is truthy and `okapi.authnUrl` is set, then user is brought to confirm logout page, otherwise logout happen immediately.
- Small lint fix.

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -51,7 +51,7 @@ export default class StripesCore extends Component {
         logger={this.logger}
         config={config}
         actionNames={this.actionNames}
-        disableAuth={(config && config.disableAuth) || false}
+        disableAuth={(config?.disableAuth) || false}
         {...props}
       />
     );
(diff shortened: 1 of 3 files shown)


aidynoJ (approved):

zburke (commented):
Does this actually work, i.e. is stripes usable when you choose "return to application"? MainNav calls `returnToLogin` before pushing `/logout` (which fires the redirect via RootWithIntl) onto the stack. That means stripes will be totally cleared out when you return. Even if it renders, I would expect it to implode as soon as you try to do something real. 

And a few bookkeeping things: 
* Please start every commit subject and PR title with the Jira slug. The release scripts we use rely on parsing the subject to keep correctly set the "fix version" field in Jira, which is the only way POs assess whether a given change made it into a given release. 
* Please update the CHANGELOG, especially when you're adding a new feature.

PR #1423, opened 2024-02-14
STCOR-815 Add SSO Login path to permissible resources in RTR code.

JohnC-80:
Attempting an inital SSO login (no http cookie present) in browser incognito mode throws a RTR token error. Including the SSO path in `permissibleResources` circumvents the token check from this login.


zburke (changes requested):
Lint doesn't like `debugger`. Please include a summary in the PR title. The fix itself LGTM.

JohnC-80:
This will be handled as part of #1415 closing.

PR #1426, opened 2024-02-16
[STCOR-803] Simplify logout workflow to bypass keycloak confirmation page

ryandberger:
- Fulfills [STCOR-803](https://folio-org.atlassian.net/browse/STCOR-803).
- Remove logout workflow which redirects to keycloak logout confirm page. Logout will always happen immediately and redirect user back to tenant selection (or login page if single tenant).
- Log error if logout request fails. At this point it was decided not to take further action since session expires within 5 minutes if not refreshed.


src/RootWithIntl.js
```diff
@@ -214,7 +208,7 @@ class RootWithIntl extends React.Component {
                         <TitledRoute
                           name="logout"
                           path="/logout"
-                          component={renderLogoutComponent(this.props.stripes)}
+                          component={renderLogoutComponent()}
```
zburke, line 211: This is fine, but we can we simplify this even further:
```
 component={<InternalRedirect to="/" />}
```
Then we can omit `renderLogoutComponent`, which doesn't do much any longer.

src/loginServices.js
```diff
@@ -429,7 +429,10 @@ export async function logout(okapiUrl, store) {
     credentials: 'include'
   })
     .then(localforage.removeItem(SESSION_NAME))
-    .then(localforage.removeItem('loginResponse'));
+    .then(localforage.removeItem('loginResponse'))
+    .catch((error) => {
+      logger.log('logout', 'Error logging out', error)
```
zburke, line 453: What do you think about an actual `console.error()` here so this message is always displayed, regardless of how stripes' logging categories are configured. You'll need `// eslint-disable ...` to keep lint from whining.

zburke (approved):
LGTM, though I think there is room for a little more cleanup if you want. Please clean up lint errors before you merge.

src/loginServices.js
ryandberger: Sounds good. I'll make the change.

PR #1430, opened 2024-02-19
STCOR-817 add token checks to XHR prior to rotating token

JohnC-80:
Most data requests throughout FOLIO are handled via `fetch` calls... but occasionally we can have an abundance of XHR's being sent out - (like when `fetch` polyfills get used - should not be necessary, but it happens...) so we need to treat these in similar fashion in regards to token rotation - checking the access token, refreshing expiry, refreshing tokens only when necessary, etc... This PR adds similar token checks to XHR requests to shore up support there.


src/components/Root/FXHR.test.js
```diff
@@ -47,4 +55,29 @@ describe('FXHR', () => {
...
+
+  it('Handles Token Errors', () => {
+    rtr.mockImplementationOnce(() => { throw new RTRError('test rotation failure'); });
+    try {
+      testXHR.addEventListener('abort', mockHandler);
+      testXHR.open('POST', 'okapiUrl');
+      testXHR.send(new ArrayBuffer(8));
+    } catch (err) {
+      expect(err instanceof RTRError).toBeTrue;
+    }
```
zburke, line 81: Even if the `catch` clause never executes, this test will pass because it won't contain any failing expectations. Maybe set it up like 
```
const err = null;
try { ... }
catch(e => { err = e; }
finally { expect(err) }
```
Also `s/toBeTrue/toBeTruthy()/` or `toBe(true)`; `toBeTrue` is not a function and is not invoked as a function.

src/components/Root/FXHR.js
```diff
@@ -16,12 +17,24 @@ export default (deps) => {
...
+        if (isValidRT(this.FFetchContext.tokenExpiration, logger)) {
+          logger.log('rtr', 'local refresh token valid, sending XHR...');
+          super.send(payload);
+        } else {
+          await rtr(this.FFetchContext)
+            .then(() => {
+              logger.log('rtr', 'local token refreshed, sending XHR...');
+              super.send(payload);
+            });
+        }
```
zburke, line 37: This doesn't seem right to me. If I'm reading correctly, it only does rtr when it thinks the RT is expired. That might work because we shrink the TTL when we store it to avoid rotating tokens right at the edge of their validity, but still... I think we want logic similar to [FFetch::ffetch](https://github.com/folio-org/stripes-core/blob/238b09157f2f0b085bce8891080e3c0cde127353/src/components/Root/FFetch.js#L233-270):
```
if (localAT is expired)
  refetch from storage in case somebody else updated it
if (localAT is valid)
  super.send()
else if (localRT is valid)
  await rtr()
  super.send()
else 
  all is lost; dispatch an RTR_ERROR_EVENT
```

zburke (changes requested):

JohnC-80:
Alright. Updated to be more in the shape of `Ffetch` where we double check the AT, possibly send, or check the RT and refresh then maybe send... and improved the tests accordingly. 😓

zburke (approved):
Sweet! This will really cut down on RTR traffic from XHR requests. I didn't realize this conservative approach to calling `rtr()` was missing in the first draft so I'm really glad we're getting it in now.

Only nit: I think it's uncommon to see `await ... .then`. It might be worth refactoring like
```
try {
  await Promise();
  stuff();
} catch (err) {
}
```
OTOH, this is very similar to FFetch, and maybe maintaining that parity is more important. Dunno.

JohnC-80:
Yeah, you're right... the nature with this is slightly different from `fetch` where with `fetch`, the promise is actually returned rather than just calling `xhr.send()`. Applications/libraries using XHR for things will , conventionally, already be wrapping it in their own promise.

PR #1435, opened 2024-03-04
UIROLES-43 Add parameter overrides

ryandberger:
- Allow for parameter `idName` to be overridden to allow for using other IDs when `id` alone is too ambiguous (such as a case where `userId`, `roleId`, etc is needed).
- Allow for parameter `limit` to be overridden if a different response limit is needed.


src/queries/useChunkedCQLFetch.js
```diff
@@ -33,10 +33,12 @@ const useChunkedCQLFetch = ({
   CONCURRENT_REQUESTS = CONCURRENT_REQUESTS_DEFAULT, // Number of requests to make concurrently
   endpoint, // endpoint to hit to fetch items
   generateQueryKey, // Passed function to allow customised query keys
-  ids: passedIds, // List of ids to fetch
+  ids: passedIds, // List of IDs to fetch
+  idName = 'id', // Named ID field to use in the CQL query (i.e. id or userId)
+  limit = 1000, // Item limit to fetch on each request
```
zburke, line 38: It's odd to me that you would ever need this given you are running a query like `id==(foo or bar or bat)`, i.e. retrieving items by an identifier in STEP_SIZE chunks. That is, the reason we're chunking queries with a small step-size is to prevent the query string from growing too-long, and 1000 is already waaaay longer than 60. So I'm just curious about the use-case here. Can you explain?

zburke (approved):
Code changes look perfect, but I have a bunch of petty complaints: 

* Please configure your editor to trim whitespace to avoid the complaints from lint :) 
* On the paperwork side, a PR in stripes-core needs a ticket in STCOR. If all you have is a UIROLES ticket, there's no way to tell from looking at Jira what version of stripes-core contains this fix, because the `fixVersion` field will contain ui-authorization-roles versions, not stripes-core versions. 
* Since there is nothing specific to the Eureka work, consider re-targeting this PR at #master and then cherry-picking it onto the #keycloak branch. That would allow the community to benefit from this work sooner. 
* Please update CHANGELOG.md when you add a new feature.

src/queries/useChunkedCQLFetch.js
zburke: Now that I've grokked folio-org/ui-authorization-roles/pull/32, I understand. If there's a table FOO with PK fooId and a table BAR with PK barId and a join table with multipart PK (fooId, barId), and you're using this tool to query the join table, you would expect the result to contain many more rows than the input query list. 

Sorry that took me so long. I am a bear of very little brain and big queries bother me.

zburke: @ryandberger, now that I understand it, I see how it's valuable! Up to you whether to leave it in or handle it as a separate feature.

ryandberger: I'll leave it in, thanks!

ryandberger:
Closing in favor of https://github.com/folio-org/stripes-core/pull/1438

zburke:
@ryandberger , this PR was targeted at `#keycloak`, but #1438 merged to `#master`. Do we need to cherry-pick that commit onto this branch now?

ryandberger:
> @ryandberger , this PR was targeted at `#keycloak`, but #1438 merged to `#master`. Do we need to cherry-pick that commit onto this branch now?

I did the cherry-pick here: https://github.com/folio-org/stripes-core/commit/ad24fa7defe709d977538dfc1d43aa8118bc673e. Should have done it as a PR for visibility and will do so in the future.

PR #1438, opened 2024-03-06
STCOR-821 Add `idName` and `limit` as passable props to `useChunkedCQLFetch`

ryandberger:
- Fulfills [STCOR-821](https://folio-org.atlassian.net/browse/STCOR-821)
- Allow for parameter idName to be overridden to allow for using other IDs when id alone is too ambiguous (such as a case where userId, roleId, etc is needed).
- Allow for parameter limit to be overridden if a different response limit is needed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,7 +21,7 @@
 * Make `<Settings>` a functional component. Update connected modules when `stripes` object changes. Fixes STCOR-797.
 * `useOkapiKy` uses || instead of ?? to apply current tenant id when override was not provided. Refs STCOR-814.
 * Correctly parse `.../_self` permissions object. Refs STCOR-813.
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -21,6 +21,7 @@
 * Make `<Settings>` a functional component. Update connected modules when `stripes` object changes. Fixes STCOR-797.
 * `useOkapiKy` uses || instead of ?? to apply current tenant id when override was not provided. Refs STCOR-814.
 * Correctly parse `.../_self` permissions object. Refs STCOR-813.
+* Add `idName` and `limit` as passable parameters to `useChunkedCQLFetch`. Refs STCOR-821.
```
zburke, line 24: React idiom is that `idName` and `limit` are "props"

zburke (approved):
This is a nice improvement. Thank you!

PR #1442, opened 2024-03-13
STCOR-789: Restore original URL on login

aidynoJ:
Use session to restore unauthorized url pathname. 
Refs [STCOR-789](https://folio-org.atlassian.net/browse/STCOR-789).


zburke (approved):
This looks great! Consider add/remove wrapper functions for the session-storage to provide more safety than we get from the native string-as-key API.

PR #1457, opened 2024-04-18
STCOR-832 invalidate react-query cache on login and logout

JohnC-80:
https://folio-org.atlassian.net/browse/STCOR-832

When a ui-module is using `react-query` it's possible that failing requests due to expired token can be cached and not be called when the user logs back in as a result. This PR invalidates the cache, so the calls that failed previously are made again.

### Approach

Pass QueryClient down appropriately from its creation through components where it can be passed through to the utility functions found within `loginServices` for logout via RTR error or via manual logout. `login` invalidation occurs within `MainNavigation` in the component's `cDM`. The locations of this logic are probably the most debatable aspect of this PR.

### DIFF
--- a/src/components/MainNav/MainNav.js
+++ b/src/components/MainNav/MainNav.js
@@ -96,7 +96,7 @@ class MainNav extends Component {
       }
     });
 
-    // invalidate QueryProvider cache to be 100% sure we're starting from a clean slate.
+    // remove QueryProvider cache to be 100% sure we're starting from a clean slate.
     this.props.queryClient.removeQueries();
   }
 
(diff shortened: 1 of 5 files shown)


src/components/MainNav/MainNav.js
```diff
@@ -95,6 +95,9 @@ class MainNav extends Component {
         }
       }
     });
+
+    // remove QueryProvider cache to be 100% sure we're starting from a clean slate.
+    this.props.queryClient.removeQueries();
```
zburke, line 100: Why are we doing this in `<MainNav>`? That feels ... random. I mean, I get that it mounts after authentication, but still it's a weird place for this. Would it be better to hook into the other post-login functions in loginServices, or stash this (and other stuff?) in a dedicated `<AuthenticatedSessionInit>` component? Dunno. Just thinkin'.

zburke:
@vashjs, please explain why you replaced `invalidateQueries()` with `removeQueries()`. I am not opposed to the change, but I would like to understand why we might prefer one or the other.

zburke:
Replaced by PR #1504.

PR #1461, opened 2024-04-29
STCOR-841 enable StrictMode by default

zburke:
Enable [`<StrictMode>`](https://react.dev/reference/react/StrictMode) by default, allowing it to be turned off via the `stripes.config.js` prop `config.ignoreStrictMode: true`.

Refs [STCOR-841](https://folio-org.atlassian.net/browse/STCOR-841)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 * Remove tag-based selectors from Login, ResetPassword, Forgot UserName/Password form CSS. Refs STCOR-712.
 * Provide `useUserTenantPermissions` hook. Refs STCOR-830.
 * Load DayJS locale data as part of `loginServices`. STCOR-771.
+* Turn on `<StrictMode>`; ignore it with `stripes.config.js` `ignoreStrictMode: true`. Refs STCOR-841.
 
 ## 10.1.0 IN PROGRESS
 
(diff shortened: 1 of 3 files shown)


MikeTaylor (commented):
I  guess you'll fix the various lint errors, so ping me again when that's done and I'll re-review.

In the mean time, `ignoreStrictMode` seems wrong. Turning this on doesn't mean that strict mode is _ignored_, it means it's not used at all. So something like `noStrictMode` would be more honest.

How do we arrange for this happen only in dev, not in production?

zburke:
1. Sorry about the lint errors, @MikeTaylor. I bashed that out in vi instead of my IDE where those errors are more front and center. Still, such an embarrassment for a founding member of the self-appointed and surprisingly officious lint police. Anyway, fixed.
2. Renamed the setting to `disableStrictMode`. Thanks for suggesting something little more, uh, accurate. 
3. `StrictMode` does not impact production builds; it is a dev-only thing.

ncovercash (approved):

zburke:
About those test failures: they are due to newly uncovered bug in [useNamespace](https://github.com/folio-org/stripes-core/blob/ab2cdf15e2296920de633afc713b7ea61297a56c/src/components/Namespace/useNamespace.js#L34), which manipulates the return value from a hook in its render function, causing it to return different values given the same props if it renders multiple times. You can plainly see this if you log the value of namespace. Before this PR, you'd see
```
LOG: '@folio/plugin-b'
```
but now you see
```
LOG: '@folio/plugin-b'
LOG: '@folio/ui-module-b'
LOG: '@folio/plugin-b'
LOG: '@folio/ui-module-b'
LOG: '@folio/plugin-b'
LOG: '@folio/ui-module-b'
```

Good boy, StrictMode, good boy 🐶 . Here's a biscuit 🍪 

I updated `useNamespace()` to access that value using an array index, rather than getting at it via `.pop()`. @mkuklis, does this look good to you?

mkuklis:
@zburke yeah that looks great! Thank you for fixing it.

PR #1462, opened 2024-05-06
STCOR-845 Redirect correctly after changing password

ryandberger:
- Fixes [STCOR-845](https://folio-org.atlassian.net/browse/STCOR-845).
- Eureka back-end doesn't recognize `/login` route, so base URL `/` should be used instead.


zburke (approved):
The code looks correct to me but :

1. The shape of `stripes` in the code and in the prop-types must align (just make stripes an object; proptypes go byebye in React v19).
2. Please describe the solution (which new code implements) not the problem (which old code, now removed) caused. If you're ever starting a commit title with "Fix...", this is a clue that you've got the wrong end of the stick. Maybe sth like "Redirect correctly after changing password"?

aidynoJ (approved):

PR #1463, opened 2024-05-07
STCOR-776 show "Keep working?" prompt then terminate sessions due to inactivity

zburke:
## TLDR 

Track activity (e.g. clicks, keypresses, etc), and after a period of inactivity show a "Keep working?" modal with a countdown timer; when the countdown reaches zero, end the session. 

## Feature details

* Separate the RTR cycle from regular API requests. Previously, we would inspect the AT on each request to make sure it was valid, and fire RTR if the AT was getting old. This meant we had to inspect requests (to see if the AT was valid) and responses (to see if a 4xx failure was due to RTR or the API itself). Since RTR and regular API requests are now separate, we can assume the AT in an API request is valid and likewise that any error response is related to that API. Much simpler.
* Show a "Keep working?" modal after a period of inactivity (default: 1 hour) with a countdown (default: 1 minute), and terminate the session if the countdown reaches 0 without the user clicking the modal's CTA. "Activity" defaults to `keydown` and `mousedown` events. Previously, an abandoned session with an expired RT showed no indication that the RT had expired until the user performed an API request, which would cause an immediate logout. 
* Provide a `/logout` route, making it possible to logout by directly accessing that URL.
* When a session is terminated due to inactivity, redirect to `/logout-timeout`, a static route with a message explaining what happened. 

## Implementation details

Activity across tabs/windows happens via [BroadcastChannel](https://developer.mozilla.org/en-US/docs/Web/API/Broadcast_Channel_API) and storage requests, both of which emit events _only_ to the channels where they were not fired, i.e. you won't receive BroadcastChannel event on the object where it was posted, nor a storage event in the window/tab where the value changed. 
* Confirming the "Keep working?" modal in any window keeps all windows alive.  
* Activity in any window keeps all windows alive.
* Logging out in any window logs out in all windows.
* The session timeout, keep-working timeout, and the events that constitute "activity" have default values but may be overridden via `stripes.config.js`:
```
config: {
  //...
  useSecureTokens: true,
  rtr: {
    // how long before an idle session is killed? default: 60m.
    // this value must be shorter than the RT's TTL.
    // must be a string parseable by ms, e.g. 60s, 10m, 1h
    idleSessionTTL: '10m',

    // how long to show the "warning, session is idle" modal? default: 1m.
    // this value must be shorter than the idleSessionTTL.
    // must be a string parseable by ms, e.g. 60s, 10m, 1h
    idleModalTTL: '30s',

    // which events constitute "activity" that prolongs a session?
    // default: keydown, mousedown
    activityEvents: ['keydown', 'mousedown', 'wheel', 'touchstart', 'scroll'],
  }   
}
```
* Turn on the logging channels `rtr` and `rtrv` (verbose) 

Refs [STCOR-776](https://folio-org.atlassian.net/browse/STCOR-776). Replaces PR #1431, which implemented something similar but with RTR still attached to regular API requests. It ... sort of worked, but not really, and was reverted in PR #1433.


src/components/Logout/Logout.test.js
```diff
@@ -0,0 +1,51 @@
...
+  getLocale: () => Promise.resolve(),
+  logout: jest.fn()
+}));
+
+jest.mock('react-router', () => ({
+  ...(jest.requireActual('react-router')),
+  Redirect: () => <div>Redirect</div>
+}));
+
+// import { getLocale, logout } from '../../loginServices';
```
ryandberger, line 20: Pls remove commented-out code.

src/components/Root/FFetch.js
```diff
@@ -60,157 +70,65 @@ const OKAPI_FETCH_OPTIONS = {
...
-        return Promise.resolve(new Response(JSON.stringify({})));
-      });
-  };
-
   /**
    * passThrough
    * Inspect resource to determine whether it's a FOLIO API request.
+   * * If it is an authentication-related request, complete the request
+   *   and then execute the RTR callback to initiate that cycle.
+   * * If it is a logout request, complete the request b
```
ryandberger, line 130: Missing the rest of the comment???

src/components/Root/FXHR.test.js
```diff
@@ -70,7 +70,7 @@ describe('FXHR', () => {
     expect(rtr.mock.calls).toHaveLength(0);
   });
 
-  it('If AT is invalid, but RT is valid, refresh the token before sending...', async () => {
+  it.skip('If AT is invalid, but RT is valid, refresh the token before sending...', async () => {
```
ryandberger, line 73: Are these still valid tests that need to be fixed?

src/components/Root/FFetch.js
```diff
@@ -60,157 +70,65 @@ const OKAPI_FETCH_OPTIONS = {
...
+   * save a reference to fetch, and then reassign the global :scream:
    */
-  isPermissibleRequest = (resource, te, oUrl) => {
-    if (isValidAT(te, this.logger)) {
-      return true;
-    }
-
-    const isPermissibleResource = (string) => {
-      const permissible = [
-        '/bl-users/forgotten/password',
```
ryandberger, line 98: Now that this whitelist is removed, are we still able to make requests when not authenticated to perform actions like reset password?

zburke: Yes. Previously, we needed the allow-list because we inspected every non-allow-listed request to see if its AT was valid. Now, since RTR requests take place out of band from regular request, we no longer do that inspection, so we no longer need the allow-list.

src/components/Root/FXHR.test.js
zburke: Nah, they can just be removed. They relate to inspecting the AT's TTL when making a regular API request and that no longer happens.

ryandberger (approved):
Overall approve. There's a lot going on here, but it is well-commented and I appreciate the high level of test coverage. I give this my blessing to put this in master and move forward with this workflow. Nice work!

JohnC-80 (approved):

craigmcnally:
@zburke is this "always on"?  Or is it essentially opt-in by setting the configuration properties above?

package.json
```diff
@@ -75,11 +75,13 @@
     "graphql": "^16.0.0",
     "history": "^4.6.3",
     "hoist-non-react-statics": "^3.3.0",
+    "inactivity-timer": "^1.0.0",
```
craigmcnally, line 78: uses the MIT license but I think that's compatible with Apache v2.  Also the last update was 4 years ago... any concerns about either of these points?

package.json
```diff
@@ -75,11 +75,13 @@
...
     "jwt-decode": "^3.1.2",
     "ky": "^0.23.0",
     "localforage": "^1.5.6",
     "lodash": "^4.17.21",
     "moment-timezone": "^0.5.14",
+    "ms": "^2.1.3",
```
craigmcnally, line 84: Also uses the MIT license.  Much more recent updates though

craigmcnally:
> @zburke is this "always on"? Or is it essentially opt-in by setting the configuration properties above?

Looks like it's always on, which means we probably should spread the word that this is coming in the community.  At the very least we should probably add a note to the Ramsons release notes wiki page.

zburke:
@craigmcnally, TLDR: no, it's not behind a feature flag. 

I think of this as a bug fix for “I walked away for N minutes, and came back, and as soon as I clicked something I got logged out. WTH?" where N is greater than the TTL for the RT ([mod-authtoken default: one week](https://github.com/folio-org/mod-authtoken/blob/d86bceaaa203d55c11b50ae2e757bf4e678037d0/README.md#command-line-optionssystem-properties)). So, folks won't have noticed this behavior in the reference environments (since the snapshots are rebuilt daily), and this bug has probably slipped through the cracks. I think folks will start to notice it in production environments for any tenant who cranks down their RT TTL, e.g. to a day or an hour or 10 minutes. 

Without much effort, we could put it behind a feature flag, but I would question the value. If we match the idle-timeout TTL with the RT TTL, the behavior when that TTL expires will be functionally the same: you'll get logged out. The only difference is that after this PR you'll get a "sorry, your session timed out" splash page.

PR #1477, opened 2024-05-23
STCOR-853 do not include credential in /authn/token request

zburke:
The request to `/authn/token` pulls an OTP from the query string and exchanges it for AT/RT cookies. If, somehow, the browser already has cookies and sends them along on this request, it causes a negative feedback look because the OTP and the cookies are out of sync. The old AT/RT cookies will cause the endpoint to return 4xx, which will result in a redirect back to keycloak, which will find its (still perfectly valid) authentication cookies, which will cause it redirect back to stripes with a new OTP ... and the cycle repeats.

Thus, when we are exchanging an OTP, we don't want to send any cookies. We want the to send the OTP and have new cookies from the response overwrite anything that was previously stored.

Refs [STCOR-853](https://folio-org.atlassian.net/browse/STCOR-853)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 * Idle-session timeout and "Keep working?" modal. Refs STCOR-776.
 * Use keycloak URLs in place of users-bl for tenant-switch. Refs US1153537.
 * Fix 404 error page when logging in after changing password in Eureka. Refs STCOR-845.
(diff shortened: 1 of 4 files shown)


ryandberger (approved):
Glad you could correct this with a small change. Thanks for adding test coverage!

zburke:
Like #1478, branches got wacky due to an accidental merge; there should only be one commit on this branch :/ Gonna try re-opening it against current tip of keycloak-ramsons.

PR #1478, opened 2024-05-23
rebase-cleanup: restore logout AND ITS TESTS

zburke:
The previous commit re-enabled logout by correctly passing the `x-okapi-tenant` header in the `/authn/logout` request. It turns out that if you want read the tenant from the store in a test, you have to mock the store in your test. WHO KNEW???


zburke:
branches got wacky due to an accidental merge; there should only be one commit on this branch :/ Gonna try re-opening it against current tip of keycloak-ramsons.

PR #1487, opened 2024-05-31
[STCOR-787] Always retrieve clientId and tenant values from config.tenantOptions in stripes.config.js

ryandberger:
- Fulfills [STCOR-787](https://folio-org.atlassian.net/browse/STCOR-787).
- Removed dependence on `okapi.tenant` and `okapi.clientId` since those same values are now defined in `config.tenantOptions`.
- Removed dependence on `config.isSingleTenant` since a single tenant in `config.tenantOptions` should always caused the tenant selection screen to be skipped.
- Unrelated: add a fallback in `loginServices.js`if user object not returned from local storage, then default to user from `/_self` response. I ran into this issue causing login to crash occasionally on localhost and it was annoying, so adding a graceful fallback here.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 * Idle-session timeout and "Keep working?" modal. Refs STCOR-776.
 * Use keycloak URLs in place of users-bl for tenant-switch. Refs US1153537.
 * Fix 404 error page when logging in after changing password in Eureka. Refs STCOR-845.
+* Omit credentials in requests to `/authn/token` to prevent redirect tailspin. Refs STCOR-853.
 
 ## [10.1.1](https://github.com/folio-org/stripes-core/tree/v10.1.1) (2024-03-25)
 [Full Changelog](https://github.com/folio-org/stripes-core/compare/v10.1.0...v10.1.1)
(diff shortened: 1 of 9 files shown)


zburke (commented):
Are we confident about support for legacy `stripes.config.js` configurations that only `okapi.tenant`? Do we consider this a breaking change? i.e. even if apps consumeing stripes don't have to change because we continue to populate `stripes.okapi.tenant`, do hosting providers using `stripes.config.js` have to change? Do we want that?

ryandberger:
> Are we confident about support for legacy `stripes.config.js` configurations that only `okapi.tenant`? Do we consider this a breaking change? i.e. even if apps consumeing stripes don't have to change because we continue to populate `stripes.okapi.tenant`, do hosting providers using `stripes.config.js` have to change? Do we want that?



> Are we confident about support for legacy `stripes.config.js` configurations that only `okapi.tenant`? Do we consider this a breaking change? i.e. even if apps consumeing stripes don't have to change because we continue to populate `stripes.okapi.tenant`, do hosting providers using `stripes.config.js` have to change? Do we want that?

Good point. The story description suggested we drop support for okapi.tenant, but instead I should rework this to be backwards-compatible.

zburke:
Sorry; my bad for failing to write a clearer story.

ryandberger:
> Sorry; my bad for failing to write a clearer story.

This has been updated to allow for backwards-compatibility with the `okapi.tenant` and `okapi.clientId` fields the community uses.

zburke (approved):

PR #1491, opened 2024-06-13
STCOR-834: refactor useUserTenantPermissions to use _self endpoint permissions instead of okapi permissions if roles interface is presented

aidynoJ:
Refs [STCOR-834](https://folio-org.atlassian.net/browse/STCOR-834). 
Create separate hooks to retrieve permissions, one for getting okapi permissions, and second getting permissions from users-keycloak/_self endpoint. And depending on if roles interface presented use okapi permissions or users-keycloak permissions.

### DIFF
--- a/src/hooks/useUserSelfTenantPermissions.js
+++ b/src/hooks/useUserSelfTenantPermissions.js
@@ -1,0 +1,51 @@
+import { useQuery } from 'react-query';
+
+import { useStripes } from '../StripesContext';
+import { useNamespace } from '../components';
+import useOkapiKy from '../useOkapiKy';
+
+const INITIAL_DATA = [];
+
+const useUserSelfTenantPermissions = (
+  { tenantId },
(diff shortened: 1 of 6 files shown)


zburke (approved):
This code looks correct and is faithful to the original `useUserTenantPermissions` hook, but I wonder if a simpler/better refactor would be to switch its endpoint between `users-keycloak/_self` and `bl-users/_self` depending on the presence of the `roles` interface. 
1. simpler
2. better consistency between the Eureka and legacy hooks
3. more correct WRT permissions, since you can always access `.../_self`, but you need special permissions to access `/perms/users/.../permissions` (IOW, the original hook we based this off was written incorrectly). 

I'm OK merging this as-is; it will fix a bunch of bugs and (3) notwithstanding, it won't change the existing behavior, even if we don't think that is the most optimal behavior.

PR #1495, opened 2024-06-18
STCOR-863 Skip link text should announce its destination rather than what it's skipping.

JohnC-80:
[STCOR-863](https://folio-org.atlassian.net/browse/STCOR-863)

The previous text: "Skip Main Navigation" vs the forward-looking message "Skip to main content" makes AT users aware of where their focus will be placed when following the link. This only involved changing the translation string.


zburke (approved):
We need to remember to check that translations have been updated with new values before the next release.

zburke (approved):

PR #1498, opened 2024-06-25
STCOR-864 correctly evaluate typeof stripes.okapi

zburke:
Stripes should render `<ModuleContainer>` either when discovery is complete or when okapi isn't present at all, i.e. when `stripes.config.js` doesn't even contain an `okapi` entry. What's most amazing about this bug is not the bug, which is a relatively simple typo, but that it didn't bite us for more than six years.

Ignore the "new" AuthnLogin test file; those tests were previously stashed in `RootWithIntl.test.js` for some reason and have just been relocated.

See also #153, where this bug was introduced, and #1497, where it is fixed on the master branch (yes, we would eventually pick that up via rebasing, but I'm impatient).

Refs [STCOR-864](https://folio-org.atlassian.net/browse/STCOR-864)

### DIFF
--- a/src/RootWithIntl.js
+++ b/src/RootWithIntl.js
@@ -72,7 +72,7 @@ const RootWithIntl = ({ stripes, token = '', isAuthenticated = false, disableAut
                             event={events.LOGIN}
                             stripes={connectedStripes}
                           />
-                          { (connectedStripes.okapi !== 'object' || connectedStripes.discovery.isFinished) && (
+                          { (typeof connectedStripes.okapi !== 'object' || connectedStripes.discovery.isFinished) && (
                             <ModuleContainer id="content">
                               <OverlayContainer />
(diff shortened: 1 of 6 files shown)


src/RootWithIntl.js
```diff
@@ -72,7 +72,7 @@ const RootWithIntl = ({ stripes, token = '', isAuthenticated = false, disableAut
                             event={events.LOGIN}
                             stripes={connectedStripes}
                           />
-                          { (connectedStripes.okapi !== 'object' || connectedStripes.discovery.isFinished) && (
+                          { (typeof connectedStripes.okapi !== 'object' || connectedStripes.discovery.isFinished) && (
```
ryandberger, line 75: ZOMG!

ryandberger (approved):
👍 👍 👍

PR #1504, opened 2024-07-16
STCOR-832 invalidate react-query cache on login, logout

zburke:
Replaces PR #1457, which was just too cumbersome to rebase after all the work that came in [STCOR-776](https://folio-org.atlassian.net/browse/STCOR-776) / PR #1463.

Invalidate all react-query caches on login and logout. Because the RQ cache is persistent across sessions, it was possible that empty values were cached if a query returns 4xx when a session ended, and then (incorrectly) reused when the session restarted.

## Approach

Pass `<QueryClient>` down from `<RootWithIntl>` to some of its children so they can clear the cache immediately after login and before logout. The exact locations of this logic in `<MainNavigation>` and `<SessionEventContainer>` are certainly debatable, but using existing components instead of embarking on a larger refactor does have merits.

Refs [STCOR-832](https://folio-org.atlassian.net/browse/STCOR-832)


ryandberger (approved):
🧹 😄

JohnC-80 (approved):

PR #1508, opened 2024-07-25
STCOR-869 do not store /logout as a "return-to" URL

zburke:
When a session ends due to timeout, the current location is stored in order to allow the subsequent session to begin where the previous one left off. If the "session timeout" event fires more than once**, however, this could lead to the `/logout` location being stored as the "return to" location with obvious dire consequences.

There are two changes here:
1. Don't allow locations beginning with `/logout` to be stored. This fixes the symptom, not the root cause, but is still worthwhile.
2. Store the session-timeout interval ID in redux, and manage that timer via a redux action. Even though this _still_ shouldn't fire more than once, if it does, this allows us to cancel the previous timer before adding the next one. This is an attempt to fix the root cause.

Refs [STCOR-869](https://folio-org.atlassian.net/browse/STCOR-869)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,5 @@
 # Change history for stripes-core
 
->>>>>>> f93f21d6 (STCOR-866 include `/users-keycloak/_self` in auth-n requests (#1502))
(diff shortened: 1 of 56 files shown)


ryandberger (approved):
LGTM

zburke:
Replaced by #1511, an identical rebase.

PR #1513, opened 2024-08-07
[STCOR-869] Add small margin to ensure /logout is called before cookie expires

ryandberger:
- Added a small time margin to call /logout request before cookie is deleted so it is included in the request.
- Fixes [STCOR-869](https://folio-org.atlassian.net/browse/STCOR-869)


aidynoJ (approved):

zburke:
@ryandberger , this is a simple fix, which is nice, but the implementation is inconsistent with existing code and makes room for a user-facing glitch. On the developer side:
* the value is an int in ms while others use a string interpreted by `ms()`.
* similar functionality for ATs is handled by RTR_AT_TTL_FRACTION using a percentage of the available TTL instead of a fixed value. 

On the user side: 
* the warning TTL and the actual TTL are no longer aligned, i.e. we clipped 200ms off the FLST but did not slide the warning back by 200ms. Maybe 200ms is not noticeable now, but let's say this missing-cookie problem continues to plague us for users with slow connections or in locales where there is greater lag between the browser and server and we decide to bump the value to 2s. Now, sessions will end when there is still 00:02 on the clock, and that will be very noticeable. 

It would be nice to fix all these issues, but the last is the most important since it is user-facing while the others are only developer-facing. If we adjust the FLST TTL, we need to incorporate that same adjustment into when the end-of-session warning is displayed as well, sliding it earlier instead of truncating it.

ryandberger:
> @ryandberger , this is a simple fix, which is nice, but the implementation is inconsistent with existing code and makes room for a user-facing glitch. On the developer side:
> 
> * the value is an int in ms while others use a string interpreted by `ms()`.
> * similar functionality for ATs is handled by RTR_AT_TTL_FRACTION using a percentage of the available TTL instead of a fixed value.
> 
> On the user side:
> 
> * the warning TTL and the actual TTL are no longer aligned, i.e. we clipped 200ms off the FLST but did not slide the warning back by 200ms. Maybe 200ms is not noticeable now, but let's say this missing-cookie problem continues to plague us for users with slow connections or in locales where there is greater lag between the browser and server and we decide to bump the value to 2s. Now, sessions will end when there is still 00:02 on the clock, and that will be very noticeable.
> 
> It would be nice to fix all these issues, but the last is the most important since it is user-facing while the others are only developer-facing. If we adjust the FLST TTL, we need to incorporate that same adjustment into when the end-of-session warning is displayed as well, sliding it earlier instead of truncating it.

These are very good points. I will create a follow-up PR to hopefully clear up the dev issues and absolutely correct the last issue.

PR #1517, opened 2024-08-14
[STCOR-869] Add small margin to fixed timeout warning to match call to /logout

ryandberger:
- Follow-up on work previously done for [STCOR-869](https://folio-org.atlassian.net/browse/STCOR-869).
- Addressing comments: https://github.com/folio-org/stripes-core/pull/1513#issuecomment-2284382894
  - Primarily that warning TTL an actual TTL were no longer aligned. By subtracting the margin from warning TTL, they are now aligned.
  - Also, switched`RTR_TIME_MARGIN` to be string to match other constants.
- Extracted `timestampFormatter()` to utils file so that it can be accessed in unit tests.


zburke (commented):
@ryandberger, This works, but it splits the "when does my session end?" math across two different places, FFetch and FixedLengthSessionWarning. Could we consolidate all of into FFetch around lines 118-119? That would allow us to leave FixedLengthSessionWarning alone, and it would simplify the `setTimeout()` calls a little since they'd each just rely on their single `rt...Interval` values. 

What do you think?

zburke:
RTR has changed significantly since this PR was introduced, most notably in #1696 with the abandoning of the timer-based rotation that made this work necessary.

PR #1519, opened 2024-08-15
STCOR-873 Ensure support for the passed `tenantId` value by `useChunkedCQLFetch` for manipulations in the context of a specific tenant

usavkov-epam:
## Purpose
https://folio-org.atlassian.net/browse/STCOR-873

useChunkedCQLFetch uses the system tenant under the hood to send requests, however, some places in the application (specifically the consortium manager) require sending requests to a tenant other than the system one. Therefore, useChunkedCQLFetch should be updated to support an additional tenantId option, which will be used by the HTTP client. In the absence of this property, the system tenant will be used by default.

**Note:** PR for `master` branch - https://github.com/folio-org/stripes-core/pull/1518

### DIFF
--- a/src/queries/useChunkedCQLFetch.test.js
+++ b/src/queries/useChunkedCQLFetch.test.js
@@ -202,7 +202,8 @@ describe('Given useChunkedCQLFetch', () => {
     it('expose queryKeys based on given ids and endpoint', async () => {
       const { result } = renderHook(() => useChunkedCQLFetch({
         ...baseOptions,
-        STEP_SIZE: 2
+        STEP_SIZE: 2,
+        tenantId: 'central',
       }), { wrapper });
 
       await waitFor(() => {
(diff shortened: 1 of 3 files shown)


src/queries/useChunkedCQLFetch.test.js
```diff
@@ -194,4 +199,37 @@ describe('Given useChunkedCQLFetch', () => {
       expect(result.current.itemQueries?.length).toEqual(2);
     });
   });
+
+  describe('allows work in provided tenant', () => {
+    const providedTenantId = 'providedTenantId';
+
+    it('sets up 1 fetch using alternate ID name', async () => {
```
zburke, line 206: More copy paste problems: you're not changing the ID name here, you're changing the tenantId. How about "sets up 1 fetch using alternate tenant ID"?

zburke (approved):
LGTM once the comment and test description are corrected to reflect the new `tenantId` argument instead of the copy-pasted code.

artem-blazhko (approved):

zburke:
@usavkov-epam , it is also worth noting that we have generally managed this branch by regularly rebasing it onto master, which would make this PR unnecessary. If we really really need this ASAP, it's fine to have open a PR as you have done here, but given this work isn't specific to Eureka, the preferred option would be to just pick up this change from master the next time we rebase.

usavkov-epam:
> @usavkov-epam , it is also worth noting that we have generally managed this branch by regularly rebasing it onto master, which would make this PR unnecessary. If we really really need this ASAP, it's fine to have open a PR as you have done here, but given this work isn't specific to Eureka, the preferred option would be to just pick up this change from master the next time we rebase.

Thank you, Zak, for noticing! I wasn't sure how often this branch is updated for changes from `master`, so I decided to create two PRs because these changes are indeed needed for the current feature implementation in the "Keycloak" environment.

PR #1520, opened 2024-08-15
STCOR-872: Return query keys from useChunkedCQLFetch 

aidynoJ:
Sometimes we need to invalidate queries, to be able to do so queryKeys generated in useChunkedCQLFetch should be available. 
**Purpose**
[STCOR-872](https://folio-org.atlassian.net/browse/STCOR-872) - expose query keys from useChunkedCQLFetch.

### DIFF
--- a/src/queries/useChunkedCQLFetch.js
+++ b/src/queries/useChunkedCQLFetch.js
@@ -96,7 +96,9 @@ const useChunkedCQLFetch = ({
     STEP_SIZE
   ]);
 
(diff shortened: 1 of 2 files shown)


zburke:
This enables folio-org/stripes-authorization-components/pull/23 (UISAUTHCOM-13), fixing the problem described in the comments of folio-org/stripes-authorization-components/pull/18 (UISAUTHCOM-10).

PR #1526, opened 2024-08-23
[STCOR-876] Remember requested URL path on Login (Regression bug) 

ryandberger:
- Refs [STCOR-876](https://folio-org.atlassian.net/browse/STCOR-876).
- Ensure `setUnauthorizedPathToSession` is called in `AuthnLogin` to remember requested path on login. I added a more explicit default path handler when user is unauthorized. I also switched from `useEffect` to `useLayoutEffect` in `AuthnLogin` so that logic is called before anything on the UI is drawn.
- Also adding `window.location.search` to session storage so URL params are remembered.

### DIFF
--- a/src/RootWithIntl.js
+++ b/src/RootWithIntl.js
@@ -163,6 +163,7 @@ const RootWithIntl = ({ stripes, token = '', isAuthenticated = false, disableAut
                       />
                       <TitledRoute
                         name="login"
+                        path="*"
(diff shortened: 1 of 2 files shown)


aidynoJ:
LGTM. I'm okay with using useLayoutEffect to store items in session storage, as the component returns a redirect on single tenant.

aidynoJ (approved):

PR #1531, opened 2024-09-06
[STCOR-885] Clear saved entry path so that subsequent logins will use default base URL.

ryandberger:
- Fixes [STCOR-885](https://folio-org.atlassian.net/browse/STCOR-885). 
- The issue was that if a user initially navigated to a routed URL such as `/users?search=test` after redirect on login, that route is opened on all consecutive logins. By clearing session storage on logout, subsequent logins go to the default `/` route as expected.
- Cleaned up some redundant code. `setUnauthorizedPathToSession()` was being called with same params as it defaults to. Also cleaned up duplicate code within `setUnauthorizedPathToSession()`

### DIFF
--- a/src/loginServices.js
+++ b/src/loginServices.js
@@ -514,9 +514,6 @@ export async function logout(okapiUrl, store) {
       localStorage.removeItem(SESSION_NAME);
       localStorage.removeItem(RTR_TIMEOUT_EVENT);
 
-      // Clear saved entry path so that subsequent logins will use default base URL.
-      removeUnauthorizedPathFromSession();
-
       store.dispatch(setIsAuthenticated(false));
       store.dispatch(clearCurrentUser());
       store.dispatch(clearOkapiToken());
--- a/src/loginServices.js
+++ b/src/loginServices.js
@@ -125,7 +125,7 @@ export const removeUnauthorizedPathFromSession = () => sessionStorage.removeItem
 export const setUnauthorizedPathToSession = (pathname) => {
   const path = pathname ?? `${window.location.pathname}${window.location.search}`;
   if (!path.startsWith('/logout')) {
-    sessionStorage.setItem(UNAUTHORIZED_PATH, pathname ?? `${window.location.pathname}${window.location.search}`);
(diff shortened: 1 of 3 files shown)


src/loginServices.js
```diff
@@ -514,6 +514,9 @@ export async function logout(okapiUrl, store) {
       localStorage.removeItem(SESSION_NAME);
       localStorage.removeItem(RTR_TIMEOUT_EVENT);
 
+      // Clear saved entry path so that subsequent logins will use default base URL.
+      removeUnauthorizedPathFromSession();
+
```
zburke, line 519: This doesn't look correct. We preserve the last-active URL for three reasons: 
1. when active sessions end due to [IST](https://github.com/folio-org/stripes-core/blob/3aef1a056c2932b94e2ed179e9eb9fc8c6922bef/src/components/SessionEventContainer/SessionEventContainer.js#L34)
2. when active sessions end due to [FLST](https://github.com/folio-org/stripes-core/blob/3aef1a056c2932b94e2ed179e9eb9fc8c6922bef/src/components/SessionEventContainer/SessionEventContainer.js#L47)
3. when inactive sessions are started from a bookmark in AuthnLogin, as you noticed here

This change will break 1. and 2. 

Instead of clearing storage on the way out during logout, I think we need to clear it on the way in, right after it's read in [OIDCRedirect](https://github.com/folio-org/stripes-core/blob/3aef1a056c2932b94e2ed179e9eb9fc8c6922bef/src/components/OIDCRedirect.js).

zburke (changes requested):
Technically, this resolves the problem described in STCOR-885, but it also reintroduces those like [STCOR-849](https://folio-org.atlassian.net/browse/STCOR-849). We need to resolve both.

src/loginServices.js
ryandberger: Makes sense, I'll make the correction. Thanks for pointing out the other use cases.

zburke (approved):
LGTM! Please update the PR description/commit commit to reflect the actual changes that we settled on in this final version.

PR #1535, opened 2024-09-18
[STCOR-888] RTR dynamic configuration, debugging event

ncovercash:
# [Jira STCOR-888](https://folio-org.atlassian.net/browse/STCOR-888)

This PR adds:

- A new event `RTR_FORCE_REFRESH_EVENT` that forces a RTR refresh
  - This is not dispatched by stripes-core, but will aid development/debugging
- Dynamic configuration for `RTR_AT_TTL_FRACTION` to allow customization from `stripes.config.js` ~~and/or at runtime via `ui-developer`~~


src/components/SessionEventContainer/SessionEventContainer.js
```diff
@@ -250,6 +250,7 @@ const SessionEventContainer = ({ history, queryClient }) => {
     // We only want to configure the event listeners once, not every time
     // there is a change to stripes or history. Hence, an empty dependency
     // array.
+    // should `stripes.rtr` go here?
   }, []); // eslint-disable-line react-hooks/exhaustive-deps
```
ncovercash, line 254: @zburke can you comment on this? As-is, this configuration cannot be modified at runtime by `ui-developer`.

mweaver-ebsco (approved):

src/components/Root/FFetch.js
```diff
@@ -77,10 +77,27 @@ const OKAPI_FETCH_OPTIONS = {
 };
 
 export class FFetch {
-  constructor({ logger, store, rtrConfig }) {
```
ncovercash, line 80: RTR config is exposed as part of the main `stripes-config`. Why should FFetch get passed it in a non-standard way?

JohnC-80: Strikes me as 'better testability' if you're able to just pass the config in vs having to mock it out/provide different mocks through some other way.

src/components/Root/FFetch.js
```diff
@@ -77,10 +77,27 @@ const OKAPI_FETCH_OPTIONS = {
...
+    };
+    window.addEventListener(RTR_FORCE_REFRESH_EVENT, this.globalEventCallback);
+  }
+
+  /**
+   * unregister the listener for the RTR_FORCE_REFRESH_EVENT
+   */
+  unregisterEventListener = () => {
+    window.removeEventListener(RTR_FORCE_REFRESH_EVENT, this.globalEventCallback);
   }
```
zburke, line 101: All our other event handlers are consolidated within SessionEventContainer. Can we move this there too? If not why not? 

Functionally it should not matter either way, but it would be nice to keep similar things in the same place.

src/components/Root/Root.js
```diff
@@ -67,13 +67,15 @@ class Root extends Component {
     // * configure fetch and xhr interceptors to conduct RTR
     // * see SessionEventContainer for RTR handling
     if (this.props.config.useSecureTokens) {
-      const rtrConfig = configureRtr(this.props.config.rtr);
+      // FFetch relies on some of these properties, so we must ensure
+      // they are filled before initialization
+      this.props.config.rtr = configureRtr(this.props.config.rtr);
```
zburke, line 72: Reassigning something on props feels icky. Why was it necessary to change this line?

src/components/SessionEventContainer/SessionEventContainer.js
zburke: Probably I started with a non-empty array and it didn't work as expected, so I added the empty array as a crutch and never looked back. Your observation/criticism is fair, but I don't have time to debug this further since the only thing "broken" about it is that it can't be modified at runtime via dev-tools.

zburke (approved):
Sorry for the delay; I got started on this and then lost track of it. 

At a high-level the code changes look good. I would prefer to consolidate the event handlers into SessionEventContainer, but that's just ergonomics. If you merge without doing this, please explain why you opted not to.

src/components/Root/Root.js
ncovercash: It's that, or reassigning the import (`configureRtr` doesn't change the input). This was changed because my FFetch imported config via `stripes-config`, whereas before it was passed in.

I can:
- revert it to being passed in instead, if preferred, but IMO it's best for the shared config to match the filled-in version from `configureRtr` (and we already do that in render(), where we have `gross: this overwrites whatever is currently stored at config.rtr`)
- make `configureRtr` mutate its input

src/components/Root/FFetch.js
ncovercash: Unfortunately, I don't think this can be consolidated; this handler specifically needs access to those callbacks and fields within `FFetch`, which `SessionEventContainer` has no way to access

src/components/Root/Root.js
zburke: There are two things happening here with the RTR config: 
1. passing it to FFetch (in the constructor)
2. passing it Stripes (in render)

As you noted, we can get around (1) by updating FFetch to leverage stripes-config instead, and call `configureRtr()` in the constructor there. Should we just do the same in `render()` here? Part of me screams, "You should only have to call that function once! More than once means you're doing it wrong!" The other part of me is like, "Eh, if the value in stripes-config may be sparse, think of `configureRtr()` as `getSafeRtrValues()`. Does that still bug ya? No? Great."

Better, worse, just different but not actually better?

ncovercash: Importing stripes-config for `render` here might make sense, but I'm not familiar enough with the rest of this to know if that's a bad idea.

What we can (and should) do is remove the `configureRtr` call inside `render()` here. It just populates the sparse values (as you said), and there's no reason any values should be removed between renders.

src/components/SessionEventContainer/SessionEventContainer.js
ncovercash: Tested it, and it works beautifully. Only the RTR config in ui-developer causes the effect to re-run, nothing else (not even the other config forms in developer)

ncovercash: It does rely on not calling `configureRtr` on render in `Root`, though.

julianladisch (approved):

PR #1541, opened 2024-10-02
STCOR-893: Misleading "server is down"

julianladisch:
https://folio-org.atlassian.net/browse/STCOR-893

Reword misleading error message.


src/components/Root/Root.js
```diff
@@ -120,7 +120,7 @@ class Root extends Component {
     const { logger, store, epics, config, okapi, actionNames, token, isAuthenticated, disableAuth, currentUser, currentPerms, locale, defaultTranslations, timezone, currency, plugins, bindings, discovery, translations, history, serverDown } = this.props;
     if (serverDown) {
       // note: this isn't i18n'ed because we haven't rendered an IntlProvider yet.
-      return <div>Error: server is down.</div>;
+      return <div>Error: server is forbidden, unreachable or down. VPN issue?</div>;
```
MikeTaylor, line 123: Shouldn't this be localized?

ryandberger: According to the comment above, looks like i18n is not possible here.

MikeTaylor: D'oh! It's right there in the comment!

OK, I am off to lie down for a while. Sorry to waste your time.

PR #1543, opened 2024-10-10
[STCOR-835] refactor getUserTenantsPermissions to leverage roles instead of permissions

ryandberger:
- Conditionally use `/users-keycloak/_self` endpoint when `users-keycloak` interface is present.
- Fulfills [STCOR-835](https://folio-org.atlassian.net/browse/STCOR-835).


ryandberger:
> I'm uncertain about this in light of the corresponding work @aidynoJ just did for [UIU-3102](https://folio-org.atlassian.net/browse/UIU-3102) in [PR 2752](https://github.com/folio-org/ui-users/pull/2752) to eliminate dependencies on the `users-keycloak` interface. I'm also confused about the directions given on this ticket. The conditional checks for the presence of `roles` but the endpoint (`/users-keycloak/_self`) comes from `users-keycloak`, so something seems amiss.
> 
> Finally, on the non-Eureka side, we should use `/bl-users/_self` instead of `/perms/users/${id}/permissions` since the former is universally accessible to any authenticated user but the latter requires specific permissions. If you don't want to do that clean up here I understand. Please file a new bug on UIU and have a chat with the ECS team who originally implemented this function since it was done incorrectly.

Seems like based on https://github.com/folio-org/ui-users/pull/2752, both Eureka and non-Eureka can call `/bl-users/_self`. Is that correct?

aidynoJ:
> > I'm uncertain about this in light of the corresponding work @aidynoJ just did for [UIU-3102](https://folio-org.atlassian.net/browse/UIU-3102) in [PR 2752](https://github.com/folio-org/ui-users/pull/2752) to eliminate dependencies on the `users-keycloak` interface. I'm also confused about the directions given on this ticket. The conditional checks for the presence of `roles` but the endpoint (`/users-keycloak/_self`) comes from `users-keycloak`, so something seems amiss.
> 
> > 
> 
> > Finally, on the non-Eureka side, we should use `/bl-users/_self` instead of `/perms/users/${id}/permissions` since the former is universally accessible to any authenticated user but the latter requires specific permissions. If you don't want to do that clean up here I understand. Please file a new bug on UIU and have a chat with the ECS team who originally implemented this function since it was done incorrectly.
> 
> 
> 
> Seems like based on https://github.com/folio-org/ui-users/pull/2752, both Eureka and non-Eureka can call `/bl-users/_self`. Is that correct?

That was the requirements for UIU-3192

aidynoJ:
I might be wrong. But based on https://folio-org.atlassian.net/browse/MODUSERSKC-30  and [Bridging the gap between Users in the system and Users of the system](https://folio-org.atlassian.net/wiki/spaces/FOLIJET/pages/109969482/Bridging+the+gap+between+Users+in+the+system+and+Users+of+the+system#Option-1---Promotion-of-users-in-the-system-to-users-of-the-system  ) we no longer need to create keycloak record directly. We can create them by assigning users to role and vice versa.

ryandberger:
> I might be wrong. But based on https://folio-org.atlassian.net/browse/MODUSERSKC-30 and [Bridging the gap between Users in the system and Users of the system](https://folio-org.atlassian.net/wiki/spaces/FOLIJET/pages/109969482/Bridging+the+gap+between+Users+in+the+system+and+Users+of+the+system#Option-1---Promotion-of-users-in-the-system-to-users-of-the-system) we no longer need to create keycloak record directly. We can create them by assigning users to role and vice versa.

In this case, the endpoint is just being used to retrieve permissions a user has with tenants within a consortium. There is no create actions taking place here.

aidynoJ:
> > I might be wrong. But based on https://folio-org.atlassian.net/browse/MODUSERSKC-30 and [Bridging the gap between Users in the system and Users of the system](https://folio-org.atlassian.net/wiki/spaces/FOLIJET/pages/109969482/Bridging+the+gap+between+Users+in+the+system+and+Users+of+the+system#Option-1---Promotion-of-users-in-the-system-to-users-of-the-system) we no longer need to create keycloak record directly. We can create them by assigning users to role and vice versa.
> 
> In this case, the endpoint is just being used to retrieve permissions a user has with tenants within a consortium. There is no create actions taking place here.

I meant that probably it is related, now when we create user, the user is not created in keycloak and calling mod-roles-keycloak _self might cause an issue

aidynoJ (approved):

src/queries/getUserTenantsPermissions.js
```diff
@@ -8,16 +8,19 @@
...
     okapi: {
       url,
       token,
     }
   } = stripes;
   const userTenantIds = tenants.map(tenant => tenant.id || tenant);
 
+  const permUrl =
+    `${url}/${stripes.hasInterface('users-keycloak') ?
+      'users-keycloak' : 'bl-users'}/_self?expandPermissions=true`;
```
zburke, line 20: Petty: This is difficult to read with the ternary split across two lines. How about 
```
const permPath = stripes.hasInterface('users-keycloak') ? 'users-keycloak' : 'bl-users';
const permUrl = `${url}/${permPath}/_self?expandPermissions=true`;
```

PR #1550, opened 2024-10-22
STCOR-900: (ECS) Automatically logged into Central tenant after logging out from Member tenant

aidynoJ:
## Purpose
[STCOR-900](https://folio-org.atlassian.net/browse/STCOR-900)  - (ECS) Automatically logged into Central tenant after logging out from Member tenant.

## Approach 
On login, we store tenant data in localStorage and use it as the initial state of the Redux store for okapi.tenant. When the affiliation changes, okapi.tenant updates, but the tenant in localStorage does not. As a result, the updated okapi.tenant is sent on logout. Since the tenant in the x-okapi-token and the x-okapi-tenant header on logout should match, use the tenant data stored in localStorage for the x-okapi-tenant header.

## Refs
https://folio-org.atlassian.net/browse/STCOR-900

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,7 @@
 * Move session timeout banner to the bottom of the page. Refs STCOR-883.
 * Conditionally use `/users-keycloak/_self` endpoint when `users-keycloak` interface is present. Refs STCOR-835.
 * Wait longer before declaring a rotation request to be stale. Refs STCOR-895.
+* Send the stored central tenant name in the header on logout. Refs STCOR-900.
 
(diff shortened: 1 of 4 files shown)


ryandberger (approved):

zburke (approved):
Code changes are fine, and simple, but even so still need test coverage. Please add this before you commit. A comment in loginService.js::logout() explaining why this call is necessary would be helpful too. 

We oughtta consider writing generic, publicly-exported local-storage and session-storage functions with namespacing etc. That would allow us to avoid writing all these tiny new specialty functions like `getFoo()` and `getBar()`, and allow us not to have to write a bunch of identical tests. It would also decrease the bikeshedding in individual applications. I'll write up an STCOR ticket for this.

PR #1551, opened 2024-10-25
STCOR-882 provide a "you have logged out" page

zburke:
Consolidate `<Logout>` and `<LogoutTimeout>` into a single component, accessible through different routes, and always redirect there. This changes the behavior of user-directed "logout" by showing a "You have logged out" page with a "login again?" button.

Previously, the sequence was a redirect to `/`, which would eventually be followed by a redirect for authn. The problem is that the router, `<RootWithIntl>`, would receive the redirect to `/` before re-rendering with `isAuthenticated: false`. This causes `<Front>` to render the "Welcome to FOLIO" page, followed by the re-render and the redirect to authn.

Refs [STCOR-882](https://folio-org.atlassian.net/browse/STCOR-882)


ryandberger (approved):
LGTM!

PR #1556, opened 2024-11-04
STCOR-905: based on 'users-keycloak' interface use bl-users or users-keycloak for _self endpoint in useUserTenantPermissions

aidynoJ:
**Purpose**
[STCOR-905](https://folio-org.atlassian.net/browse/STCOR-905) - based on 'users-keycloak' interface use bl-users or users-keycloak for _self endpoint in useUserTenantPermissions #1556

**Approach**
Remove redundant hooks and make changes in useUserTenantPermissions


ryandberger:
Is this still a draft PR?

ryandberger (approved):

ryandberger:
Please fix lint issue.

zburke (commented):
Code changes look correct to me, but changing the hook's name is a breaking change. If the target is sunflower, that's OK (we know that release will contain breaking changes), but we should do that work in two phases: 
1. in stripes-core, export this hook by its old and new names
2. in other repositories, refactor to the new name
3. in stripes-core, remove the export of the old name

That avoids the need to merge PRs here and across the UI repositories at the same moment. If we go that route, how about a shorter name, e.g. `useTenantPermissions`? Permissions don't exist outside the context of a user, so that part of the name feels redundant. 

Alternatively, just leave the name as-is. That allows consumers of this hook (inventory, marc-authorities, quick-marc, and stripes-authority-components) to stay as-is, and ui-users can name its local hook whatever it wants to, even if it overlaps; because it isn't importing from stripes-core there isn't a naming conflict, even if it's confusing. 

Also, lint :)

aidynoJ:
> Code changes look correct to me, but changing the hook's name is a breaking change. If the target is sunflower, that's OK (we know that release will contain breaking changes), but we should do that work in two phases:
> 
> 1. in stripes-core, export this hook by its old and new names
> 2. in other repositories, refactor to the new name
> 3. in stripes-core, remove the export of the old name
> 
> That avoids the need to merge PRs here and across the UI repositories at the same moment. If we go that route, how about a shorter name, e.g. `useTenantPermissions`? Permissions don't exist outside the context of a user, so that part of the name feels redundant.
> 
> Alternatively, just leave the name as-is. That allows consumers of this hook (inventory, marc-authorities, quick-marc, and stripes-authority-components) to stay as-is, and ui-users can name its local hook whatever it wants to, even if it overlaps; because it isn't importing from stripes-core there isn't a naming conflict, even if it's confusing.
> 
> Also, lint :)

Let's choose the easiest way, leave it as is and go further. Developers can decide how to call their local hooks whatever they want :)

PR #1560, opened 2024-11-14
STCOR-910 provide IfAnyPermission, stripes.ifAnyPerm

zburke:
`<IfPermission>` and `stripes.ifPerm()` return `true` when the user has ALL the given permissions. `<IfAnyPermission>` and `stripes.ifAnyPerm()` return `true` when the user has ANY of the given permission.

Refs [STCOR-910](https://folio-org.atlassian.net/browse/STCOR-910)


BogdanDenis (approved):
Thank you @zburke!

PR #1565, opened 2024-11-21
STCOR-913 do not overwrite initialState in BTOG setup

zburke:
Test setup must honor the values it accepts in the `initialState` argument. It can augment them, but must not ignore them.

Refs [STCOR-913](https://folio-org.atlassian.net/browse/STCOR-913)


zburke:
🙄 duplicates #1564

PR #1567, opened 2024-11-26
STCOR-906 consume react-query QueryClient in <Logout>

zburke:
Correctly pass all three required arguments from `<Logout>` to `loginServices::logout()`: gateway URL, redux store, react-query QueryClient. The third was missing, resulting in an NPE during `logout()` that caused flow to skip to the `finally` clause, skipping the code that cleared localforage. Sometimes localforage appears to get cleared anyway, but sometimes not, in which case the session may fail to terminate even when the fixed-length session timer reaches 00:00.

This is one of those "How did this ever work?!? And how did we not notice?!?" situations.

Refs [STCOR-906](https://folio-org.atlassian.net/browse/STCOR-906)


aidynoJ (approved):

ryandberger (approved):
Lint, but otherwise good to go!

PR #1578, opened 2024-12-23
STCOR-932 implement useModuleInfo(path) to map path -> module

zburke:
Query discovery to retrieve information about the module implementing a given endpoint. Simple usage:
```
const { module } = useModuleInfo('/path');
// { name: 'mod-foo', ... }
```

Refs [STCOR-932](https://folio-org.atlassian.net/browse/STCOR-932)


JohnC-80:
Pedantic, `useModuleFor` seems like an odd name... `useModuleInfo`, `useModuleName` or `useBEModuleName` ?

Also needs a top-level  export.

zburke:
> Pedantic, useModuleFor seems like an odd name... 

Fair, @JohnC-80. My thinking was "you give it a path, it gives you module info" but I agree the name is awkward. You get more than just the module-name as a return value, hence avoiding "name" in the name. So ... `useModuleInfo()`? `useGetModuleFromPath()`?

JohnC-80:
@zburke I like `useModuleInfo()` The API is documentable, but the name tells you what it provides - info about the module.

zburke:
@JohnC-80: I added exports and renamed the hook. Take another look?

JohnC-80 (approved):

PR #1581, opened 2025-01-06
STCOR-926 validate that cookies and storage are available

zburke:
Detect the presence of localStorage, sessionStorage, and cookies early early early in the stripes-init process and show an error message if any are unavailable.

This prevents a white screen of death if, say, session storage is unavailable but we call it anyway, resulting in an untrapped exception (Here's looking at you, OIDCRedirect).

Refs [STCOR-926](https://folio-org.atlassian.net/browse/STCOR-926)


JohnC-80 (approved):
LGTM

PR #1584, opened 2025-01-27
STCOR-936 - implement App-reordering, user preference management in stripes-core.

JohnC-80:
## [STCOR-936](https://folio-org.atlassian.net/browse/STCOR-936)

This PR Implements 2 things -

1. A centralized mechanism for managing user preferences and interacting with `mod-settings`.
2. A context that reads/writes a persisted user preference for the order of application icons in the main navigation of the UI.

1 is fairly straightforward. A hook - `usePreferences` - that accepts an object of options with the key and scope of a user preference. It will automatically apply the logged in user's id as well as maintain the id of an existing preference in its internal state. As per typical, it try/catches the requests. TODO: Better error handling than `console.log`. 

`usePreferences` supplies methods `getPreference`, `setPreference`, and `deletePreference`. 

`getPreference/setPreference` queries the preferences with the `userId`, `scope`, and `key`. If the preference is found, it will grab the id and store that in local state to use when `setPreference` is called. If a preference is not found, then the `id` remains `null` and when `setPreference` is called, a new preference is created according to `mod-settings`' API. `deletePreference` removes the preference and resets the found preference id if there is any.

2 involved setting up a context with a Provider - `AppOrderProvider`- that wraps the FOLIO UI at a high level - outside of the main navigation and around the ui-module container. It access the user preference and provides an ordered list of apps filtered by user permissions (named `apps`). It also provides the persisted preference value - an array containing objects with app information - only `name` as well as an `isNew` field if a platform app does not have a corresponding list order value.

The app list -building logic was lifted from `MainNav.js` and `<MainNav>` refactored as a functional component, trading out HOC's for hooks. One detail that's subject to further inspection is line 58 where it uses `logout` instead of an apparently undefined `returnToLogin` function.

`ResizeContainer` and `AppListDropdown` were two components that had to be touched as they were forcing alphabetical order of the list - switching from `Object.keys` to using the order of the array of apps/preference.

## Still Remaining:
- [x] Tests!


src/components/MainNav/MainNav.js
```diff
@@ -19,206 +14,107 @@ import {
...
+      const module = curModule;
+      const state = store.getState();
 
       // If user has timed out, force them to log in again.
       if (state?.okapi?.token && state.okapi.authFailure
         && find(state.okapi.authFailure, { type: 'error', code: 'user.timeout' })) {
-        this.returnToLogin();
+        // this.returnToLogin();
+        logout(state.okapi.url, store, queryClient);
       }
```
zburke, line 60: In RTR-land where tokens only exist in HTTPOnly cookies, we don't have to worry about this condition because `state.okapi.token` will never evaluate to true. IOW, I think the right clean-up here is "remove this entire conditional". 

It looks like the bad commit is `a9b860dfe591f7e7b236640fc1ceb6d25401b30c`, itself a "clean-up" commit  from mega-merge PR #1532. The comment there alludes to fixing up logout, but unfortunately doesn't offer any details about why, specifically, `returnToLogin` was removed. If anybody knows this @zburke character we should find him and tell him to write better commit messages.

package.json
```diff
@@ -38,6 +39,16 @@
       {
         "permissionName": "settings.enabled",
         "displayName": "UI: settings area is enabled"
+      },
+      {
+        "permissionName": "mod-settings.owner.read.stripes-core.prefs.manage",
+        "displayName": "UI: read the user's own preference for the order of links in the main navigation.",
```
MikeTaylor, line 45: Could we say something like "UI: read the user's own preferences, such as the order of links in the main navigation"? We are surely going to want to use this for lots of other things, too, once it's in place.

package.json
```diff
@@ -38,6 +39,16 @@
...
+        "visible": false
+      },
+      {
+        "permissionName": "mod-settings.owner.write.stripes-core.prefs.manage",
+        "displayName": "UI: update the user's own preference for the order of links in the main navigation.",
```
MikeTaylor, line 50: See previous comment.

src/components/MainNav/AppList/AppList.js
```diff
@@ -91,7 +91,7 @@ class AppList extends Component {
         {
           apps.map(app => {
             const isHidden = hiddenItemIds.includes(app.id);
-
+            // const isHidden = false;
```
MikeTaylor, line 94: What's this about?

src/components/MainNav/AppOrderProvider.js
```diff
@@ -0,0 +1,166 @@
...
+    const perm = `module.${name}.enabled`;
+
+    if (!stripes.hasPerm(perm)) {
+      return null;
+    }
+
+    const id = `clickable-${name}-module`;
+
+    const pathRoot = pathname.split('/')[1];
+    const entryRoot = entry.route.split('/')[1];
```
MikeTaylor, line 41: It looks this means each FOLIO UI module has to have a unique top-level path component — e.g. you can't have ui-foobar that uses `/foo/bar` and ui-foobaz that uses `/foo/baz`. Am I reading that right? If so, is that limitation new, or was it there in the old code?

src/components/MainNav/AppOrderProvider.js
```diff
@@ -0,0 +1,166 @@
...
+export const AppOrderProvider = ({ children }) => {
+  const { lastVisited } = useContext(LastVisitedContext);
+  const { app } = useModules();
+  const stripes = useStripes();
+  const { pathname } = useLocation();
+  const { formatMessage } = useIntl();
+  const { getPreference, setPreference, removePreference } = usePreferences();
+
+  const { data: userAppList, isLoading, refetch: refetchAppList } = useQuery('UserPrefsAppListing', () => {
+    return getPreference({ key: APPORDER_PREF_NAME, scope: APPORDER_PREF_SCOPE });
```
MikeTaylor, line 91: Maybe use an expression here rather than a block that does nothing but return the expression?

src/hooks/usePreferences.js
```diff
@@ -0,0 +1,78 @@
...
+    // this will return an array that's likely a single item long. We store the item's id in state so that we can update it later.
+    let respJSON = null;
+    let resp;
+    try {
+      resp = await ky.get(`settings/entries?query=userId=="${userId}" and scope=="${scope}" and key=="${key}"`);
+      if (resp.ok) {
+        respJSON = await resp.json();
+        if (respJSON.items.length === 1) {
+          setId(respJSON.items[0].id);
+        }
```
MikeTaylor, line 25: `else` ... ?

src/hooks/usePreferences.js
```diff
@@ -0,0 +1,78 @@
...
+import { useStripes } from '../StripesContext';
+
+export default () => {
+  const { user } = useStripes();
+  const ky = useOkapiKy();
+  const [id, setId] = useState(null);
+  const userId = user.user.id;
+
+  const getPreference = useCallback(async ({ scope, key, callback }) => {
+    // const res = await localforage.getItem(key, callback);
```
MikeTaylor, line 14: What's going on with this commented-out line?

src/hooks/usePreferences.js
```diff
@@ -0,0 +1,78 @@
...
+      key,
+      value,
+      userId
+    };
+
+    // if we didn't store an id upon retrieving user preferences, then one probably doesn't exist,
+    // so we use the `POST` endpoint for saving... 'PUT', and including the id in the path is used for updating
+    if (!id) {
+      try {
+        setId(prefId);
```
MikeTaylor, line 47: Would it make more sense to to move this line after the `ky.post`, so we only set the ID if the operation was successful?

src/hooks/usePreferences.js
```diff
@@ -0,0 +1,78 @@
...
+        return ky.post('settings/entries', { json: payload });
+      } catch (err) {
+        console.log(`error creating preference at scope: ${scope}, and key: ${key} for user: ${userId} - ${err.message}`);
+      }
+    } else {
+      try {
+        return ky.put(`settings/entries/${prefId}`, { json: payload });
+      } catch (err) {
+        console.log(`error updating preference at scope: ${scope}, and key: ${key} for user: ${userId} - ${err.message}`);
```
MikeTaylor, line 56: We can move this outside the `try` block to avoid duplicating it.

src/hooks/usePreferences.js
```diff
@@ -0,0 +1,78 @@
...
+      }
+    }
+  }, [id, ky, userId]);
+
+  const removePreference = useCallback(async ({ scope, key, callback }) => {
+    try {
+      if (id) {
+        await ky.delete(`settings/entries/${id}`);
+        setId(null);
```
MikeTaylor, line 65: I see we use `ky.delete` and _then_ `setId` for deletion, as I think we should for creation.

MikeTaylor (approved):
I've not read every line of this, but it looks goof.

src/components/MainNav/AppOrderProvider.js
zburke: Good question, @MikeTaylor. It is not a new limitation; [old code is identical](https://github.com/folio-org/stripes-core/blame/master/src/components/MainNav/MainNav.js#L134-L136) and hasn't changed in 7+ years. Before that, history is too murky to easily assess, so I don't know if the limitation has _always_ been there but certainly it's been there for a long time.

src/components/MainNav/AppList/AppList.js
JohnC-80: Thanks! need to remove...

src/components/MainNav/AppOrderProvider.js
MikeTaylor: OK, no need to fix this now, then.

src/components/MainNav/AppOrderProvider.js
```diff
@@ -0,0 +1,165 @@
...
+        }
+
+        return false;
+      });
+
+      // find the apps from the platform that are not saved in user-reordered list and tack them on at the end.
+      // these cover permission changes/apps added... can be labeled as 'new' in the preference settings.
+      platformApps.forEach((platApp) => {
+        const orderedIndex = orderedApps.findIndex((oa) => oa.name === platApp.name);
+        if (orderedIndex === -1) {
```
zburke, line 124: Simpler?
```
if (!orderedApps.find((oa) => oa.name === platApp.name)) { ... }
```

src/components/MainNav/MainNav.js
```diff
@@ -27,198 +22,93 @@ import { CurrentAppGroup } from './CurrentApp';
...
+    };
+  }, []); // eslint-disable-line
+
+  useEffect(() => {
+    setSelectedApp(apps.find(entry => entry.active));
+    const nextCurModule = getCurrentModule(modules, location);
+    if (nextCurModule) {
+      setCurModule(getCurrentModule(modules, location));
+      updateQueryResource(location, nextCurModule, stripes.store);
     }
```
zburke, line 80: True confession: I'm lost here. nextCur, setCur, getCurrent ... they're all very similar. Maybe header-comments to explain what each `useEffect()` block does?

zburke (commented):
Code looks great but was occasionally a bit hard for me to follow. Some comments describing data structures, and some describing a function's general purpose, would be very helpful. 

The functionality in the PR description sounds amazing. I'll update this review with further comments once I take this branch out for a spin in a workspace alongside https://github.com/folio-org/ui-myprofile/pull/238.

src/components/MainNav/AppList/components/ResizeContainer/ResizeContainer.js
```diff
@@ -40,9 +40,12 @@ class ResizeContainer extends React.Component {
   }
 
   componentDidUpdate(prevProps) {
+    const { currentAppId, items } = this.props;
     // Update hidden items when the current app ID changes
     // to make sure that no items are hidden behind the current app label
-    if (this.props.currentAppId && this.props.currentAppId !== prevProps.currentAppId) {
+    if (currentAppId !== prevProps.currentAppId ||
+      items !== prevProps.items
```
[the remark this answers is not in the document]
JohnC-80: 👍 can switch to `isEqual` for this... less dependency on what the parent component actually does is better.

src/components/MainNav/AppOrderProvider.js
```diff
@@ -0,0 +1,165 @@
...
+    let navList = []; // contains the ultimate reordered array of app nav items.
+
+    // No length in the persisted apps means apps ordered by alpha (default)
+    if (!orderedApps?.length) {
+      // default ordered apps just contain a subset of the fields in the app object...
+      orderedApps = platformApps.map(({ name }) => ({ name }));
+      navList = platformApps;
+    } else {
+      // reorder apps to the persisted preference value...
+      navList = orderedApps.map((listing) => {
```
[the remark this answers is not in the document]
JohnC-80: `getProvisionedApps` (probably renamed) happens earlier and filters the apps based on perms AND does the filter boolean thing... that array is in whatever order stripe-config sets it as. At that point, you may or may not have loaded in the restored pref and if the pref doesn't exist, we don't worry about changing the order. The `map` maps through the pref array, if any, and results in an array of those app objects, re-ordered according to the pref.

src/components/MainNav/AppOrderProvider.js
```diff
@@ -0,0 +1,165 @@
...
+        const { name: appName } = listing;
+        const appIndex = platformApps.findIndex((module) => appName === module.name);
+
+        if (appIndex !== -1) {
+          return platformApps[appIndex];
+        }
+
+        return false;
```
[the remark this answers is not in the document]
JohnC-80: Actually - this is a good catch - a detail of the old logic that I didn't catch/ wasn't aware of. I might switch to reduce instead so maybe I can do this without further filtering of navList...

src/components/MainNav/AppOrderProvider.js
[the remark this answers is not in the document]
zburke: ☝️ That's a very helpful comment. Let's get it into the codebase.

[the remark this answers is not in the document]
JohnC-80: I knew you were going to say that XD pushing comments up soon.

src/components/MainNav/AppOrderProvider.js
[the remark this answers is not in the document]
JohnC-80: Refactored re-ordering logic to use sort().

src/components/MainNav/AppOrderProvider.js
```diff
@@ -0,0 +1,212 @@
...
+  return apps;
+}
+
+/**
+ * Context provider component for AppOrderContext.
+ * @param {object} props - The component props
+ * @param {React.ReactNode} props.children - The children to render within the provider
+ * @returns {JSX.Element} - The context provider with children
+ */
+export const AppOrderProvider = ({ children }) => { // eslint-disable-line react/prop-types
```
zburke, line 115: Why are you turning off lint here? It's courteous to leave a comment explaining what you know about this line that lint does not.

src/components/MainNav/AppOrderProvider.js
```diff
@@ -0,0 +1,212 @@
...
+    apps.push({
+      displayName: formatMessage({ id: 'stripes-core.settings' }),
+      name: 'settings',
+      id: 'clickable-settings',
+      href: lastVisited.x_settings || SETTINGS_ROUTE,
+      active: pathname.startsWith(SETTINGS_ROUTE),
+      description: formatMessage({ id: 'stripes-core.folioSettings' }),
+      iconData: {
+        src: settingsIcon,
+        alt: formatMessage({ id: 'stripes-core.tenantSettingsLabel' }),
```
zburke, line 100: I see that you are just copying/improving the existing code (by adding i18n) but I don't think "Tenant settings" is the right value here. Why would the icon have a different value than the man item itself? Seems like `stripes-core.folioSettings` is more correct. Agreed?

zburke (approved):
I still have a few open comments that it would be nice to see resolved, but none are show-stoppers. 

This is really nice work, and the extra perks (like refactoring MainNav to functional) are an added bonus. Impressive, most impressive. I fully expect people to go gaga over this feature.

PR #1592, opened 2025-02-12
STCOR-918: *BREAKING* remove token-based authentication code

UladzislauKutarkin:
https://folio-org.atlassian.net/browse/STCOR-918
https://folio-org.atlassian.net/browse/STCOR-922

Here we removed token-based authentication endpoints. Also, removed `useSecureTokens` logic part.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 * Provide `useModuleInfo(path)` hook that maps paths to their implementing modules. Refs STCOR-932.
 * Remove remaining references to stripes.isEureka. Refs STCOR-938.
 * *BREAKING* Upgrade `@folio/stripes-*` dependencies.
+* *BREAKING* remove token-based authentication code. Refs STCOR-918.
 
 ## [10.2.0](https://github.com/folio-org/stripes-core/tree/v10.2.0) (2024-10-11)
 [Full Changelog](https://github.com/folio-org/stripes-core/compare/v10.1.1...v10.2.0)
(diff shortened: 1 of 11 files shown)


zburke (approved):
For reasons I don't know remember, I filed STCOR-918 and STCOR-922 as separate tickets, but that seems silly now. I think it would make sense to handle them together since they all amount to the same change you made here in `requestLogin`. Your call if you want to tack that work onto this PR. 

This ticket description also mentions purging the JWT-related code. It doesn't matter to me which PR that happens under, but if you don't do it here, please move that requirement to STCOR-922 and make sure it happens there :)

UladzislauKutarkin:
> For reasons I don't know remember, I filed STCOR-918 and STCOR-922 as separate tickets, but that seems silly now. I think it would make sense to handle them together since they all amount to the same change you made here in `requestLogin`. Your call if you want to tack that work onto this PR.
> 
> This ticket description also mentions purging the JWT-related code. It doesn't matter to me which PR that happens under, but if you don't do it here, please move that requirement to STCOR-922 and make sure it happens there :)

@zburke you mean delete such conditions const getTenant = (params, token, store) => {
  const tenant = config.useSecureTokens
    ? params?.tenantId
    : parseJWT(token)?.tenant;

  return tenant || store.getState()?.okapi?.tenant;
}; ?

zburke:
Yes, @UladzislauKutarkin . Conditionals like
```
const tenant = config.useSecureTokens ? params?.tenantId : parseJWT(token)?.tenant;
config.useSecureTokens && <SessionEventContainer />
```
should be converted to
```
const tenant = params?.tenantId;
<SessionEventContainer />
```
i.e. just leaving the side of the conditional that evaluates to true. Code that deals with tokens can also be removed because the UI no longer has access to the token. Responses from `.../login` return the token in the body so Stripes would keep it in local storage; reponses from `.../login-with-expiry` return the token in an HTTPOnly cookie which (of course) Stripes cannot access.

PR #1593, opened 2025-02-12
STCOR-931: Change Help icon aria label to just Help in MainNav component

UladzislauKutarkin:
https://folio-org.atlassian.net/browse/STCOR-931

Here we changed the `aria-label` in `MainNav` component from "Help button" to "Help" regarding screen-reader recommendations.


zburke (approved):
This is the right fix but it's the wrong problem. I'm approving based on that, but The Real Right Way to fix this is with a new translations key: 
```
aria-label={intl.formatMessage({ id: 'stripes-core.help' })}
```
as seen a few lines up on line 191. Then it'll have the correct label in any locale, not just English. Please consider implementing this change before you merge.

PR #1594, opened 2025-02-14
[STCOR-942] Partial revert which prevents login screen from using localstorage on login screen

ryandberger:
- Revert login control changes from https://github.com/folio-org/stripes-core/pull/1589 so localstorage is not used determine Tenant ID and Okapi URL values.
- Previously, login was looking at props passed in, which would default to localstorage. This makes builds unable to run with different tenant IDs or Okapi URL values without manually clearing localstorage in the browser.


zburke (approved):
LGTM. We'll need to re-open STCOR-942 to figure out how to implement it while also avoiding this caching problem.

PR #1595, opened 2025-02-17
STCOR-947: Cannot switch affiliations

aidynoJ:
**Approach**: fix updateTenant function by providing fetchUserWithPerms `isKeycloak` parameter.

Refs: [STCOR-947](https://folio-org.atlassian.net/browse/STCOR-947).

### DIFF
--- a/src/loginServices.js
+++ b/src/loginServices.js
@@ -989,7 +989,7 @@ export function updateUser(store, data) {
  */
 export async function updateTenant(okapiConfig, tenant) {
   const okapiSess = await getOkapiSession();
-  const userWithPermsResponse = await fetchUserWithPerms(okapiConfig.url, tenant, okapiConfig.token);
+  const userWithPermsResponse = await fetchUserWithPerms(okapiConfig.url, tenant, okapiConfig.token, false, Boolean(okapiConfig.authnUrl));
   const userWithPerms = await userWithPermsResponse.json();
   await localforage.setItem(SESSION_NAME, { ...okapiSess, tenant, ...spreadUserWithPerms(userWithPerms) });
 }


UladzislauKutarkin (approved):

zburke (commented):
@aidynoJ , do you recall why we [changed loginServices::fetchUserWithPerms](https://github.com/folio-org/stripes-core/commit/fa2ffcf45be74065fff73dff5b8af36e5f64725e#diff-c9d695b1d57404dfed01196c8cf479c0e4cd48583f6c78e7db23402582846864) in PR #1589 to move the "is Eureka" check from an internal one to an explicit flag set by an argument? To be clear, I approved that PR and should have asked this question then ... but reflecting on that change now, I don't understand what it bought us. 

Do you think this fix is the best one, or would it be even better to restore `fetchUserWithPerms` to how it used to work?

aidynoJ:
> @aidynoJ , do you recall why we [changed loginServices::fetchUserWithPerms](https://github.com/folio-org/stripes-core/commit/fa2ffcf45be74065fff73dff5b8af36e5f64725e#diff-c9d695b1d57404dfed01196c8cf479c0e4cd48583f6c78e7db23402582846864) in PR #1589 to move the "is Eureka" check from an internal one to an explicit flag set by an argument? To be clear, I approved that PR and should have asked this question then ... but reflecting on that change now, I don't understand what it bought us. 
> 
> 
> 
> Do you think this fix is the best one, or would it be even better to restore `fetchUserWithPerms` to how it used to work? 

You are right. Now I realize that the purpose of the STCOR-942 was not to get rid of usage of okapi object entirely, but the okapi.url field. I will restore the one with the next follow up PR. 
And I'm really sorry that you had to partially revert my changes back. I should have tested the functionality more carefully.

zburke:
> And I'm really sorry that you had to partially revert my changes back. I should have tested the functionality more carefully.

Pffft, NBD. All of us tested it. None of us caught it. So it's not on you.

PR #1598, opened 2025-02-20
STCOR-946: set tenant context based on authentication response with overrideUser parameter on login

aidynoJ:
Refs [STCOR-946](https://folio-org.atlassian.net/browse/STCOR-946).

**Purpose**. This PR fixes the issue where a member-tenant user authenticates against the central-tenant but is expected to return to the member-tenant after authentication instead of central-tenant. 

**Approach** : OIDCLanding page appears if user is not authenticated yet, so even if you change the affiliation OIDC won't be appeared again, only if you logged out and login.
 Create a function that calls the self endpoint with `overrideUser=true` query parameter. if it returns `originalTenantId` then set it to okapi session and save it. Run this function in OIDCLanding. 
 Calling this function even on the non-consortia envs doesn't affect the session process. 
 Backend side PR https://github.com/folio-org/mod-users-keycloak/pull/133
 
 

https://github.com/user-attachments/assets/92266390-9141-4e42-a41f-a65d54c3acb3


BKadirkhodjaev (approved):

zburke (approved):
Code LGTM. My notes here are really notes to myself to make sure I understand both what these different functions do and why we need them. I think it would be valuable to include some of this documentation in function-comments, but it's up to you. Maybe I am the only one who has trouble following all the threads here.

PR #1602, opened 2025-02-25
STCOR-942: get okapi url from the store, not directly from config in LoginCtrl.js

aidynoJ:
Refs [STCOR-942](https://folio-org.atlassian.net/browse/STCOR-942). 
Previously, changes were partially reverted because the tenant was stored in localStorage after login, and that tenant was then set in the stripes store. The okapi.url is set in the store—not from localStorage or other external sources—but from stripes-config, making it safe to use from the store.
I would like to leave the rest of the functionality as it is for now, but we definitely need to refine our login handling strategies for the cases listed below.
1. single tenant;
2. multi-tenant with multi-select option;
3. multi-tenant, ECS - [Single tenant UX](https://folio-org.atlassian.net/browse/STCOR-946).


zburke (approved):
Code change LGTM; I agree it avoids the problem from PR #1589 (reverted in PR #1594) by only changing `okapiUrl`, whereas that one also changed the handling of `tenant` . 

Please update the CHANGELOG.

PR #1604, opened 2025-02-25
STCOR-952 break QueryStateUpdates out to their own class-based component.

JohnC-80:
Changes from [App reordering](https://folio-org.atlassian.net/browse/STCOR-936) - specifically a functional refactor of `MainNav` unearthed a handful of errors at the app level, causing behaviors with logic that updates query parameters/location. This PR takes the previous code from the class-based `MainNav` and places it in its own component: `QueryStateUpdater` - naming aside, this removes the potentially offending logic from `MainNav`, leaving it to the task of simply rendering the navigation/menus of the UI.

The useEffect implementations of the functional `MainNav` were  written to mirror the `componentDidMount` and `componentDidUpdate` lifestyle methods of the class-based version. When the `location` was added to the `cDM` dependency array in [This PR](https://github.com/folio-org/stripes-core/pull/1600), this meant that the logic would no longer execute stricly on mount/dismount - but anytime that the `location` updated... this would cause page refreshes, as the logic would find a black `prevQueryState` every time and always update the location, causing a refresh.

`componentDidUpdate` carried around its `prevProps`, unlike functional react, where a state might be needed to accomplish such a comparison - but this doesn't seem like an ideal solution for the functional version, so here we are.

### DIFF
--- a/src/components/MainNav/QueryStateUpdater.js
+++ b/src/components/MainNav/QueryStateUpdater.js
@@ -1,0 +1,87 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { compose } from 'redux';
+import { injectIntl } from 'react-intl';
+import { withRouter } from 'react-router';
+import { isEqual } from 'lodash';
+
+import { withModules } from '../Modules';
+
+import {
+  updateQueryResource,
+  getLocationQuery,
+  updateLocation,
+  getCurrentModule,
+  isQueryResourceModule,
+  getQueryResourceState,
+} from '../../locationService';
+
+// onMount of stripes, sync the query state to the location.
+class QueryStateUpdater extends React.Component {
+  static PropTypes = {
+    queryClient: PropTypes.shape({
+      removeQueries: PropTypes.func.isRequired,
+    }).isRequired,
+    stripes: PropTypes.shape({
+      store: PropTypes.shape({
+        subscribe: PropTypes.func.isRequired,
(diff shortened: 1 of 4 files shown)


src/components/MainNav/QueryStateUpdater.js
```diff
@@ -0,0 +1,98 @@
...
+    this._unsubscribe = this.store.subscribe(() => {
+      const { history, location } = this.props;
+      const module = this.curModule;
+      const state = this.store.getState();
+
+      // If user has timed out, force them to log in again.
+      if (state?.okapi?.token && state.okapi.authFailure
+        && find(state.okapi.authFailure, { type: 'error', code: 'user.timeout' })) {
+        this.returnToLogin();
+      }
```
zburke, line 61: This condition will never evaluate to true >= Sunflower because in RTR the token is only in the HTTPOnly cookie, so I guess it doesn't matter that `this.returnToLogin()` is undefined. IOW, this block looks like cruft to me and I think we should just remove it.

JohnC-80: Oh right! We removed this in original refactor... gonna do...

zburke (approved):
LGTM. I appreciate the detailed explanation of why we need this change in the PR description. Consider moving some of into a header comment on QueryStateUpdater to help folks understand why that component exists and that we know about its warts.

PR #1608, opened 2025-03-11
STCOR-955 Improve useModuleInfo hook

vashjs:
This PR enhances the performance of the `useModuleInfo(path)` hook by eliminating unnecessary network requests that previously occurred on every page render.

Changes: 

- Replaced the `react-query` based implementation with `useStripes()`.
- Now leverages module information already available from the `discoverServices` (`modules?full=true`) after login, avoiding extra network calls.
- Since this hook is new and there are no explicit requirements for the returned information, I suggest limiting it to 3 fields -  `{id, name, provides}`, so as not to overload it with `permissions` and other `meta information`, which can be a thousand lines for one module. 
For example, in `bulk-edit`, where this hook will be used first, only the `module name` is needed.

Potential problems:

- Since the module data is collected once at login, there is a small chance that if module information changes later, the user might still see outdated data. However, the probability of this happening is extremely low, I hope 😄 

Some tests:
image

Refs: [STCOR-955](https://folio-org.atlassian.net/browse/STCOR-955)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## 11.1.0 IN PROGRESS
 * CSS Support for printing of results lists content. Refs STCOR-956.
+* Improve useModuleInfo hook. Refs STCOR-955.
 
 ## [11.0.0](https://github.com/folio-org/stripes-core/tree/v11.0.0) (2025-02-24)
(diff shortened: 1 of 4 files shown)


zburke (approved):
This is a great refactor, nicely documented in the PR description. 👏 👏 👏  Thank you!

JohnC-80 (approved):

PR #1611, opened 2025-03-20
STCOR-957 provide useQueryLimit hook

zburke:
The `useQueryLimit()` hook returns a number which can be used in a query's limit clause when retrieving entries from a lookup table and you want _all_ the records.

```
const limit = useQueryLimit();
```

`useQueryLimit()` first looks for a tenant-level mod-settings entry, then for `stripes.config.maxUnpagedResourceCount` from `stripes.config.js`, and then finally gives up and returns the value of the constant `MAX_UNPAGED_RESOURCE_COUNT`, currently set to 2000.

`userTenantPreferences` was shamelessly copied from `usePreferences`; both should be rewritten to share common code. Probably we want `usePreferences` to be the top-level, which looks up a value in user-settings, then tenant-settings.

Refs [STCOR-957](https://folio-org.atlassian.net/browse/STCOR-957)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## 11.1.0 IN PROGRESS
 * CSS Support for printing of results lists content. Refs STCOR-956.
 * Improve useModuleInfo hook. Refs STCOR-955.
+* Provide `useQueryLimit()` hook. Refs STCOR-616, STCOR-617.
 
 ## [11.0.0](https://github.com/folio-org/stripes-core/tree/v11.0.0) (2025-02-24)
 [Full Changelog](https://github.com/folio-org/stripes-core/compare/v10.2.0...v11.0.0)
--- a/package.json
+++ b/package.json
@@ -49,6 +49,16 @@
         "permissionName": "mod-settings.owner.write.stripes-core.prefs.manage",
         "displayName": "UI: update the user's own central preferences, such as order of links in the main navigation.",
         "visible": false
+      },
+      {
+        "permissionName": "mod-settings.global.read.stripes-core.prefs.manage",
+        "displayName": "UI: read the tenant's preferences",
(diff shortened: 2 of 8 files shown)


zburke:
@vashjs , thank you for looking at this and suggesting how to improve it!
> I see async operations inside, and it looks like we won’t be able to use this async hook in components or other hooks, because react won’t let us use await.

Can you elaborate on this? I think it means, in part, that this is wrong:
```
const limit = await useQueryLimit();
```
and it has to be something ugly like
```
const limitPromise = useQueryLimit();
const limit = await limitPromise();
```
With `useEffect()` we would something like this, correct? 
```
const limit = useQueryLimit();
const { ... } = useQuery({
  queryKey: ...,
  queryFn: ...,
  enabled: !!limit,
});
```
I’ve also talked with @JohnC-80 about a stripes-init procedure that would copy everything out of mod-settings into stripes.config. This would mean the only time we have to worry about `useEffect()` is during init, and after that everything would be synchronously available. He has some (legit) concerns about keeping stripes.config in sync with mod-settings when people use `setTenantPreference()`, `deleteTenantPreference()`, etc., but I still wonder if this would be worth the effort in order to offer a nicer API in situations like https://github.com/folio-org/ui-bulk-edit/pull/710 so you wouldn’t even have to mess with useBulkEdit, you could just do 
```
const { stripes : { config: { queryLimit } } } = useStripes();
```
or something similar. Maybe we leave config alone since it already exists and name these values `settings` or something, but you get the idea. Or maybe your first idea -- just wrap this in `useEffect()` -- is enough, and the most KISS solution?

JohnC-80:
@zburke The hook itself would need to store the limit within its own internal state (as @vashjs proposed) and update that state when the promise resolves, and would return the value from its own state (null at first until the promise resolves.)

PR #1615, opened 2025-04-02
STCOR-884: IST modal with long TTL shows even when FLST warning with short TTL is already displayed

aidynoJ:
## Purpose

[STCOR-884](https://folio-org.atlassian.net/browse/STCOR-884) - IST modal with long TTL shows even when FLST warning with short TTL is already displayed

## Approach
The issue was that the `fixedLengthSessionTimeout` timer had a life of its own, and we didn’t have access to it in `SessionEventContainer` to determine whether we could trigger the keepWorking modal.
Just lift the fls timer state up and check if it is less than `idleModalTTL`. If so, don’t show the keepWorking modal; otherwise, show it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
--# Change history for stripes-core
+- # Change history for stripes-core
 
 ## 11.1.0 IN PROGRESS
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-# Change history for stripes-core
+-# Change history for stripes-core
 
 ## 11.1.0 IN PROGRESS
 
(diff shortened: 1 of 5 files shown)


zburke (changes requested):
Moving the FLS setInterval call out of FixedLengthSessionWarning is a nice move that simplifies that component and leaves it with one clear purpose. Good thinking, thank you!

Although I agree that SessionEventContainer is a better home for that logic, I'm not confident it's correctly implemented or correctly explained. I'm glad to see comments in the code -- we need more of that! -- but we need to make sure they align with the code. 

Consider changing some confusing variable names, and update the comments, and then I think this is good to go.

zburke (commented):
Code looks good now, but we need to do a little more cleanup in the comments before we can merge. It's important that the code and the comments say the same thing.

zburke (approved):
Great work; LGTM

PR #1618, opened 2025-04-07
250407 reduce test noise

JohnC-80:
This PR fixes one potentially flakey test by removing a 404 on an image from some static JSON that was included with the test suite. I moved the static JSON into the test file and imported the `svg` as similar tests for `AppIcon` do...

Additionally, added a bit to `setupApplication` to suppress the console noise warnings from `[@formatjs/intl]` about compiling translations that happen any time a `<FormattedMessage>` renders. Leaving some other console messages in there - i.e. Mirage messages about the app trying to access unhandled endpoints.


zburke (approved):
Nah, [BRING THE NOISE](https://www.youtube.com/watch?v=kl1hgXfX5-U)!!!

Alternatively, we could resolve the react-intl complaints by updating `package.json` like 
```
"test:bigtest": "stripes translate compile && stripes test karma",
```

zburke:
@JohnC-80 , another alternative to tweaking console: use stripes-config to tell stripes-core to tell react-intl to STFU by setting 
```
    suppressIntlErrors: true,
    suppressIntlWarnings: true,
```
in [stripes-cli's stripes.config](https://github.com/folio-org/stripes-cli/blob/481713d9ab41ca2d93aecbd27c464d69e51663af/lib/platform/tenant-config.js).

JohnC-80:
It may come up again in the future, but closing this guy to just reduce the noise...

PR #1619, opened 2025-04-08
STCOR-964 (again) compare app-names with localeCompare

zburke:
Use `localeCompare` to perform case-insensitive sort (if that's what a locale expects) of application names, instead of using lodash's `sortBy` which sorts strings by their unicode code points.

This will give us lists like "Data Import, eHoldings, Finance" rather than "Data Import, Finance, eHoldings".

Refs [STCOR-964](https://folio-org.atlassian.net/browse/STCOR-964)


JohnC-80 (approved):
LGTM-lint

PR #1628, opened 2025-04-30
STCOR-968: Enhance locale management to support user and tenant locale configurations from both mod-configuration and mod-settings.

Dmytro-Melnyshyn:
## Description
A new setting in `My profile` -> `Language & Localization` is created, which should use mod-settings API instead of deprecated mod-configuration. It was decided not to make a breaking change for stripes-core and still support mod-configuration in a backwards-compatible way.

  The main changes are in the `loadResources` function. First, if the user can read mod-settings, it fetches locale information using two endpoints (one for tenant and one for the user). When these calls return, it checks if at least one response contains a valid setting. If so, it applies those settings; otherwise, if the user has mod-configuration reading permission, it fetches locale information using the mod-configuration API for both user and tenant settings and applies them. If the user doesn't have mod-settings permission but has mod-configuration permission, it fetches locale information using the mod-configuration API. Additionally, if the user has mod-configuration permission, it fetches plugins and bindings. It also checks if the okapi instance is not in "withoutOkapi" mode, it fetches module information using `discoverServices`. Finally, it returns a flattened array of all the results from the promises.

## Changes in other modules
https://github.com/folio-org/ui-myprofile/pull/251
https://github.com/folio-org/ui-tenant-settings/pull/456
https://github.com/folio-org/ui-developer/pull/468
https://github.com/folio-org/stripes-smart-components/pull/1577
https://github.com/folio-org/ui-plugin-query-builder/pull/246

## Issues
[STCOR-968](https://folio-org.atlassian.net/browse/STCOR-968)

## Screencasts
https://github.com/user-attachments/assets/d1131450-6d43-4e6d-91e8-71d4d0633616


JohnC-80:
@Dmytro-Melnyshyn I like this work a lot - thank you! I think `stripes-core` itself could export a hook that might combine the two levels of preference - tenant and user, with the user preference overriding - so that modules like `ui-myprofile` could consume it rather than have their own logic similar to the `useSettings` hook that you have in `ui-myprofile`.

src/hooks/usePreferences.js
```diff
@@ -7,41 +10,50 @@ import { useStripes } from '../StripesContext';
...
+  // use a ref to store the id of the setting so that it will be available for the next request when we chain.
+  const idRef = useRef(null);
+  const setId = useCallback(newId => { idRef.current = newId; }, []);
+
   const userId = user.user.id;
 
-  const getPreference = useCallback(async ({ scope, key }) => { // eslint-disable-line
+  const getPreference = useCallback(async ({ scope, key, userId: providedUserId }) => { // eslint-disable-line
     // get preferences by querying preferences with the userId, scope, and key...
     // this will return an array that's likely a single item long. We store the item's id in state so that we can update it later.
```
JohnC-80, line 22:
```suggestion
    // this will return an array that's likely a single item long. We store the item's id in a ref so that we can update the preference later.
```
Minor - just a comment that would need to be updated given the changes to the implementation.

mariia-aloshyna (approved):

JohnC-80 (approved):

PR #1633, opened 2025-05-15
STCOR-968: Use the isEnabled flag, which specifies whether tenant settings should be applied when handling locale (follow-up).

Dmytro-Melnyshyn:
## Description
User locale settings must follow the tenant's locale settings when the tenant's locale is selected in `Settings - My profile - language localization`. Currently in stripes-core we apply the tenant's locale settings only if user has not set their own locale or their locale is the same as the tenant's locale `if (!userLocaleSettings?.locale || userLocaleSettings?.locale === tenantLocaleSettings?.locale) {}`, but the tenant's locale can be changed and the condition ` userLocaleSettings?.locale === tenantLocaleSettings?.locale` will not be true anymore. The solution is to add the `isEnabled` flag which will indicate whether we need to use user's settings or not. The flag is added in https://github.com/folio-org/ui-myprofile/pull/252.

The recording of the issue is the first [video ](https://folio-org.atlassian.net/browse/UIMPROF-107?focusedCommentId=263844).

## Issues
[STCOR-968](https://folio-org.atlassian.net/browse/STCOR-968)

## Screencasts

https://github.com/user-attachments/assets/96908ea3-a7f2-49d4-93fd-61483463ee92

### DIFF
--- a/src/loginServices.js
+++ b/src/loginServices.js
@@ -531,8 +531,7 @@ const processLocaleSettings = (store, tenantLocaleData, userLocaleData) => {
   let timezone = userLocaleSettings?.timezone || tenantLocaleSettings?.timezone;
   let currency = userLocaleSettings?.currency || tenantLocaleSettings?.currency;
 
-  // we should use tenant's settings if user has not set their own locale or their locale is the same as tenant's locale
-  if (!userLocaleSettings?.locale || userLocaleSettings?.locale === tenantLocaleSettings?.locale) {
+  if (!userLocaleSettings?.isEnabled) {
     locale = tenantLocaleSettings?.locale;
     numberingSystem = tenantLocaleSettings?.numberingSystem;
     timezone = tenantLocaleSettings?.timezone;
--- a/src/loginServices.test.js
+++ b/src/loginServices.test.js
@@ -870,6 +870,7 @@ describe('loadResources', () => {
           locale: 'en-GB',
           timezone: 'Europe/London',
           currency: 'GBP',
+          isEnabled: true,
         },
       },
     ],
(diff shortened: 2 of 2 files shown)


mariia-aloshyna (approved):

zburke:
> User locale settings must follow the tenant's locale settings when the tenant's locale is selected in Settings - My profile - language localization

Sorry, @Dmytro-Melnyshyn, I don't understand this statement. My understanding of preferences -- _all_ preferences, not just language and localization -- is that they exist in a hierarchy and we should use one of four values, in descending order: 

* user setting from storage (user-runtime)
* tenant setting from storage (tenant-runtime)
* tenant setting from stripes.config.js (tenant-build-time)
* hard-coded value from source-code (global-default)

That's it. It doesn't matter whether the user and tenant values match, so we don't ever need to compare them, nor do we need an `isEnabled` flag, because as we descend through the hierarchy (user-runtime > tenant-runtime > tenant-build-time > default), we use the first value we find. A value's existence functions as its `isEnabled` flag.

Maybe this means we need a "Reset to default" button (same as we have in Settings > My profile > Application display order) that removes a user's runtime values from storage?

Dmytro-Melnyshyn:
> If the My profile > Locale (for language display, date format etc.) was changed then changed back to the default (Tenant > Locale (for language display, date format etc.)  selection and then any change to Tenant > Locale (for language display, date format etc.)  selection should be the My profile > Locale (for language display, date format etc.)  selection

@zburke, we have a requirement to use the tenant's settings when the user's settings locale matches the tenant's settings locale ([original story](https://folio-org.atlassian.net/browse/UIMPROF-107)).

> 1. user setting from storage (user-runtime)
> 2. tenant setting from storage (tenant-runtime)
> 3. tenant setting from stripes.config.js (tenant-build-time)
> 4. hard-coded value from source-code (global-default)

The order above is still working.

> Maybe this means we need a "Reset to default" button (same as we have in Settings > My profile > Application display order) that removes a user's runtime values from storage?

The reset button is a good solution, however, it was rejected by PO during refinement.

zburke:
@Dmytro-Melnyshyn, this story is really hard to parse. I think I finally understand the last bullet. Please correct me if I am wrong. I am reading it like this: 

* when the tenant-locale is set to A
* if the user changes their locale to B, and then back to A (matching the tenant's selection), and then the tenant changes their locale to C, the user's locale should be C. 

Wouldn't the logic be simpler to just remove the user-setting when they switch it to match the tenant's? Why is it better to use an `isActive` flag? 

I still think the user experience would be better with an explicit "reset to default" button, especially since that's what we implemented for "Application display order", but I don't need to pick a fight with the PO if the choice has already been made.

Dmytro-Melnyshyn:
@zburke correct.

> Wouldn't the logic be simpler to just remove the user-setting when they switch it to match the tenant's? Why is it better to use an isActive flag?

If we go with removing the user setting:

* when the tenant-locale is set to A
* if the user changes their locale to B
* if the user changes their locale's numbering system
* if the user changes their locale back to A (matching the tenant's selection)
* if the user changes their locale back to B 

The numbering system will be lost, but it is a part of the full locale.

PR #1634, opened 2025-05-15
STCOR-972 passing tenant as a URL param to keycloak redirect_uri

BogdanDenis:
## Description
I'm going to use @zburke explanation on this because I couldn't have explained the issue better myself

> In an ECS setup, there are three required steps to authentication; keycloak is step-2, and skipping step-1 causes step-3 to fail:
> 1. choose a tenant and save it to local-storage, then redirect to keycloak
> 2. authenticate in keycloak, then redirect to stripes with an OTP in the URL
> 3. make an API request to trade the OTP for AT/RT cookies, /authn/token?otp=${OTP from step-2} with the X-Okapi-Tenant header set to the tenant from step-1
>
> Bookmarking the authentication URL in step-2 means the tenant is never stored to local-storage, thus causing step-3 to fail. 
IIRC, we can pass values to Keycloak in the URL and it will include them along with the OTP when it redirects back to Stripes.

Instead of storing selected `tenant` and `client_id` in `localStorage`, we can in step 1 pass them as part of `redirect_uri`, so that we can read them from url and use in step 3.
We're still keeping code related to storing tenant in localStorage, because we actually need to keep the tenant that a user has initially logged in with, so if they change affiliation - we can logout with the initial tenant.

## Screenshots

https://github.com/user-attachments/assets/6c031f1d-5659-4049-99fc-419af59bbce8

## Issues
[STCOR-972](https://folio-org.atlassian.net/browse/STCOR-972)

### DIFF
--- a/src/components/OIDCLanding.js
+++ b/src/components/OIDCLanding.js
@@ -17,7 +17,7 @@ import {
   getOIDCRedirectUri,
   requestUserWithPerms,
   setTokenExpiry,
-  TENANT_LOCAL_STORAGE_KEY,
+  storeLogoutTenant,
 } from '../loginServices';
 import { useStripes } from '../StripesContext';
 
@@ -87,9 +87,9 @@ const OIDCLanding = () => {
               })
               .then(() => {
                 // for login itself we're storing selected tenant and clientId in the url
-                // but we still need to store the same data in localStorage for logout purposes
+                // but we still need to store tenantId in localStorage for logout purposes
                 // `logout` function in loginServices.js provides details about this.
-                localStorage.setItem(TENANT_LOCAL_STORAGE_KEY, JSON.stringify({ tenantName: okapi.tenant, clientId: okapi.clientId }));
+                storeLogoutTenant(okapi.tenant);
               });
           } else {
             return resp.json().then((error) => {
--- a/src/components/OIDCLanding.js
+++ b/src/components/OIDCLanding.js
@@ -13,10 +13,14 @@ import {
 } from '@folio/stripes-components';
 
 import OrganizationLogo from './OrganizationLogo';
-import { requestUserWithPerms, setTokenExpiry } from '../loginServices';
+import {
+  getOIDCRedirectUri,
+  requestUserWithPerms,
+  setTokenExpiry,
+} from '../loginServices';
+import { useStripes } from '../StripesContext';
 
 import css from './Front.css';
-import { useStripes } from '../StripesContext';
 
 /**
  * OIDCLanding: un-authenticated route handler for /oidc-landing.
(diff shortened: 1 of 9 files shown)


src/loginServices.js
```diff
@@ -131,10 +131,18 @@ export const setUnauthorizedPathToSession = (pathname) => {
 };
 export const getUnauthorizedPathFromSession = () => sessionStorage.getItem(UNAUTHORIZED_PATH);
 
-const TENANT_LOCAL_STORAGE_KEY = 'tenant';
+export const getOIDCRedirectUri = (tenant, clientId) => {
+  // we need to use `encodeURIComponent` to separate `redirect_uri` URL parameters from the rest of URL parameters that `redirect_uri` itself is part of
+  return encodeURIComponent(`${window.location.protocol}//${window.location.host}/oidc-landing?tenant=${tenant}&client_id=${clientId}`);
```
zburke, line 136: Petty: why rename `clientId` to `client_id` in the string you return? Just use `clientId` consistently, here and in `getStoredTenant()` too.

zburke (changes requested):
Thanks for picking this up. I appreciate introducing `getOIDCRedirectUri()` to make sure that value is always consistent, and the other minor cleanup that is in this PR. 

Looks like we need to use `getOIDCRedirectUri()` in [AuthnLogin](https://github.com/folio-org/stripes-core/blob/325d9e3d0810d72a505891b946fbd056271e82a9/src/components/AuthnLogin/AuthnLogin.js#L56) for this to work for single-tenant installs as well. For ECS, this works great!

Can we also remove related code in [App.js](https://github.com/folio-org/stripes-core/blob/325d9e3d0810d72a505891b946fbd056271e82a9/src/App.js#L74)? We should either keep/rewrite the test for `getStoredTenant`, or remove the import so lint doesn't complain.

src/loginServices.js
zburke: Eh, I see other params to keycloak use snake_case, so `client_id` makes sense in that sense. Never mind; sorry for the noise.

OleksandrHladchenko1 (approved):

src/loginServices.js
```diff
@@ -752,6 +767,7 @@ export async function logout(okapiUrl, store, queryClient) {
     // clear shared storage
     .then(localforage.removeItem(SESSION_NAME))
     .then(localforage.removeItem('loginResponse'))
+    .then(() => localStorage.removeItem(TENANT_LOCAL_STORAGE_KEY))
```
zburke, line 770: Can we move this above with the other `localStorage.removeItem()` entries? If not, please add a comment explaining why it needs to be handled separately.

src/components/OIDCLanding.js
```diff
@@ -77,6 +84,12 @@ const OIDCLanding = () => {
               })
               .then(() => {
                 return requestUserWithPerms(okapi.url, store, okapi.tenant);
+              })
+              .then(() => {
+                // for login itself we're storing selected tenant and clientId in the url
+                // but we still need to store the same data in localStorage for logout purposes
+                // `logout` function in loginServices.js provides details about this.
+                localStorage.setItem(TENANT_LOCAL_STORAGE_KEY, JSON.stringify({ tenantName: okapi.tenant, clientId: okapi.clientId }));
```
zburke, line 92: Do we need `clientId` for logout? I think it can be omitted.

zburke (approved):
There may be room for a little more clean up, but on the whole, these changes look good/correct to me. Thank you!

PR #1642, opened 2025-08-07
STCOR-978 Use BroadcastChannel to isolate token to last-focused window.

JohnC-80:
## [STCOR-978](https://folio-org.atlassian.net/browse/STCOR-978)

This PR resolves an issue where users experienced an abrupt logout due to failures in token rotation across multiple tabs of usage - one bad apple among all of their open tabs spoiled it for the rest.

### Approach
Only a single, active tab is able to actually rotate. The rest just schedule their next rotation and *don't* actually rotate. This will be the last tab that the user interacted with and will persist through usage of other applications.

With this approach, each new tab (single `FFetch` instance) creates a unique id for the tab and sets up a `focusin` handler so that when a new tab is focused, it will broadcast the focused tab's unique id so that other windows can set it in their session and use it to check whether token rotation should occur. If the id in session doesn't match the current tab's id, rotation will not occur in that tab. All tabs share the same cookie, so as long as its kept alive by the current window, it will remain for other tabs to use. Each tab stores the last-focused tab in their session.

Other attempts:
- relying solely on `document.hasFocus()` to dictate whether or not a window *can* rotate. This is limited because the browser tab that only has the tab focus (and focus isn't in the UI) will not be considered focused, leaving the status in limbo. It also causes rtr to not happen if other apps are in use when rotation happens - and we don't need that!

- syncing via `storage` event.  This seemed very difficult to debug - the events didn't seem to fire with sessionStorage updates across windows. I ended up switching to a `localStorage` implementation, but even that was problematic. The events just didn't want to stick across windows. The persistent nature of `localStorage` disqualifies it as a lone possibility since the previously generated window id would stick around and fail to match a new window after the entire application is closed.

The video displays the logic in action (with rtr logging set in stripes-config) you can see when the focus changes and the newly focused tab's id is broadcast to all of the other tabs...

https://github.com/user-attachments/assets/fb1fffee-c698-4c56-9904-9608bc955f5f


src/components/Root/token-util.js
```diff
@@ -212,6 +214,27 @@ export const isRotating = () => {
...
   logger.log('rtr', '** RTR ...');
 
+  /**
+   * Check the windowId against the active windowId stored in sessionStorage.
+   * If they don't match, this window is not the active one, so we skip the
+   * rotation request and return a promise that resolves
+   * with the current token expiry data. This is to prevent multiple windows
+   * from trying to rotate the token at the same time.
+   */
+  const thisWindowId = window.stripesRTRWindowId;
```
BogdanDenis, line 224: I'm confused about this logic - aren't `window` and `sessionStorage` tab-specific, and so comparing them would always be equal?

JohnC-80: `window.stripesRTRWindowId` stores the UUID of each window - it's only set once per tab and does not change.  
Session storage is updated across tabs when a new tab is focused via `BroadcastChannel` sending out the handled `message` to the other tabs to update their `SessionStorage`value with the `stripesRTRWindowId` of the active window.

PR #1644, opened 2025-08-14
STCOR-978 Focus refresh

JohnC-80:
This PR appends the focus handler to check the access token expiry and possibly refresh before any data calls can be made.
This resolves a situation where a user navigates away from FOLIO entirely (different window, app, etc) and closes their most recent window without going to any previously used FOLIO tab... once they refocus a FOLIO tab, if the at is expired, the logic will refresh.

### DIFF
--- a/src/components/Root/FFetch.js
+++ b/src/components/Root/FFetch.js
@@ -49,7 +49,7 @@ import {
   setRtrFlsWarningTimeout,
 } from '../../okapiActions';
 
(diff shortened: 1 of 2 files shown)


zburke (approved):
This follow-up to #1642 LGTM. 

As we have discussed, I still worry a little about the race condition of a fetch arriving immediately after focus but before this handler has had time to initiate RTR, but this substantially reduces that window. I think it's good enough and a good improvement over what we had in place before. 🤝

PR #1649, opened 2025-08-28
STCOR-985 correctly implement tenant-selection on init

zburke:
During initialization, derive the tenant via one of the following sources, in order of preference:
1. URL in the `tenant` and `client_id` search params
2. stripes-config in the `config.tenantOptions` object
3. stripes-config in the `okapi` object

Previous implementations ignored option-2, causing tenant to be undefined in Eureka-FOLIO builds where the (deprecated) `okapi` object had been omitted from `stripes.config.js` and a session has not been started, e.g. because the user has not authenticated and is submitting the "Forgot password" form.

Refs [STCOR-985](https://folio-org.atlassian.net/browse/STCOR-985)

### DIFF
--- a/src/loginServices.js
+++ b/src/loginServices.js
@@ -136,15 +136,6 @@ export const getOIDCRedirectUri = (tenant, clientId) => {
   return encodeURIComponent(`${window.location.protocol}//${window.location.host}/oidc-landing?tenant=${tenant}&client_id=${clientId}`);
 };
 
-export const getTenantAndClientIdFromLoginURL = () => {
-  const urlParams = new URLSearchParams(window.location.search);
-
-  return {
-    tenantName: urlParams.get('tenant'),
-    clientId: urlParams.get('client_id'),
-  };
-};
-
 // export config values for storing user locale
 export const userLocaleConfig = {
   'configName': 'localeSettings',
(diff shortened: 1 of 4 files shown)


src/loginServices.js
```diff
@@ -136,12 +136,29 @@ export const getOIDCRedirectUri = (tenant, clientId) => {
...
+  let clientId = urlParams.get('client_id');
+  if (tenant && clientId) {
+    return { tenant, clientId };
+  }
+
+  // derive from stripes.config.js::config::tenantOptions
+  if (stripesConfig.tenantOptions && Object.keys(stripesConfig?.tenantOptions).length === 1) {
+    const key = Object.keys(stripesConfig.tenantOptions)[0];
+    tenant = stripesConfig.tenantOptions[key]?.name;
+    clientId = stripesConfig.tenantOptions[key]?.clientId;
```
JohnC-80, line 152: Maybe a comment with the anticipated shape of the object this is accessing or describing the arbitrary(?) key i.e. `stripesConfig.tenantoptions.foo.name`

JohnC-80 (approved):
LGTM - minor thing

src/loginServices.js
```diff
@@ -136,12 +136,44 @@ export const getOIDCRedirectUri = (tenant, clientId) => {
...
+  // derive from the URL
   const urlParams = new URLSearchParams(window.location.search);
+  let tenant = urlParams.get('tenant');
+  let clientId = urlParams.get('client_id');
+  if (tenant && clientId) {
+    return { tenant, clientId };
+  }
+
+  // derive from stripes.config.js::config::tenantOptions
+  if (stripesConfig.tenantOptions && Object.keys(stripesConfig?.tenantOptions).length === 1) {
```
usavkov-epam, line 164: I think it's safer to use optional chaining here `?.`: this place of code breaks a lot of tests (at least in the ACQ modules).

zburke: Yep, fixed in #1655

PR #1652, opened 2025-09-02
STCOR-986 handle OTP in StrictMode

zburke:
Effects run twice in `<StrictMode>`, but OTP must only happen once; make it so.

This PR changes the way OIDCLanding handles OTP exchange, moving it from an effect (which will run x2 in development when `<StrictMode>` is turned on, which of course we want it to be), [outside the component lifecycle](https://react.dev/learn/synchronizing-with-effects#not-an-effect-initializing-the-application) so it only runs once no matter how many times the component is [mounted/unmounted/remounted in development-mode](https://react.dev/learn/synchronizing-with-effects#how-to-handle-the-effect-firing-twice-in-development).

It may bother you that there is both an event-handler and an effect to handle the same thing, and that's fair. The problem is, there is a race condition between the OTP-exchange request and the component loading, and we want to initialize the session no matter who wins the race:
* if the component loads before the request finishes, then storage will be empty when the effect runs so nothing will happen then 😢 but the event handler will catch the event emitted when the request finally finishes 😍 
* if the request finishes before the component loads, then there won’t be a handler to listen for the event so nothing will happen 😢 but the effect runs when the component mounts and will find the data in storage 😍 

Refs [STCOR-986](https://folio-org.atlassian.net/browse/STCOR-986)

### DIFF
--- a/src/components/OIDCLanding.js
+++ b/src/components/OIDCLanding.js
@@ -1,155 +1,169 @@
-import React, { useEffect, useState } from 'react';
-import { useLocation } from 'react-router-dom';
-import queryString from 'query-string';
+import React, { useEffect, useRef, useState } from 'react';
 import { useStore } from 'react-redux';
-import { FormattedMessage } from 'react-intl';
+import { okapi as okapiConfig, config } from 'stripes-config';
 
 import {
-  Button,
-  Col,
-  Headline,
   Loading,
-  Row,
 } from '@folio/stripes-components';
 
-import OrganizationLogo from './OrganizationLogo';
 import {
+  getLoginTenant,
   getOIDCRedirectUri,
+  getTokenExpiry,
   requestUserWithPerms,
   setTokenExpiry,
   storeLogoutTenant,
 } from '../loginServices';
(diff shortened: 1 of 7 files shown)


JohnC-80 (approved):

src/components/OIDCLanding.js
```diff
@@ -1,155 +1,185 @@
...
+      .catch(e => {
+        // eslint-disable-next-line no-console
+        console.error('@@ Oh, snap, OTP exchange failed!', e);
+        // copy the error into the global otpError
+        otpError = e;
+      });
+  }
+};
+
+exchangeOtp(setTokenExpiry, storeLogoutTenant);
```
NikitaSedyx, line 102: @zburke what is the purpose of calling this function on module level? it's called in tests when at least smth is imported from stripes-core

zburke: These are good points, @NikitaSedyx. This PR solves a real problem, but it does so in a pretty gross way. See PR #1656 to revert this PR. I reopened the ticket and will work on a better approach.

PR #1661, opened 2025-09-30
STCOR-976: Implement async modules loading infrastructure and update related components.

Dmytro-Melnyshyn:
## Purpose
Make `modules` loading asynchronous to prepare for future dynamic module loading capabilities. Currently, `modules` is imported synchronously from `stripes-config` at build time.

## Description
* The `entitlementService.js` is added. The new service provides Promise-based `getModules()` function. Currently wraps static `stripes-config` modules in Promise.
* The `getModules()` is applied in the `ModuleTranslator` provider to share the modules.
* The `getModules()` is also applied in the `App` component; perhaps, it might be worth wrapping the `App` component in `ModuleTranslator`.
* Updated `ResizeContainer`. Modules now appear asynchronously, so we need to recalculate the width of the nav list items not only on mount, but also when updating modules.
* Removed the `withModules` HOC from the `Root` component, since the provider of modules resides below the `Root`, and desn't provide the requested modules. `modules` is now passed down from the parent `App` component, since `modules` is also necessary there.

## Issues
[STCOR-976](https://folio-org.atlassian.net/browse/STCOR-976)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@
 * Remove `RTR_IS_ROTATING` from local storage before dispatching the `RTR_SUCCESS_EVENT` to correctly determine the `RTR_IS_ROTATING` state in the `rotationHandler` function. Refs STCOR-983.
 * Correctly implement tenant-selection ringdown during application init. Refs STCOR-985.
 * Conform `@apollo/client` init to the v4 API to suppress warnings. Refs STCOR-992.
+* Implement async `modules` loading infrastructure and update related components. Refs STCOR-976.
 
 ## [11.0.0](https://github.com/folio-org/stripes-core/tree/v11.0.0) (2025-02-24)
 [Full Changelog](https://github.com/folio-org/stripes-core/compare/v10.2.0...v11.0.0)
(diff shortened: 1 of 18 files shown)


zburke:
@JohnC-80, I’d appreciate if you can take a look at this too, especially the changes in ResizeContainer. 

I cloned platform-complete and then pinned stripes-core to this branch and everything works perfectly AFAICT 🏆 I’m combing through the update to make sure I understand it all, but so far everything is coming up roses.

JohnC-80 (approved):
Looks good! Very nice work @Dmytro-Melnyshyn!

Dmytro-Melnyshyn:
> Historically, that didn't happen because ModuleTranslator needs access to the IntlProvider that is buried inside App > Root > IntlProvider. Since PR https://github.com/folio-org/stripes-core/pull/1476 / STCOR-851, however, I think we could move IntlProvider up to the top level, and then update it if necessary inside Root when tenant-specific and user-specific settings become available.

@zburke, that's a good point. Moving `IntlProvider` up might be the right approach, but I think it would be better to handle that in a separate ticket since it could introduce breaking changes and would require additional testing. Given that the sprint ends today, I'd suggest merging this PR as is and then tackling that improvement separately. What do you think?

zburke (approved):
This is awesome! Thank you. I appreciate the high-level comments in the PR description that lay out the scope of work completed here, and the potential for future developments.

PR #1704, opened 2026-03-27
STCOR-891 Init the session in the last active tenant on logout by timeouts (don't handle multi-tab/window issues)

usavkov-epam:
### Purpose

When a user's session expires (idle timeout, fixed-length session expiry, or RTR error), they are redirected to `/logout-timeout`. After re-authenticating, FOLIO previously dropped them at the root (`/`), losing their last-active tenant affiliation (relevant in ECS/consortia environments) and the URL they were on. This PR preserves both so the user is returned to their exact previous context.

### Approach

Two things must be preserved across the forced-logout boundary and restored on the next login:

1. **Last-active tenant** — saved to `sessionStorage` as `unauthorized_tenant` at the moment a forced-logout event fires in `SessionEventContainer` (RTR error, idle timeout, or fixed-length session expiry). This happens _before_ navigating to `/logout-timeout`, ensuring the tenant is captured regardless of what `Logout` does.

   Additionally, `AuthnLogin` (the authn-URL login flow) caches the selected tenant into `sessionStorage` whenever the user picks one from the tenant dropdown — but only if no forced-logout path is already stored, to avoid overwriting a tenant saved by a prior timeout.

2. **Pre-logout URL** — already saved to `sessionStorage` as `unauthorized_path` by `SessionEventContainer` before redirecting to `/logout-timeout`.

On the next login, each entry-point (`LoginCtrl`, `OIDCLanding`, `useSSOSession`) reads the saved tenant via the new `consumeUnauthorizedTenantFromSession()` helper (which atomically clears it) and passes it as `options.preservedSessionTenant` to the session-creation chain:

```
requestLogin / requestUserWithPerms
  → processOkapiSession
    → createOkapiSession
      → resolveSessionTenant(tenant, data, options.preservedSessionTenant)
          // Priority: preserved > originalTenantId (consortia) > tenant (non-consortia) > default tenant
```

The saved tenant takes the highest priority, overriding even the `originalTenantId` returned by the login response (which always reflects the user's primary affiliation, not the last-active member tenant).

### Changes

**`src/loginServices.js`**
- Added `unauthorized_tenant` sessionStorage helpers: `setUnauthorizedTenantToSession`, `getUnauthorizedTenantFromSession`, `removeUnauthorizedTenantFromSession`, and the atomic `consumeUnauthorizedTenantFromSession()`.
- Added `resolveSessionTenant(defaultTenant, apiResponse, preservedTenant)` helper to centralise session-tenant priority logic (reduces code duplication).
- Added `@typedef OkapiSessionOptions` documenting the `preservedSessionTenant` option in one place.
- Updated `createOkapiSession`, `processOkapiSession`, `requestLogin`, and `requestUserWithPerms` to accept and thread `OkapiSessionOptions` through the call chain.
- `createOkapiSession` now delegates to `resolveSessionTenant` instead of the inline `data.originalTenantId || data.tenant || tenant`.
- Fixed missing trailing semicolon on `localforage.removeItem(stripesHubAPI.REMOTE_LIST_KEY)`.

**`src/components/SessionEventContainer/SessionEventContainer.js`**
- Added `setUnauthorizedTenantPathToSession(tenant, path)` helper that atomically writes both `unauthorized_tenant` and `unauthorized_path` to `sessionStorage`.
- Updated all forced-logout event handlers (`thisWindowRtrError`, `thisWindowRtrIstTimeout`, `thisWindowRtrFlsTimeout`, both branches of `otherWindowStorage`) to call `setUnauthorizedTenantPathToSession(stripes.okapi.tenant)` instead of `setUnauthorizedPathToSession()`, so the current tenant is captured alongside the path at the point the session ends.

**`src/components/AuthnLogin/AuthnLogin.js`**
- Imports `setUnauthorizedTenantToSession` and `getUnauthorizedPathFromSession`.
- In `setTenant`, when the authn-URL flow is active (`okapi.authnUrl` is set) and no forced-logout path is already stored, caches the selected tenant to `sessionStorage` so it can be restored if the user is timed out while on that tenant.

**`src/components/Login/LoginCtrl.js`**
- Passes `consumeUnauthorizedTenantFromSession()` as `options.preservedSessionTenant` to `requestLogin`.

**`src/components/OIDCLanding.js`**
- Passes `consumeUnauthorizedTenantFromSession()` as `options.preservedSessionTenant` to `requestUserWithPerms`.

**`src/components/SSOLanding/useSSOSession.js`**
- Passes `consumeUnauthorizedTenantFromSession()` as `options.preservedSessionTenant` to `requestUserWithPerms`.

**`src/components/SSOLanding/useSSOSession.test.js`**
- Updated test assertions to expect the new `options` argument (`{ preservedSessionTenant: ... }`) in `requestUserWithPerms` calls.
- Uses real implementations of `consumeUnauthorizedTenantFromSession` and `getUnauthorizedTenantFromSession` via `jest.requireActual` to keep assertions accurate.

### Login tenant vs. Session (preserved) tenant

**Login tenant** — the tenant sent in the `X-Okapi-Tenant` header of the authentication request (`/bl-users/login-with-expiry` or `/_self`). It identifies *where* to authenticate, i.e. the user's home/primary tenant. In consortia environments this is always the central tenant.

**Session tenant** — the tenant stored in the session and used for all subsequent API calls. It represents *where the user was working*, which can differ from the login tenant when the user has switched affiliation to a member tenant mid-session.

The login response itself reflects the home tenant (`data.tenant` / `data.originalTenantId`), so naively using it as the session tenant would silently reset the user back to their home tenant on every re-authentication.

**Priority chain in `resolveSessionTenant`:**

```
preserved tenant       ← last active tenant saved before forced logout
  || originalTenantId  ← consortia home tenant from login response
  || data.tenant       ← non-consortia tenant from login response
  || request tenant    ← fallback: what we sent in the header
```

**Preserved tenant** is written to `sessionStorage` (`unauthorized_tenant`) at the moment a forced-logout event fires — before the user ever sees the login page. On the next successful login it is consumed atomically and injected as `options.preservedSessionTenant`, restoring the user to exactly the tenant they were on.

#### Single entry ECS setup:

https://github.com/user-attachments/assets/d070e94b-cb07-4992-a652-c7468ca3dd8a

#### Multi-option ECS setup:

https://github.com/user-attachments/assets/057ae21f-c9eb-4f2c-b898-753f5f5a4d04

#### SSO Login (ECS):

https://github.com/user-attachments/assets/5a866830-d713-470d-bd51-db8edd4711f8

## Open questions:

- _What if a user has several tabs in the browser with different tenants and clicks logout on one of them?_
    > **_Answer_** - The first tab that initializes the session will restore the correct tenant for the redirect. Restoring the tenant in other tabs is not guaranteed, as the session may have already been initialized in another tab, causing this step to be skipped.
   >
   > Currently, the ECS initialization logic (loading tenants and updating the session to include consortium information) resides in `ui-consortia-settings`. Updating the tenant within an already initialized session from `stripes-core` may require architectural changes, so this part is out of scope for this PR.
 

- _What if a user logs in to a different account after a timeout logout?_ - This issue is partly related to current changes, but is also directly related to the preservation of the URL, so it should be considered separately.


usavkov-epam:
Hi @folio-org/stripes-force !

I’d really appreciate your input on this PR. I’ve implemented a solution that allows a user to restore their last tenant (to stay consistent with the persisted URL) if they were forcefully logged out. During session re-initialization, we can specify the desired tenant, which is what this code does.

However, the PR does not fully address the issue when multiple tabs/windows are used simultaneously. More details are provided in the PR description.

zburke:
@usavkov-epam, there's a lot here, and I have not looked at it all in detail. That said, in response to this specific issue:
> the PR does not fully address the issue when multiple tabs/windows are used simultaneously

My gut reaction is that this value should be set via `localStorage` (which is cross-tab) rather than `sessionStorage` (which is per-tab). The vocabulary is difficult here because when a user authenticates and gets a cookie, that initiates a "session", but that cookie is valid across multiple tabs, and we store the data about this "session" (the response from the `/users-bl/_self` API request) in localStorage. 

How would changing the storage medium to `localStorage` affect this PR?

usavkov-epam:
@zburke In my opinion, it makes sense to store the tenant in `sessionStorage`, similar to how the URL is stored, so that the user can fully restore their session. However, the issue with multiple tabs is not so much about where the tenant is stored, but rather how it is restored.

To briefly answer the question:

> *How would changing the storage medium to localStorage affect this PR?*

It wouldn’t affect it. If the session is initialized in another tab, the current tab will open in the tenant where the session was initialized (`sessionStorage` vs `localStorage` doesn’t matter).

Let me illustrate this with a concrete multi-tab scenario:

A user ("Test user") with a primary affiliation of "Consortium" has two tabs open:
- one in the **"College"** tenant at `/inventory/view/{college-instance-uuid}`
- another in the **"University"** tenant at `/inventory/view/{university-instance-uuid}`

Assume the user is logged out from the **"University"** tab due to a timeout. During logout, the URL is stored in sessionStorage for redirect after re-login.

This PR also proposes storing the tenant in the same way for consistency. Otherwise, the user would log in to the **"Consortium"** tenant and be redirected to a record that does not exist in that tenant.

Now, suppose the user logs back in via the **"College"** tab. In this case, the session is initialized with the **"College"** tenant, and the user is redirected to a **valid** record (`/inventory/view/{college-instance-uuid}`).

However, if the user then switches to the **"University"** tab (which shows the “Log in again” UI) and clicks the button, they will be redirected to a **non-existent** record `/inventory/view/{university-instance-uuid}` in the **"College"** tenant—because the session has already been initialized in **"College"**.

Currently, there is no logic in that tab to switch the tenant. Therefore, it doesn’t matter whether the tenant is stored in `sessionStorage` or `localStorage`—the user is already being redirected using a URL from a different tenant.

The problem is not strictly tied to timeout-based logout. Even a manual logout with multiple tabs can lead to the same situation.

| img |
|:--:|
| * The image is generated using AI |

usavkov-epam:
@zburke @folio-org/stripes-force @folio-org/fe-tl-reviewers, what do you think about the [comment above](https://github.com/folio-org/stripes-core/pull/1704#issuecomment-4183026219) and PR itself?

zburke:
@usavkov-epam, what I think about that comment above is: there is no such thing as "user is logged into two tenants in separate tabs". This is why I have struggled with this PR from the start. When multiple tabs are open and you switch affiliations in a single tab, only that tab reloads, but in fact your affiliation has silently changed across the other tabs (because shared cookie and localStorage values have changed) without notifying you and without reloading. This is a bug; all tabs must reload when the affiliation changes. 

I understand the desire to have the 'Log in again' button on the `/logout-timeout` route return users to the page and tenant they were most-recently using. This is a legitimate use-case/expectation. That said, such an implementation (i.e. this PR) should be evaluated against a properly-implemented switch-affiliation feature, and this is not yet possible. 

The changes here in the already-serpentine `loginServices.js` functions make them even more complicated. Choosing to ignore the multi-tab issue feels like trading one bug for another rather than actually squashing a bug. In light of all this, my strong preference would be to block this work on hold until the affiliation-switching work is completed. Christine and Khalilah agreed to write the ticket describing the multi-tab affiliation-switching feature. Please consult with them about its priority and timeline, and likewise the priority of this PR. If they agree this ticket has high priority and needs to be handled soon, I am willing to reconsider, but it is my very strong preference that we should address the multi-tab affiliation-switching bug first.

usavkov-epam:
@zburke, could you please take another look at this PR?

PR #1705, opened 2026-04-03
STRIPES-901 - validate for conflicting peer dependencies.

JohnC-80:
## [STRIPES-901](https://folio-org.atlassian.net/browse/STRIPES-901)

Currently, we strip out shared singleton dependencies from ui-module builds with the [`import: false` configuration in `stripes-webpack`](https://github.com/folio-org/stripes-webpack/blob/main/module-federation-config.js#L53) so that there's no chance they'll be loaded in the bundle.
They're all mapped (forced) to use the host app's loaded dependency, so they don't make the request for a conflicting version (as they would if `import: false` were not used. Without that request, Module-federation's built-in dependency check/warn won't fire - and so we have to implement our own via this very PR!

Despite being missing from the bundled code, the version that the ui-module was built with is maintained within its generated `mf-manifest.json` file. The code here loads that file for each remote, inspects the JSON and checks for compatibility on the versions of shared dependencies. If an incompatibility is detected, a warning is emitted in the console.

 This check does *not block the rest of the modules from loading, nor does a module failing the check.
 
 This PR does include the addition of the `semver` library to `stripes-core`.
 
The console warning:
image

Example `mf-manifest.json` from compatible `react` dep:
```
  "shared": [
    {
      "id": "folio_users:react",
      "name": "react",
      "version": "18.3.1", // version ui-module is built with...
      "singleton": true,
      "requiredVersion": "~18.3",  // @folio/stripes-core's version...
      "assets": {
        "js": {
          "async": [],
          "sync": []
        },
        "css": {
          "async": [],
          "sync": []
        }
      },
      "fallback": ""
    },
```

Offending app `mf-manifest.json`:
```
  "shared": [
    {
      "id": "folio_minimal_r_19:react",
      "name": "react",
      "version": "19.2.4",  // version ui-module is built with...
      "singleton": true,
      "requiredVersion": "~18.3",  // @folio/stripes-core's version...
      "assets": {
        "js": {
          "async": [],
          "sync": []
        },
        "css": {
          "async": [],
          "sync": []
        }
      },
      "fallback": ""
    }
  ],
```

### DIFF
--- a/src/components/EntitlementLoader.js
+++ b/src/components/EntitlementLoader.js
@@ -6,6 +6,7 @@ import { useStripes } from '../StripesContext';
 import { useCallout } from '../CalloutContext';
 import { ModulesContext, useModules, modulesInitialState } from '../ModulesContext';
 import { loadEntitlement } from './loadEntitlement';
+import { validateRemoteDependencies } from './remoteDependencyValidation';
 
 // class for carrying formatted callout error messages.
 class RemoteModuleLoadingError extends Error {
@@ -212,11 +213,14 @@ const EntitlementLoader = ({ children }) => {
         }
 
         let cachedModules = modulesInitialState;
+        const validatedRemotes = remotes || [];
         const remotesWithLoadedAssets = [];
         const loadFailures = [];
 
         // if the signal is aborted, avoid all subsequent fetches, state updates...
         if (!signal.aborted) {
+          validateRemoteDependencies(validatedRemotes, signal);
+
           // load module assets (translations, icons)...
           const assetResults = await loadAllModuleAssets(stripes, remotes);
           assetResults.forEach((r, i) => {
--- a/src/components/EntitlementLoader.js
+++ b/src/components/EntitlementLoader.js
@@ -213,13 +213,12 @@ const EntitlementLoader = ({ children }) => {
         }
 
         let cachedModules = modulesInitialState;
-        const validatedRemotes = remotes || [];
         const remotesWithLoadedAssets = [];
         const loadFailures = [];
 
         // if the signal is aborted, avoid all subsequent fetches, state updates...
         if (!signal.aborted) {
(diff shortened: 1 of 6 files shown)


src/components/EntitlementLoader.js
```diff
@@ -212,11 +213,14 @@ const EntitlementLoader = ({ children }) => {
...
 
         let cachedModules = modulesInitialState;
+        const validatedRemotes = remotes || [];
         const remotesWithLoadedAssets = [];
         const loadFailures = [];
 
         // if the signal is aborted, avoid all subsequent fetches, state updates...
         if (!signal.aborted) {
+          validateRemoteDependencies(validatedRemotes, signal);
+
```
zburke: The naming is weird here. How can we set `validatedRemotes` to `remotes` on L216 if we don't call `validateRemoteDependencies()` on that list until L222?

src/components/remoteDependencyValidation.js
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  if (!satisfies(version, requiredVersion, { includePrerelease: true })) {
+    return formatDependencyMismatch(remoteName, pkgName, version, requiredVersion);
+  }
+
+  return null;
+};
+
+export const validateRemoteDependencies = async (remotes = [], signal) => {
```
zburke: "validate" is a tough prefix because it doesn't provide a clear indication how this function will behave. Still, I expected it do something more than logging the incompatibilities it found: either return booleans, or to return null or throw.

Maybe it should be called `logRemoteDependencyViolations`?

zburke (approved):
The internal naming here confused me, but in the end it does what the PR describes: log a console warning when a conflict is detected. 

I'd love to see function headers for each of the function in remoteDependencyValidation.js describing the arguments and purpose. 

I agree with sonar about the order of arguments with default values.

PR #1710, opened 2026-04-24
STCOR-1048 correctly handle unreliable AT metadata during RTR

zburke:
## Summary 

If the AT cookie and its corresponding expiration data in the session object in local storage get out of sync (e.g. somebody deletes cookies), Stripes will be in an inconsistent state: its AT is missing, but its session data indicates that a valid AT is present. This PR addresses that inconsistency, allowing RTR to recover and proceed with rotation. 

## Approach

Favor simplicity over cleverness: we know the token's metadata can be unreliable, so we don't rely on it. If a cookie was deleted prior to expiration but metadata in storage is never updated, there's nothing we can do to detect that. In that situation, the previous algorithm worked like this: all requests through the queue would check storage, find it valid, replay, and return the response without inspecting for success, allowing 401s to bubble to their origin (whoops). This PR simplifies and corrects the logic: all requests through the queue attempt replay and all responses are inspected. The first one will fail, rotate, and replay. Subsequent requests replay and (assuming they are successful) short-circuit.

This means the first request in the queue pays a performance penalty (an extra failed replay attempt) but every other request in the queue goes through more efficiently (no async storage check). It also simplifies the logic because we don't spend any effort trying to figure out "Do we think this request is likely to succeed because we have an unverifiable belief that our token is valid?" We just make requests and deal with the consequences.

This PR also addresses a wart with the earlier implementation regarding the handling of non-ok responses that entered the rotation queue but were determined not to have failed due to a missing AT (e.g. a 404, 403, etc). Now, if rotation is determined to be unnecessary, `rotateAndReplay()` returns the response instead of throwing it. Not rotating is not an exceptional condition, it's just ... not rotating. Back in `FFetch`, we would check for this and return the response from the `catch` clause. By not throwing this non-error, `FFetch` can be simplified too, omitting the now-unnecessary `catch` that would do nothing except rethrowing.

Refs [STCOR-1048](https://folio-org.atlassian.net/browse/STCOR-1048)

### DIFF
--- a/src/components/Root/rotateAndReplay.js
+++ b/src/components/Root/rotateAndReplay.js
@@ -6,34 +6,48 @@ export const RTR_LOCK_KEY = '@folio/stripes-core::RTR_LOCK_KEY';
 
 /**
  * rotateAndReplay
- * rotateAndReplay is called when a response has a 4xx code and if it looks
- * like an authentication problem, rotate the tokens, replay the original
- * request, and resolve with that response. If rotation does not happen (there
- * are many possible reasons, detailed below), reject with the original response.
- * If there are any errors related to rotation itself, report them via a callback
- * and then reject with the original response.
  *
- * 1. inspect the response and the config to determine if rotation is needed.
+ * IN SUMMARY
+ *
+ * rotateAndReplay is called when a request in FFetch returns a non-ok response.
+ * Here, we inspect the request-options and response in detail and if it looks
+ * like the problem was an expired (and therefore absent) AT cookie, rotate the
+ * tokens and then replay the original request, returning the new response. If
+ * the initial inspection fails (maybe we got a 404, which is a problem of
+ * course but not an RTR problem), reject with the original response so it can
+ * bubble back to the caller. If rotation itself fails, report the failure via
+ * the config callback and then reject with the original response (if
+ * available) or the rotation error.
+ *
+ * IN DETAIL
+ *
+ * 1. 🧐 inspect the response and the config to determine if rotation is needed.
  *    reject if any of the following are true:
  *    * we have a response but its code is not one we handle
  *    * the original request was configured with an `ignoreRtr` option
  *    * the config's shouldRotate() function returned false
- * 2. get an exclusive, browser-level lock, to avoid multiple tabs/windows
+ * 2. 🔒 get an exclusive, browser-level lock, to avoid multiple tabs/windows
  *    attempting to rotate simultaneously. the lock takes an async function
- *    and releases when the promise settles. within that promise, rotate.
+ *    and releases when the promise settles. within the lock's promise, rotate.
  *    if the rotation succeeds, replay the request and return the response.
  *
- * Inside the lock, the rotation process looks like this:
- * 1. Check storage to see if the token is still expired. When several requests
- *    fire together and then fail together, they'll queue up due to the browser
- *    lock. Once the first promise resolves, we can use its tokens for the
- *    remaining queued requests. This means that failed requests become
- *    single-threaded since they're all stuck here, waiting for the lock. Alas.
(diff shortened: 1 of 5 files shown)


JohnC-80 (approved):

src/components/Root/rotateAndReplay.js
```diff
@@ -77,15 +90,18 @@ export const rotateAndReplay = async (fetchfx, config, error) => {
...
-      // 1. if rotation completed elsewhere, we don't need to rotate!
-      // maybe another tab rotated, maybe it happened while awaiting the lock.
-      if (await config.isValidToken()) {
-        config.logger.log('rtr', 'reusing token supplied by another request');
-        return replayRequest();
+      // 1. 👀 If rotation completed elsewhere, we don't need to rotate!
+      // Replay the request and inspect the response. If it's good, we're good.
+      // If it looks like an authentication failure, however, proceed to rotate.
+      config.logger.log('rtr', 'reusing token supplied by another request');
+      const replayResponse = await replayRequest();
```
JohnC-80, line 101: Since we're potentially going to have to make this request an additional time, do we need to clone it here as we do in FFetch?

JohnC-80 (approved):

src/components/Root/rotateAndReplay.js
zburke: @JohnC-80, This is a good question. My first reaction was, "oooh, you're correct, The Right Way is to reevaluate this new response the same way we did the original one, calling `shouldRotate()` (in order to clone-and-inspect as you suggest) AND checking the status code as we do here. 

But now that I've thought about it for a bit, I think the logic here can be simpler. We already know the original request failed AND that it passed the criteria for needing rotation. What we are testing here is "Did the replayed response succeed?", not "Does this replayed response need rotation?" 

So, yes, this code needs to change, but in the direction of being even simpler (test for success) rather than more elaborate (test for different kinds of failure that indicate a need for rotation): 
```
if (replayResponse?.ok)
  return replayResponse;
}
```
Agreed?

JohnC-80: What are we really checking for in the lock? Successful request or successful rotation?
I know I'm back and forth on things, and I apologize - I'm still stuck thinking we should still try for fewer requests going out. In my mind the extra request (and its clones) add up. 
I think we should bite the bullet and see about conforming FXHR so that fewer failed attempts occur.
Current flow:
```
Request w/o rotation
   |
  Check for failure
      |
     Queue Rotation
        |
        Clone, Replay request
          |
          Check for failure (and thus, successful rotation)
            |
            Rotate if no successful rotation found.
              |
              Replay request (again) --> then, post-success, continue to next request that asked for rotation.
```

Seems simpler, less chatty - using a sentinel for rotation.
Previously, I mentioned removing the expiry - not what's best and removes isolation from this system.
```
Request w/o rotation
  |
  Check for failure
    |
    (failure) setRotationSentinel(false) and queue locked rotation.
      |
      Check rotation sentinel
        |               |
        Replay request  |
                        |
                      Rotate
                          |
                          setRotationSentinel(true)
                              |
                              Replay request --> then, post-success, continue to next request that asked for rotation, it sees the sentinel, returns its response...
```

zburke: > I think we should bite the bullet and see about conforming FXHR so that fewer failed attempts occur.

Filed [STCOR-1055](https://folio-org.atlassian.net/browse/STCOR-1055) to improve FXHR's handling of expired tokens

JohnC-80: Resolving to leave for future work, changes out of scope for this since we just want to achieve the goal of preventing 401's to bubble up here.

JohnC-80 (approved):
Yay! The experience feels much more solid with this and #1713 in place! 🎉 🎉

PR #1714, opened 2026-04-29
STCOR-1055 adopt FFetch's rotation logic in FXHR

zburke:
Use similar patterns in these similar classes, i.e. attempt to make a request, trap unsuccessful responses, rotate, and replay. We abandoned checking the (unreliable) AT expiration data in FFetch in PR #1710; here we do the same for FXHR, capturing `error` events and inspecting them to divine the reason for the failure.

There is not a lot to go on in an XHR request; all we can really see is how much data has been transfered and how much data we expect to transfer. Here, we interpret "no data sent" as "I bet we couldn't send data because we didn't have a valid token" so we rotate and try again. We only try rotation once. Errors received after a second attempt or after any data has been transfered are passed straight up to the superclass.

This PR also resolves a minor wart on the earlier implementation, replacing the async/await based-logic with a Promise-based implementation. Because `send()` and `onerror()` are void, we cannot leverage `await`. Instead, we just have to do our work inside an un-returned promise chain.

NB: I couldn't figure out how to write a test that could validate `send()` being called twice (the original invocation, plus the replayed one after the error is trapped). Logging bears out that it works correctly. Why doesn't the spy work?

Refs [STCOR-1055](https://folio-org.atlassian.net/browse/STCOR-1055)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 * Export AuthenticatedError component. Refs STCOR-1047.
 * Add case for `users-keycloak`'s 404 status to rotate keys on failure. STCOR-1054.
 * Test-logic only belongs in tests, Refs STCOR-1050.
+* Adopt FFetch's rotation logic in FXHR for consistency. Refs STCOR-1055.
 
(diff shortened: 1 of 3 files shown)


zburke:
Replaced by #1723

PR #1718, opened 2026-04-30
STCOR-1060 - clone response again prior to attempting a request before rotation.

JohnC-80:
This is a follow-up to PR https://github.com/folio-org/stripes-core/pull/1710, addressing an issue that slipped through the cracks there: when a request is a literal Request object, it can only be consumed once, but the rotate-and-replay process will replay a request twice (once to test the short-circuit in case rotation already completed elsewhere, and second time after rotation if the short-circuit failed).

This PR addresses the problem by always cloning Request objects inside replayRequest(). This means requests that require rotation will be cloned an extra time since the second time they are replayed is the final time they are replayed, but this feels like an acceptable price to pay for keeping the logic simple.

Refs [STCOR-1060](https://folio-org.atlassian.net/browse/STCOR-1060)

### DIFF
--- a/src/components/Root/FFetch.js
+++ b/src/components/Root/FFetch.js
@@ -179,7 +179,7 @@ export class FFetch {
       // or a Request object (which can only be consumed once, and needs
       // to be cloned before it is used the first time in case this fetch
       // triggers rotation and it needs to be replayed.
-      const reusableResource = resource instanceof Request ? resource.clone() : resource;
+      const reusableResource = resource instanceof Request || resource instanceof URL ? resource.clone() : resource;
 
       // readers/writer lock pattern: don't fetch while rotation is in-progress
       // https://developer.mozilla.org/en-US/docs/Web/API/LockManager/request
--- a/src/components/Root/rotateAndReplay.js
+++ b/src/components/Root/rotateAndReplay.js
@@ -5,6 +5,34 @@ import { RTRError } from './Errors';
 export const RTR_LOCK_KEY = '@folio/stripes-core::RTR_LOCK_KEY';
 
 /**
+ * replayRequest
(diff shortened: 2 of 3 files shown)


src/components/Root/FFetch.js
```diff
@@ -179,7 +179,7 @@ export class FFetch {
       // or a Request object (which can only be consumed once, and needs
       // to be cloned before it is used the first time in case this fetch
       // triggers rotation and it needs to be replayed.
-      const reusableResource = resource instanceof Request ? resource.clone() : resource;
+      const reusableResource = resource instanceof Request || resource instanceof URL ? resource.clone() : resource;
```
zburke, line 183: I know I handed you this line, but I was wrong about it; `URL` does not implement `clone()`. Sorry.

```suggestion
      const reusableResource = resource instanceof Request ? resource.clone() : resource;
```

src/components/Root/rotateAndReplay.js
```diff
@@ -69,10 +69,10 @@ export const rotateAndReplay = async (fetchfx, config, error) => {
...
-  const replayRequest = async () => {
-    if (error?.resource) {
-      config.logger.log('rtr', 'replaying ...', error.resource);
-      const response = await fetchfx.apply(globalThis, [error.resource, config.options(error.options)]);
+  const replayRequest = async (res) => {
+    if (res) {
+      config.logger.log('rtr', 'replaying ...', res);
+      const response = await fetchfx.apply(globalThis, [res, config.options(error.options)]);
       return response;
     }
```
zburke: How about we just pull `replayRequest` up and pass all its params? It feels a little weird to have some function arguments and some references to things defined in the outer scoped.

zburke (approved):

PR #1719, opened 2026-04-30
rotation logic is consolidated and easier to follow, maybe?

zburke:


zburke:
Nope. Rotation logic is just different, not any better. Moving the problem is not the same as solving it.

PR #1724, opened 2026-05-05
STCOR-1061 consolidate logout functionality into <Logout>

zburke:
This looks messy, but it's actually a clean up job:

1. If Stripes initialization fails, redirect to `/logout` and allow the `<Logout>` component to take care of cleaning up. 
2. Consolidate all logout functionality into a hook called from `<Logout>`. Don't invoke logout functions from anywhere else; instead, redirect to `/logout`. 
3. If the Stripes and Keycloak sessions go out of sync (e.g. a user removes their cookies), redirect to Keycloak to sync the sessions, then automatically logout to terminate them.

The most important thing changing here is the first: the error-handling function passed from `checkOkapiSession()` to `validateUser()` no longer calls `logout()`; instead it redirects to `/logout`, allowing all logout-related functionality to be consolidate there. `checkOkapiSession()` is called via a lifecycle method on `Root.js`. Starting the logout process there, and then picking it up a second time when `<Logout>` renders, was the RCA behind the terrible behavior described in this ticket. Thus, this is very much a follow up to PRs #1713 and #1720, addressing the unexpected session termination that could occur when opening a new tab, initializing a new instance of Stripes, which will fail if the `_self` API query fails due to an expired cookie. Those PRs improve handling of the API query; this one improves error handling if something still goes awry.

Additionally, this PR moves `<Logout>` up in the component hierarchy, pulling it out from  under `<MainContainer>` for authenticated users. The `/logout` route needs to be accessible both to authenticated users (so they can use it to logout) and to unauthenticated users (to indicate to previously-authenticated users that the logout process was successful). When it was inside `<MainContainer>`, untrapped errors during sign-out caused `<Logout>` to display within the main container, a jarring glitch. Both problems are now resolved: we trap the errors AND `<Logout>` is rendered above/alternatively to `<MainContainer>`, eliminating the possibility of showing "you have logged out" inside the application navigation that clearly demonstrates you are still logged in.

Refs [STCOR-1061](https://folio-org.atlassian.net/browse/STCOR-1061)


JohnC-80 (approved):
The 'actual' logout flow for a failed logout call gets head-spinny! But it's probably the best that we can do with what we've got to resync keycloak and stripes.

zburke:
@Dmytro-Melnyshyn, your feedback has been crucial to the correct implementation of earlier RTR work. Do you mind taking a look at this PR? Although not strictly RTR, this PR aims to resolve the "unexpected logout" problem I thought I was solving in PR #1696, so it feels related. Thank you!

src/components/Logout/Logout.test.js
```diff
@@ -33,16 +40,24 @@ describe('Logout', () => {
...
+      const mutate = jest.fn();
+      mockUseLogoutMutation.mockReturnValue({ mutate });
+
+      act(() => {
+        render(<Logout sessionTimeoutTimer={{ clear: jest.fn() }} sessionTimeoutWarningTimer={{ clear: jest.fn() }} />);
+      });
+      waitFor(() => {
+        expect(mockUseLogoutMutation.mutate).toHaveBeenCalled();
+        screen.getByText('LoadingView');
+      });
```
Dmytro-Melnyshyn: `waitFor` should always be used together with `await`, otherwise the test will pass.

src/components/OIDCRedirect.js
```diff
@@ -41,9 +53,13 @@ const OIDCRedirect = () => {
 
   const getUrl = () => {
     if (stripes.okapi.authnUrl) {
-      if (unauthorizedPath) {
+      if (unauthorizedPathRef.current) {
         removeUnauthorizedPathFromSession();
-        return unauthorizedPath;
+
+        if (unauthorizedPathRef.current === AUTOMATIC_LOGOUT_LOCATION) {
+          return 'logout';
```
Dmytro-Melnyshyn: `<Redirect to="logout">` from `/oidc-landing` resolves to `/logout` by luck (it replaces the last path segment). This works in the current route layout, but is fragile. If `OIDCRedirect` were ever nested deeper (e.g., `/auth/oidc-landing`), the redirect would produce `/auth/logout` instead.

src/loginServices.js
```diff
@@ -1150,9 +1056,9 @@ export async function validateUser(okapiUrl, store, tenant, session, handleError
  * @param {redux store} store
  * @param {string} tenant
  */
-export async function checkOkapiSession(okapiUrl, store, tenant) {
+export async function checkOkapiSession(okapiUrl, store, tenant, history) {
```
Dmytro-Melnyshyn, line 1063: JSDoc is missing the `history` parameter.

src/components/Logout/useLogoutMutation.js
```diff
@@ -0,0 +1,165 @@
...
+ * session timers) that is unique per-tab and therefore must be cleared
+ * in each instance of stripes, i.e. in each tab.
+ *
+ * @param {object} redux store
+ * @param {object} queryClient react-query client
+ * @param {array} timers FLST and IST timers
+ *
+ * @returns {Promise}
+ */
+export async function clearSessionStorage(store, queryClient, timers) {
```
Dmytro-Melnyshyn: Let's rename it to `clearSharedStorage` or `clearSession`.

src/components/Logout/Logout.js
```diff
@@ -15,75 +14,59 @@ import {
...
 /**
  * Logout
- * Make an API call to /logout to terminate the session server-side and render
- * a "you've logged out" message. This component can be invoked from different
- * UI routes and will display a different message depending on that route:
- *
- * /logout: a user chose to end the session
- * /logout-timeout: stripes redirects to this location due to idle-session timeout
- *    the "reason" key in the query string provides additional details
+ * call useLogoutQuery() to terminate the session server-side, remove cookies,
```
Dmytro-Melnyshyn:
```suggestion
 * call useLogoutMutation() to terminate the session server-side, remove cookies,
```

Dmytro-Melnyshyn (approved):
Please fix the "is defined but never used" before merging.

zburke:
> Please fix the "is defined but never used" before merging.

I ran Sonar but not Lint 🤦

PR #1726, opened 2026-05-11
STCOR-1064 Add cleanup methods to FFetch, use them in Root's unmount.

JohnC-80:
## [STCOR-1064](https://folio-org.atlassian.net/browse/STCOR-1064)
Having StrictMode turned on will cause the existing RTR implementation to hang in the init process since multiple fetch calls courtesy of `checkOkapiSession` will be made, and `<Root>` will unmount and remount before any rotation even occurs. Speculatively, this is causing the first `exclusive` rotation request to suspend indefinitely, so it never unlocks and the UI is frozen.

The solution - Abort the fetch when `<Root>` unmounts during the strict-mode double-lifecycle. This will kick out the request before rotation gets started so once normal functionality resumes the exclusive lock can be unclogged.

### DIFF
--- a/src/components/Root/Root.js
+++ b/src/components/Root/Root.js
@@ -147,6 +147,14 @@ class Root extends Component {
     return !this.withOkapi || nextProps.okapiReady || nextProps.serverDown;
   }
 
+  componentWillUnmount() {
+    // clean up timers and pending requests on unmount
+    this.ffetch.destroy('Root component unmounted, aborting pending requests');
+    this.ffetch.restoreFetch();
+    this.ffetch.restoreXMLHttpRequest();
+    this.ffetch = null;
+  }
+
   updateQueryResourceStateKey = () => {
     const { modules, history } = this.props;
     const appModule = getCurrentModule(modules, history.location);
--- a/src/components/Root/Root.js
+++ b/src/components/Root/Root.js
@@ -148,11 +148,11 @@ class Root extends Component {
   }
 
   componentWillUnmount() {
-    // clean up timers and pending requests on unmount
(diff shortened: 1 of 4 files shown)


src/components/Root/Root.js
```diff
@@ -147,6 +147,14 @@ class Root extends Component {
     return !this.withOkapi || nextProps.okapiReady || nextProps.serverDown;
   }
 
+  componentWillUnmount() {
+    // clean up timers and pending requests on unmount
```
zburke:
```suggestion
    this.sessionTimeoutTimer.clear();
    this.sessionTimeoutWarningTimer.clear();
```

zburke (commented):
~Contrary to [React's [deprecated] documentation](https://legacy.reactjs.org/docs/strict-mode.html#detecting-unexpected-side-effects) that `<StrictMode>` for Class components will double-invoke "constructor, render, and shouldComponentUpdate", logging shows `Root::constructor` being called only once, followed by cDM, cWU, cDM. At the same time, logging in `ffetch` shows it _is_ being called after that cycle, even though it shouldn't be, since cWU sets it to null and cDM does not restore it.~

~So, reading documentation suggests your PR should work as-is, and playing your PR suggests it does. But logging suggests the only things we should clean up in cWU are those that we set up in cDM, **not those we set up in the constructor, which means we don't need changes here in cWU, and in fact they should be harmful, but they aren't :/~

~I'm so confused. Can you explain at all?~

Further review shows this working as-expected according to documentation. That means we should add clean up of the ResetTimers in cWU. Everything else looks good as is. Thanks for this. It's excellent clean up!

zburke (approved):

PR #1727, opened 2026-05-12
STCOR-717 Address Sonar's Software Quality > Reliability complaints

zburke:
There's nothing of real interest here; most changes are either completely inconsequential (e.g. replacing `throw Error()` with `throw new Error()` or `string.replace(/.../g)` with `string.replaceAll(/.../g)`) or generally obvious (e.g. replacing `list.sort()` with `list.sort(comparatorFunction)`), all to appease Sonar's "Software Quality > Reliability" complaints.

The only potentially interesting thing is changing `async cDM()` to `cDM()` in `Root.js`, but even this is not really interesting, since all it does is restore the correct reporting of `cDM`'s signature: it is not async. I wonder what idiot added that. Or, right, the usual idiot in PR #1696. 🤦 👊 

Refs [STCOR-717](https://folio-org.atlassian.net/browse/STCOR-717)

### DIFF
--- a/src/AppRoutes.js
+++ b/src/AppRoutes.js
@@ -30,7 +30,7 @@ const AppRoutes = ({ modules, stripes }) => {
         ModuleComponent = connect(module.getModule());
       } catch (error) {
         console.error(error); // eslint-disable-line
-        throw Error(error);
+        throw new Error(error);
       }
 
       const moduleStripes = stripes.clone({ connect });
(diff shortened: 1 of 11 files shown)


JohnC-80 (approved):
Thanks! It's all one part 'fine, dad' the other part 'yeah, this probably helps'... All around, it's good to address! I wonder if it will ever come to recommend array rest/spread over Array.push() for adding items...

PR #1728, opened 2026-05-13
STCOR-1002 handle Error input in OIDCLandingError

zburke:
`error` may contain structured JSON from a failed API response, or a simple string, or an Error object if the API request itself fails. Handle them all. Previously, Error-object props caused a failure because the object was simply supplied as a child in render, which React does not support.

Refs [STCOR-1002](https://folio-org.atlassian.net/browse/STCOR-1002)


JohnC-80 (approved):
I've seen this happening! But had no idea, room to sus where it was actually coming from or what to write for the solution. ❤️

PR #1732, opened 2026-05-27
STCOR-1069 - Check AT pre-emptively on fetches to avoid lots of 401's

JohnC-80:
## [STCOR-1069](https://folio-org.atlassian.net/browse/STCOR-1069)

Given that we've got good expiry information stored, we'll leverage it to reduce the number of initial 401's.

This approach reduces the pile of initial 401's from the `shared` lock by circumventing their initial request and putting it off until rotation has happened.

Approach:
Check atExpiry before each initial fetch. 
If the token is expired, set an `atExpired` flag and pass it to `rotateAndReplay` as part of the `original` object.
Rotate if `original.atExpired` is true, rather than inspecting the original response (circumvent the `shouldRotate()` call.)

FXHR: left as-is... there's nothing async about XHR - if we DO want this check prior to, then we have to move the expiry from indexDB to localStorage and check that directly - our wrapping utility in `loginServices` was well-intended, but not usable in this case.


src/components/Root/rotateAndReplay.js
```diff
@@ -156,7 +156,7 @@ export const rotateAndReplay = async (fetchfx, config, original) => {
   // 🧐 are we certain we need to rotate? check these two conditions:
```
zburke, line 156: "check these three conditions"

src/components/Root/rotateAndReplay.js
```diff
@@ -156,7 +156,7 @@ export const rotateAndReplay = async (fetchfx, config, original) => {
...
   // 1. shouldRotate() asked for rotation
   // 2. the response status indicates we need rotation
-  if (await shouldRotate(original?.response) || statusCodes.includes(original?.response?.status)) {
+  if (original.atExpired || await shouldRotate(original?.response) || statusCodes.includes(original?.response?.status)) {
```
zburke: Or do we think `shouldRotate()` should be refactored to consume `original` instead of `original.response` and we can consolidate all this "inspect and decide" logic in one place? It's fine to say, "Interesting idea, but that should be a separate piece of work."

zburke: `original?.atExpired` ... assuming that at some point we _can_ get sync access to  expiration data and thus want to supply this value from an XHR.

zburke (commented):
`shouldRotate()` is async because the present source of token-expiration data is IndexedDB/localforage. Moving to a sync store is sensible (for many reasons we have discussed), but potentially a larger piece of work to bite off. Do you think it's worth it? Do you think it would fit in the scope of this PR?

JohnC-80:
@zburke Thank you for the feedback. Going to see if I can simplify on these conditionals... We'll see if we can save the conversion to localStorage for another PR or if it overwhelmingly makes sense to just do it here.

zburke (approved):
I'm happy with the way this came together. I worried it would lead to a bunch of messy conditionals, but in fact the clear-labeling (via `rtrIgnore`) of requests to publicly-accessible endpoints feels like a nice cleanup, causing us to be explicit about our expectations. 🤝

PR #1737, opened 2026-06-09
STCOR-1069 proactive rotation skips initial play

zburke:
## setup

Requests can enter `rotateAndReplay()` for two reasons:
1. inspecting storage shows expired tokens, so we are rotating preventively
2. response inspection suggests expired tokens, so we are rotating reactively

Prior to this PR, the steps in `rotateAndReplay()` were roughly these:
1. attempt replay of the request in case tokens have been refreshed
2. rotate
3. replay

## approach

This PR addresses a logic flaw in step 1: When a request invokes preventive rotation, there should not be a pre-rotation replay. Preventive rotation amounts to "I'm invoking rotation due to a current problem with storage". As long as that problem persists, there is no value in replay because we would expect replay to fail. Reactive rotation amounts to "I'm invoking rotation due to past problem with storage". In this situation, it's worth inspecting storage and replaying if the problem appears to have cleared. Of course, it's possible storage appears to be ok but is actually out of sync, in which case the replay will fail, and we will proceed with rotation.

The new function, `shouldPreRotate`, enables preventive rotation. 

## gory details/additional work

* `usePasswordRules.js` retrieve `ky` from a common hook
* `PasswordValidationField.js` trivial lint cleanup
* `useOkapiKy.js::usePublicGatewayKy()` new hook to supply `rtrIgnore` to native fetch; previously, ky-based requests with this property were not actually bypassing RTR; instead, they were triggering RTR and then succeeding during the (now-removed) pre-rotation replay (whoops)
* `loginServices.js::getSSOEnabled()` supply missing `rtrIgnore` attribute
* `withOkapiKy.js` retrieve `ky` from a common hook
* `PasswordValidationField.js` remove this failing test since the component is orphaned; at some point, we should remove the component too
* `profile-dropdown-test.js` must supply `userLoggedIn` to cause storage to be properly populated so that when a fetch is triggered, RTR's new preventive-rotation inspector finds data in storage to inspect

<details>
<summary>even gorier details and historical context</summary>
RTR can happen proactively (because we know tokens are expired) and reactively (because the response from a fetch indicates that credentials were missing, and presumably expired). We already handle reactive rotation by inspecting responses, i.e. the fetch-rotate-replay flow. This PR adds better handling for proactive rotation by inspecting token expiration data in storage before a fetch, allowing for a simpler rotate-fetch flow when we know (or at least expect) that the initial fetch will fail due to missing tokens.

In the first implementation of the RTR rewrite ([STCOR-1012](https://folio-org.atlassian.net/browse/STCOR-1012)/PR #1696), we would capture failures, rotate, and replay. As failed requests proceeded single-file through the rotation-lock, we would attempt to replay each one before proceeding to rotation. Thus, the first request through the lock would actually be made three times:
1. as originally made [failing]
5. in the lock, attempting replay before rotation [failing]
6. in the lock, final replay after rotation [succeeding]

whereas subsequent requests through the lock would only be made twice:
1. as originally made [failing]
2. in the lock, attempting replay before rotation [succeeding]

This was good but not great; the first request happens three times, and the browser console shows a wall of request failures.

Next, the original [STCOR-1069](https://folio-org.atlassian.net/browse/STCOR-1069)/PR #1732 implementation improved the algorithm, inspecting the tokens in storage in order to skip straight from FFetch to rotateAndReplay when the tokens were expired. This popped the first failed requests off each stack above, reducing (but not eliminating) the wall of failed requests in the browser console, because it left in place the pre-rotation "replay" failure of the first request to proceed through the rotation lock.

Finally, we are here with a follow-up STCOR-1069 PR.  executes the same pre-request check in `rotateAndReplay()` that we execute pre-request in `FFetch()`, fully cleaning up the console by skipping any requests that are guaranteed to fail. This work revealed that `rtrIgnore` was not being properly threaded through to fetch, and the requests were in fact proceeding through `rotateAndReplay()` and being completed during pre-rotation replay. This PR al
</details>

Refs [STCOR-1069](https://folio-org.atlassian.net/browse/STCOR-1069)


zburke:
@JohnC-80 , IDK why Karma is choking in CI. Locally, it's clean:
```
Finished in 5.856 secs / 2.128 secs @ 17:56:30 GMT-0400 (Eastern Daylight Time)

SUMMARY:
✔ 25 tests completed
```

src/useOkapiKy.js
```diff
@@ -1,26 +1,79 @@
...
+
+export const useUnauthenticatedOkapiKy = ({ tenant, timeout } = {}) => {
+  // Kong has a default timeout of 60 seconds
+  const {
+    locale = 'en',
+    timeout: defaultTimeout = 60000,
+    tenant: currentTenant,
+    url,
+  } = useStripes().okapi;
+  console.log('asdf')
```
JohnC-80: remove comment...

src/useOkapiKy.js
```diff
@@ -1,26 +1,79 @@
...
+  return ky.create({
+    ...defaultOptions({ locale, tenant, currentTenant, timeout, defaultTimeout, url }),
+    fetch: async (resource, options) => {
```
JohnC-80: This already returns a promise by returning the `fetch(...)` call - I don't think this has to be wrapped in async/await. If it further fussed with the data, with a `json()` call or smth would be a different story....

JohnC-80: `ky@0.23.0` itself does not pass `options` from the `ky.[method]()` function as a parameter to this function... despite their docs saying how 'this function should match the fetchAPI - The implementation there doesn't :/. https://github.com/sindresorhus/ky/blob/v0.23.0/index.js#L180

However, future versions of ky DO pass options - so it will be used whenever we eventually do upgrade... a comment would probably be nice for other eyes and our future eyes that might be looking at this code and expecting their `options` to come through.

src/components/CreateResetPassword/components/PasswordRequirementsList/queries/usePasswordRules.js
```diff
@@ -8,6 +8,7 @@ const usePasswordRules = (rulesLimit) => {
   const { locale = 'en', tenant, url } = useStripes().okapi;
 
   const kyInstance = ky.create({
+    fetch: globalThis.unauthenticatedFetch,
```
JohnC-80: I feel like we should trade this in for `(req) => fetch(req, {rtrIgnore: true}` for sake of consistency and just scrap the extra thing created by FFetch on globalThis.

JohnC-80 (changes requested):

JohnC-80 (approved):
This actually did clean up in a couple places. Thanks for the work!!

PR #1738, opened 2026-06-12
STCOR-1067 reliably sync browser, FOLIO, and Keycloak sessions on logout

zburke:
Reliably terminate both the FOLIO and Keycloak sessions on logout and when there is an error during OTP-exchange on login.

Three things need to be in sync for a session to function correctly:
* FOLIO cookies (used to make API requests)
* Keycloak cookies (used to track authentication)
* Browser storage (use to track user state)

Stripes only has access to browser storage and has to intuit whether the cookies are present based on API responses. Effectively, data in storage is a signal to Stripes that cookies are present, and therefore that API requests should succeed. During logout, if storage is populated but API requests fail, Stripes interprets this as "sessions are out of sync" and will redirect to Keycloak to guarantee that its session gets terminated.

Stripes makes the same assumption, and handles things the same way, during login if OTP exchange fails, e.g. if Keycloak is available but the API gateway is down. In this situation, the user will be able to authenticate and start a Keycloak session but will not be able to start a FOLIO session. Previously, Stripes simply displayed an error message in this situation; now, Stripes recognizes the OTP exchange failure as a sync error and redirects to Keycloak accordingly to terminate that session.

The most important changes here are in useLogoutMutation, now structured like this:
```
try {
  API call to /logout
}
catch {
  if error, redirect to Keycloak
}
finally {
  clear browser storage
}
```
When storage is populated and the `/authn/logout` API request fails, Stripes treats this as a session-sync error and redirects to Keycloak to force that session to terminate. If storage is empty, as it will be on a secondary tab after the primary tab has already made the API call and then cleared storage, Stripes recognizes this as an expected outcome.

Refs [STCOR-1067](https://folio-org.atlassian.net/browse/STCOR-1067)


src/components/Logout/useLogoutMutation.js
```diff
@@ -58,116 +77,98 @@ export async function clearSession(store, queryClient, timers) {
...
+      try {
+        // Calling `/authn/logout` without an AT destroys the RT, making it
+        // impossible to then rotate and replay the `/authn/logout` call. 🤦
+        // Why doesn't it just return a 401 like every other endpoint? 🤷
+        // To increase the likelihood that logout succeeds, we thus call
+        // `/authn/refresh` with the hope that it will bless us with active
+        // tokens, allowing us to logout successfully and destroy both the
+        // keycloak and FOLIO sessions, keeping the world in sync.
+        await ky.post('authn/refresh', { rtrIgnore: true });
+        await ky.post('authn/logout', { rtrIgnore: true });
```
JohnC-80: Passing the options doesn't work with `useOkapiKy` - they don't get passed through to the 'fetch' unless you override it here with a `fetch` key that passes it in... 

At the same time, I notice the `usePublicGatewayKy ` does not set `credentials` in its options ,  so no cookie getting passed there either.

src/loginServices.js
```diff
@@ -665,7 +662,7 @@ export async function loadResources(store, tenant, userId) {
 
   if (hasSetting) {
     await processLocaleSettings(store, tenantLocaleData, userLocaleData);
-    promises.push(responses.map(res => res?.value));
+    promises.push(...responses.map(res => res?.value));
```
JohnC-80: Why spread the guts of responses here? Already an array from `Promise.allSettled`, and initialized as such.

src/components/Logout/Logout.js
```diff
@@ -31,6 +51,7 @@ import styles from './Logout.css';
 const Logout = ({ sessionTimeoutTimer, sessionTimeoutWarningTimer }) => {
   const { branding, okapi } = useStripes();
   const location = useLocation();
+  const history = useHistory();
```
JohnC-80: SonarDad feels like your neglecting history (unused var)... location.assign instead of `history.push()` ? If we're full on reloading, probably best to remove this unused thing.

JohnC-80 (changes requested):

src/loginServices.js
zburke: I'll move this to an alternate PR since it's entirely unrelated. Thanks for calling attention to it. 

It patches a Sonar complaint about creating a nested array: 
```
list = ['a'];          // => ['a']
list.push(['b', 'c']); // => ['a', ['b', 'c']]
```
A lot of this old code is super messy and undocumented, executing promises that dispatch redux actions but otherwise neglecting to return promises. Is that good because code is just reacting to redux and allowing things to happen in parallel? Or is it bad because it really feels like action-at-a-distance when some random promise-executing function here in loginServices gets called because state changed in a component somewhere?

src/components/Logout/useLogoutMutation.js
zburke: Hard to believe I missed this. I think the best fix is for `useOkapiKy()` to supply the same fetch-override that `usePublicGatewayKy()` does, currying the options arg so we can, indeed, pass in `rtrIgnore`. Agreed?

JohnC-80: We hit this wall before since `useOkapiKy` uses `ky.create()` and still isn't reachable at all through the parameters passed to `ky[method]` usage - unless we add a parameter to the `useOkapiKy` hook itself that bakes it in (I don't think we have tried this yet, but maybe now's the time.)
So it would be like 
```
useOkapiKy(optionsForFetch) {
  ky.create({
    ...
    fetch: (req) => globalThis.fetch(req, optionsForFetch)
  })
}
```

zburke: The options are passed not when the hook is invoked but when the ky-object it returns is used. ~If we pass our own `fetch` implementation that curries its `options` argument, I think we're in good shape. I wrote new tests that check for this and will validate this in a workspace as well.~

😠 , workspace testing shows you are correct. Back to the drawing board.

JohnC-80: Right - My comment was for in case we wanted to build this functionality into the hook somehow - but passing an overriding `fetch: (req) => global.fetch(req, options)` to the `ky.post()` call is fine in this case, also.

JohnC-80 (approved):

PR #1741, opened 2026-06-23
STCOR-1076 - Add RTR timeout listeners to FXHR

JohnC-80:
## [STCOR-1076](https://folio-org.atlassian.net/browse/STCOR-1076)

Currently, if a user executes a long-running upload when approaching their fixed-session timeout, they are caught in a loop without proper messaging as 
1.  Navigation to /logout is attempted
2.  The handler to interrupt navigation (confirmation modal) fires, providing hopeful messaging that the action can continue. 
Sadly, it cannot. No valid token means the action and any following actions will fail.

This PR will hopefully resolve this issue and abort the upload when either of the RTR_TIMEOUT events fire. This should make room for the `/logout` navigation to work and respect the token expiration/system functionality around token timeout.

### DIFF
--- a/src/components/Root/FXHR.js
+++ b/src/components/Root/FXHR.js
@@ -244,8 +244,10 @@ export default (deps) => {
             globalThis.dispatchEvent(new Event(e.type));
           }, 0);
         };
+
         globalThis.addEventListener(RTR_FLS_TIMEOUT_EVENT, abortOnTimeout);
         globalThis.addEventListener(RTR_TIMEOUT_EVENT, abortOnTimeout);
+
         this._removeTimeoutListeners = () => {
           globalThis.removeEventListener(RTR_FLS_TIMEOUT_EVENT, abortOnTimeout);
           globalThis.removeEventListener(RTR_TIMEOUT_EVENT, abortOnTimeout);
--- a/src/components/Root/FXHR.js
+++ b/src/components/Root/FXHR.js
@@ -167,12 +167,17 @@ export default (deps) => {
           */
           this._removeTimeoutListeners?.();
 
-          // fire the change with the current 'DONE but error' state.
+          this.rtrAbortReason = err.message || 'RTR failure while attempting XHR';
+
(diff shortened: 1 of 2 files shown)


src/components/Root/FXHR.js
```diff
@@ -45,6 +47,12 @@ export default (deps) => {
       this.userOnReadyStateChange = null;
       this.FFetchContext = deps;
 
+      /**
+       * Cleanup function for session-timeout abort listeners; null when no
+       * in-flight Okapi request is active.
```
zburke, line 52: "in-flight API request"? "in-flight FOLIO request?"

src/components/Root/FXHR.js
```diff
@@ -70,7 +78,7 @@ export default (deps) => {
      * Invoke the user callback with native XHR-style this binding.
      */
     invokeUserOnReadyStateChange = (event) => {
-      this.userOnReadyStateChange?.(this, event);
+      this.userOnReadyStateChange?.call(this, event);
```
zburke, line 81: IDK why this line is changing; was the previous line wrong? Sonar, too, is skeptical of rebinding the context via `call`; given this is an instance method, the `this` you're rebinding to is already implied.

src/components/Root/FXHR.test.js
```diff
@@ -59,7 +59,7 @@ describe('FXHR', () => {
 
   it('calls other prototype methods...', () => {
     testXHR.addEventListener('abort', mockHandler);
-    testXHR.open('POST', 'okapiUrl');
+    testXHR.open('POST', 'notOkapiUrl');
```
zburke, line 62: Why is this changing?

zburke (commented):
"hopeful messaging" 😆 

Sorry, I'm struggling to connect all the changes in the class and tests back to the issue described in the PR. Can you describe how the changes here implement the desired outcome?

JohnC-80:
Closing in favor of properly handling navigation checks within apps.

PR #1742, opened 2026-06-25
STCOR-1080 do not handle non-rotation errors as fatal errors

zburke:
`rotateAndReplay()` has two jobs: rotate and replay. Rotation-related errors are fatal, but replay-related errors are not. For example, if the rotation request itself fails, or times out, or the response data cannot be stored, those are all fatal errors that need to cause session termination. But if a replay request fails, then such errors should bubble back to the caller, same as they would if the request failed outside `rotateAndReplay()`. Until this PR, this PR, both failures called `config.onFailure()`, effectively treating non-rotation (non-fatal) errors as fatal errors. Whoops.

This PR wraps all rotation-related errors in `RTRError` in order to make it easy for the `rotateAndReplay()` catch-block to distinguish fatal rotation errors from non-fatal fetch errors, calling `config.onFailure()` only for the former.

A common-cause of fetch-failures appears to be AbortControllers used to interrupt fetch calls when navigating away from the route that fired the fetch. For example, some stripes-smart-components provide `useQuery` functions that accept a `signal` argument that react-query can use to cancel an in-flight fetch. As noted at [MDN](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal#examples), 
> When abort() is called, the fetch() promise rejects with a DOMException named AbortError.

There are two approaches we could take in the catch clause:
1. call `config.onFailure()` only when we find specific errors we know are fatal (e.g. an RTRError)
2. call `config.onFailure()` by default, skipping it only when we find specific errors we know are non-fatal (e.g. an AbortError)

Given the severe consequences of a fatal error (i.e. forced logout), this PR implements the former, choosing to look only for the specific errors we know to be fatal. The tradeoff is that this allows non-fatal errors to escape back to the application where it may not be completely clear how to handle them. At present, this feels like an acceptable tradeoff. 

Refs [STCOR-1080](https://folio-org.atlassian.net/browse/STCOR-1080)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@
 * Inspect local expiry information prior to initial request, avoiding 401's in rtr flow. Refs STCOR-1069.
 * Rotate preventively prior to logout. Refs STCOR-1068.
 * Reliably handle unsynchronized FOLIO/Keycloak sessions. Refs STCOR-1067.
+* During rotation, allow non-rotation errors to bubble. Refs STCOR-1080.
 
 ## [11.1.1](https://github.com/folio-org/stripes-core/tree/v11.1.1) (2026-04-20)
 [Full Changelog](https://github.com/folio-org/stripes-core/compare/v11.1.0...v11.1.1)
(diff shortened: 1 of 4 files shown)


JohnC-80 (approved):
I do have a mild concern that this could mean 401's get back to the UI if *anything else triggers the catch clause. We talked about the approach of looking for AbortErrors, specifically, but perhaps this is fine.

PR #1744, opened 2026-06-29
STCOR-1082 supply isGuardable

zburke:
Supply `isGuardable(route)`, a function that returns `true` if the given route may be protected by a navigation-guard to allow the user the opportunity to avoid navigating away from work-in-progress such as a dirty form or ongoing upload, or `false` if the navigation is obligatory, such as redirecting to `/logout` upon session termination.

Refs [STCOR-1082](https://folio-org.atlassian.net/browse/STCOR-1082)

### DIFF
--- a/src/helpers/isGuardable.js
+++ b/src/helpers/isGuardable.js
@@ -1,0 +1,18 @@
+/**
+ * isGuardable
+ * Return true if a route may show a navigation-guard when there is work in
+ * progress (e.g. due to a dirty form or ongoing upload) to prompt whether the
+ * user wants to abandon this work, or false if the navigation is obligatory
+ * (e.g. due to a session being terminated) and user input is irrelevant.
+ *
+ * @param {string} route the destination route
+ * @returns {boolean} true if a route may be guarded; false otherwise
+ */
+const isGuardable = route => {
+  const obligatoryRoutes = ['/logout'];
+
+  // if the given route matches an obligator route, it cannot be guarded
+  return !(obligatoryRoutes.some(i => route.startsWith(i)));
+};
+
+export default isGuardable;
--- a/src/helpers/isGuardable.js
+++ b/src/helpers/isGuardable.js
@@ -8,11 +8,11 @@
  * @param {string} route the destination route
  * @returns {boolean} true if a route may be guarded; false otherwise
  */
-const isGuardable = route => {
+const isGuardable = (route) => {
   const obligatoryRoutes = ['/logout'];
 
   // if the given route matches an obligator route, it cannot be guarded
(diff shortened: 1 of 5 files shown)


src/helpers/isGuardable.js
```diff
@@ -0,0 +1,18 @@
...
+ * (e.g. due to a session being terminated) and user input is irrelevant.
+ *
+ * @param {string} route the destination route
+ * @returns {boolean} true if a route may be guarded; false otherwise
+ */
+const isGuardable = route => {
+  const obligatoryRoutes = ['/logout'];
+
+  // if the given route matches an obligator route, it cannot be guarded
+  return !(obligatoryRoutes.some(i => route.startsWith(i)));
```
JohnC-80: `startsWith` gets through it for this case, and definitely saves you the characters - but it asks for replacement later if we ever need to expand beyond `/logout` pathing. Also `ui-logouthouse`  is going to get special priority with this, as well as the completely speculative `mod-xyz-auth-not-as-crazy-as-keycloak--honest` demands some `/xyz-logout` pathing - so - unless there's something that truly prevents it -  we should just specify the list of unguardable-paths for equality check as opposed to just `startsWith`.

JohnC-80 (changes requested):

src/helpers/isGuardable.js
zburke: This is copied verbatim from the existing implementations in [stripes-form](https://github.com/folio-org/stripes-form/pull/188) and [stripes-final-form](https://github.com/folio-org/stripes-final-form/pull/110). Sure, we can do something else, but there is a reason why this looks the way it does. 

What is your proposed implementation? 
```
// Option 1, ===
const obligatoryRoutes = [
  '/logout', 
  '/potential-unknown-route',
  // ...
];
return !(obligatoryRoutes.some(i => route === i));

// Option 2, function pointer
const predicates = [
  (str) => str === '/logout',
  (str) => { return potentialCustomLogic(str); },
  // ...
];

return !(predicates.some(predicate => predicate(route));
```

JohnC-80: I think for now Option 1 looks best - and as we'll be switching stripes-form and stripes-final-form over to this, they're going to be doing whatever this does. I wonder if there's a reason they just used `startsWith`... aside from the obvious current-developer benefits.

zburke: We had `/logout` and `/logout-timeout` (now `/logout?reason=timeout`). Adding two paths felt too icky so we whitewashed it to "starts with" to justify "hard coding is kinda gross but super easy so let's just do this".

JohnC-80 (approved):

PR #1747, opened 2026-07-08
STCOR-1083 - Use ref for paneTitleRef instead of element for focus-management fix.

JohnC-80:
## [STCOR-1083](https://folio-org.atlassian.net/browse/STCOR-1083)
This resolved the focus-management issue where focus should move to the listing pane when the user navigates to Settings from the Apps dropdown in the main navigation.

This was a simple fix of providing a ref instead of the `ref.current` for `paneTitleRef` on the listing pane of Settings.


zburke (approved):
Looks like this was introduced in #1412. Amazing we didn't notice it for two years! Thanks for the surgical fix.

PR #1750, opened 2026-07-24
STCOR-997 refactor from HotKeys to CommandList

zburke:
Replace `<HotKeys>` with `<CommandList>`. `<CommandList>` delegates to `<HotKeys>` internally, so there is no net effect to this change.

The history is a little murky, but I think it amounts to "We started with `<HotKeys>`, and then wrote `<CommandList>` to provide better ergonomics and to delegate to `<HotKeys>` internally, but we didn't notice that stripes-core still called out to `<HotKeys>` directly."

Converting to `<CommandList>` here removes the last external consumer of `<HotKeys>` from `@folio/stripes-components`, allowing it to be removed there as a public export, freeing us from needing to protect that API, since it will now be for internal use only.

Refs [STCOR-997](https://folio-org.atlassian.net/browse/STCOR-997)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@
 * Supply `mod-settings` permissions for managing others' settings. Refs STCOR-1072.
 * `<Settings>` - properly assign `paneTitleRef` for focus management. Refs STCOR-1083.
 * Supply `<EntitlementChangeBanner>` to indicate entitlement changes. Refs STCOR-780.
+* Refactor from `<HotKeys>` to `<CommandList>`. Refs STCOR-997.
 
 ## [11.1.1](https://github.com/folio-org/stripes-core/tree/v11.1.1) (2026-04-20)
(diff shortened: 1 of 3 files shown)


zburke:
CC from comments on the ticket:
> Simply refactoring from HotKeys to HasCommand doesn’t have a clear benefit at present, given there are no bindings here and no way to supply them (removed in https://folio-org.atlassian.net/browse/UITEN-65 and https://folio-org.atlassian.net/browse/UID-90). 
> 
> We don’t see a clear value in refactoring to `<HasCommand>`. Given we don’t know what shape our near-future hotkeys work will take, it’s likely that whatever we do now will have to be revisited in the near future. Cleaning up the existing implementation is non-trivial; there’s a provider in `<RootWithIntl>`, but also reduxy-code sprinkled all over. Probably, most of it will be removed, but (again) we feel like the best option is to decide on the path forward and then start the work.

PR #1753, opened 2026-08-04
STCOR-1094 Support module-contributed main navigation entries

usavkov-epam:
## Purpose
https://folio-org.atlassian.net/browse/STCOR-1094

Stripes Core currently renders a static Main Navigation bar with only hard-coded buttons (Help, Profile). There is no API for FOLIO modules to contribute their own entries to the main navigation area, which blocks use cases that require a module-owned, always-visible nav trigger — such as a stateful popover, a real-time badge indicator, or a custom event-driven action.

This PR introduces the `stripes.links.mainNavigation` registration API and the `MainNavButtons` component that powers it.

## Approach

Module authors declare nav entries in package.json under `stripes.links.mainNavigation`. Each entry specifies an icon, an i18n caption key, an optional visibility `check`, and exactly one behavior field. Stripes Core resolves the behavior at runtime in priority order:

1. **`render`** — the module root exposes a renderer function that receives `renderTrigger` and `triggerProps`. This covers stateful triggers (popovers, overlays, badges) where the module owns the interaction model.
2. **`event`** — the click is dispatched through the existing `handlerService`, keeping side-effect-only actions out of the render tree.
3. **`href`** — renders a plain link nav button.
4. **`route`** — renders an internal-route nav button.

Button refs are cached in a stable `Map` across renders so that overlay anchors (e.g. `Popper`) remain valid while a panel is open.

The existing hard-coded Help button was moved into `MainNavButtons` to consolidate all right-side nav rendering in one place.

## Changes

- **MainNavButtons.js** *(new)* — core component; resolves and renders all module-contributed nav entries plus the Help button.
- **MainNavButtons.css** *(new)* — flex layout with design-token gap for the button group.
- **index.js** *(new)* — barrel export.
- **MainNav.js** — replaces inline Help button markup with `<MainNavButtons />`.
- **NavButton.js** — adds `badgeSize` prop (`'small' | 'medium'`) forwarded to `Badge`, enabling `render`-mode modules to control badge sizing.
- **index.js** — exports `NavButton` as part of the public API so `render`-mode modules can use it directly if needed.
- **main-navigation-links.md** *(new)* — full developer guide: registration, all four modes, visibility checks, the `renderTrigger`/`triggerProps` contract, and annotated examples.
- **README.md** — links to the new guide.
- **CHANGELOG.md** — entry added.

## Screen recording

https://github.com/user-attachments/assets/408b5e2c-1732-407c-9ba2-ed92045d11af

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@
 * Supply `<EntitlementChangeBanner>` to indicate entitlement changes. Refs STCOR-780.
 * Correctly display session-remaining time after refresh. Refs STCOR-1087.
 * Rename `button.new` from "+ New" to "New". Refs STCOR-1093.
+* Support module-contributed main navigation entries. STCOR-1094.
 
 ## [11.1.1](https://github.com/folio-org/stripes-core/tree/v11.1.1) (2026-04-20)
 [Full Changelog](https://github.com/folio-org/stripes-core/compare/v11.1.0...v11.1.1)
--- a/README.md
+++ b/README.md
@@ -23,6 +23,7 @@ The documentation roadmap has moved.  Please refer to the [Stripes](https://gith
 * [A component hierarchy example: the "Patrons" module](doc/component-hierarchy.md) shows by example how a set of components -- some connected, some not -- can work together to implement part of an application. **NOTE that this is somewhat out of date**, but still helpfully illustrative.
 * [Adding new permissions to FOLIO UI modules](doc/adding-permissions.md).
 * UX and implementation concerns for [Settings and Preferences](doc/settings-and-preferences.md)
+* [Main Navigation links developer guide](doc/main-navigation-links.md)
 
 
 ## Additional information
--- a/doc/main-navigation-links.md
+++ b/doc/main-navigation-links.md
@@ -1,0 +1,426 @@
+# Main Navigation Links: Developer Guide
+
+## Overview
+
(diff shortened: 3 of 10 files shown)


BogdanDenis (approved):

usavkov-epam:
Hello @folio-org/stripes-force, could you please take a look at the PR? Thanks in advance!

zburke (commented):
@usavkov-epam, this is exactly the approach I asked you to take (thank you!) ... but on review with wiser council, I think it may not be the best because of the extent to which it opens the navbar to modification (some might even say pollution) by any application. I feel badly about this because there is a lot of great, thoughtful work here that may not be usable. 

The approach we'd like to see in stripes-core instead is the one we initially discussed and rejected: somewhere in `<MainNav>`, add 
```
<Pluggable type="notifications" />`
```
and put the entire implementation in the plugin. This explicitly enshrines `notifications` (or whatever you decide the type should be) in stripes-core. You asked about this approach initially and I rejected it due to not wanting to add a specifically-notifications-shaped API surface to stripes-core. As noted above, however, the concerns about adding a completely generic navbar-icon-shaped API surface weighed even more heavily when I discussed this with colleagues. 

I'm sorry for the pivot, and for the extreme delay here, both coming at your expense. Please get in touch if you would like to discuss this in greater detail. Again, my apologies. 

CC: @NikitaSedyx

usavkov-epam:
Closing this PR as we’ve chosen a different solution for the implementation.

PR #1756, opened 2026-08-20
STCOR-1099 supply rtrIgnore to forgot-* API requests

zburke:
Supply `rtrIgnore: true` on the forgot-password and forgot-username API requests, allowing them to avoid invoking RTR when credentials are not present, which of course they **won't be because the whole reason for accessing these routes is that an end-user cannot remember their credentials.

This bug has been lurking for a while but was masked by the initial RTR implementation that amounted to "play a request; on failure, if credentials are required-but-missing, rotate; replay the request". The loophole is that all requests would be played once prior to invoking RTR. This meant requests that require credentials (as most do) would fail, polluting the console, but it also meant the requests that do NOT require credentials (and these two do NOT) would succeed, even though they do not specify `rtrIgnore: true` to indicate that they should bypass rotation.

The loophole was closed in [STCOR-1069](https://folio-org.atlassian.net/browse/STCOR-1069) in PRs #1732 and #1737 and was revealed by testing for [STCOR-993](https://folio-org.atlassian.net/browse/STCOR-993) in PR #1755.

New unit tests courtesy Codex GPT 5.6 Luna

Refs [STCOR-1099](https://folio-org.atlassian.net/browse/STCOR-1099)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@
 * Supply `<EntitlementChangeBanner>` to indicate entitlement changes. Refs STCOR-780.
 * Correctly display session-remaining time after refresh. Refs STCOR-1087.
 * Rename `button.new` from "+ New" to "New". Refs STCOR-1093.
+* Supply `rtrIgnore` on forgot-password and forgot-username API requests. Refs STCOR-1099.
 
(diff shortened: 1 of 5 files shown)


JohnC-80 (approved):
👍 Tiny code change.
👍 Hooray Tests!

PR #1759, opened 2026-09-18
STCOR-1101 provide expected locale-region values for moment

zburke:
FOLIO locales look like `zh-TW`, but moment's look like `zh-tw`, so the FOLIO values must be lower-cased for moment to find them.

Refs [STCOR-1101](https://folio-org.atlassian.net/browse/STCOR-1101)


zburke:
Replaced by #1760, which does all of this and more.

PR #1760, opened 2026-09-21
STCOR-1102 default to en-US given an invalid locale

zburke:
Given an invalid locale, recover and default to `en-US`. Previously, an invalid locale string such as `en-US-u-nu-latn-u-nu-latn` would cause stripes to halt midway through initialization, showing a flash of available applications followed by a blank screen.

The built-in Intl APIs have come a long way since i18n as first introduced to stripes in 7d8f0fc74 in 2017. This makes processing locale data both simpler and more robust, allowing us to use the `Intl.Locale` constructor to validate the input, and to use its return value to provide the `language` and `region` values we previously parsed out manually.

Bad locale values may come from improperly stored data, e.g. if a locale is stored with a duplicate numbering-system attribute like
```
{ locale: "en-US-u-nu-latn", numberingSystem: "latn" }
```
then numbering-system will be applied twice.

Refs [STCOR-1102](https://folio-org.atlassian.net/browse/STCOR-1102), [STCOR-1101](https://folio-org.atlassian.net/browse/STCOR-1101)


JohnC-80 (approved):
The times we live in.. can't even trust a provided 'locale' parameter to be valid. Will Intl constructor throw anything besides a 'RangeError'? Why limit the ~`en_US`~ er um `en-US` fallback to just that case?

zburke:
> Will Intl constructor throw anything besides a 'RangeError'? 

Good question, @JohnC-80. From [mozdev docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl#locales_argument):
> If a locale identifier is not a string or an object, a [TypeError](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypeError) is thrown. If a locale identifier is a string that's syntactically invalid, a [RangeError](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RangeError) is thrown.

Nonethless, given a bad locale will cause the `IntlProvider` to barf, capturing all errors here and falling back to `en-US` feels like a responsible approach.

